PR #231, opened 2018-08-06
Be extra safe with missing nonces in getNonce

### LINKED ISSUE #230: getNonce fails if there is no outgoing tx on the chain
The logic in the code is flawed, as if nonce is undefined, the eval will fail and throws.

Error:
```
TypeError: Cannot read property 'nonce' of undefined
```

https://github.com/iov-one/web4/blob/b07ddbc/packages/iov-web4/src/web4write.ts#L53
```
  public async getNonce(chainId: ChainId, addr: Address): Promise<Nonce> {
    const nonce = await this.mustGet(chainId).client.getNonce({ address: addr });
    return nonce.data.length === 0 ? (Long.fromInt(0) as Nonce) : nonce.data[0].nonce; #### This line is the failure, we can't return nonce.data[0].nonce ####
  }
```

ethanfrey:
Resolves #230

### DIFF
--- a/packages/iov-core/src/writer.ts
+++ b/packages/iov-core/src/writer.ts
@@ -56,7 +56,7 @@ export class IovWriter {
   // not the ful bcp info.
   public async getNonce(chainId: ChainId, addr: Address): Promise<Nonce> {
     const nonce = await this.mustGet(chainId).client.getNonce({ address: addr });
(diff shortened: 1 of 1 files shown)


Isabello:
I cant reproduce the issue, will close this and revisit later if it shows up again.

PR #249, opened 2018-08-09
Expose binary from @iov/cli

### LINKED ISSUE #248: @iov/cli should install binary (work outside of monorepo)
To make the tool more useful, we want to be able to do something like:

```
yarn add -D @iov/cli
// OR npm install -g @iov/cli
iov-cli // or other name
```

and that would launch the repl, complete with init script..

It seems that the cli only works inside the package in the monorepo, when called via `yarn cli`

ethanfrey:
Resolves #248


ethanfrey:
Now has `bin` exposed and `yarn test-bin` should run the same code as if we installed it.

It is a trimmed-down copy of ts-node repl now and basically works.
Current issue:

`import { Encoding} from '@iov/encoding'` works
`import { IovWriter } from '@iov/core'` fails with unknown module....

not sure how to set up the imports all proper.
but once that is working, it shouldn't be too hard to expose the `init` file like i do the demo `index.ts`

Ideally we could get rid of the `c` prefix and you find a cleaner way, but at least this version is an installable binary that should work with `npm install -g @iov/cli`

webmaster128:
This is getting interesting: I have a wrapper that evaluates JavaScript code in an async context (i.e. allowing the use of top level await. The replEval on the other hand has an asynchronous interface anyway (via callback).

The two main issues remaining now are:
- the TypeScript compiler blocks the top level await as invalid syntax
- the async wrapper creates a sub-scope, such that after doing `const x = away myCall()` in the repl, `x` is not defined in the main repl scope

packages/iov-cli/src/helpers.ts
```diff
@@ -0,0 +1,37 @@
+import { TSError } from "ts-node";
+import { Script } from "vm";
+
+export function executeJavaScript(code: string, filename: string) {
+  const script = new Script(code, { filename: filename });
+  return script.runInThisContext();
```
ethanfrey: I'm still wondering where this "context" comes from, with all the variables declared in the repl. Seems a bit magic, but maybe this is a module variable from vm?

packages/iov-cli/tsconfig_repl.json
```diff
@@ -0,0 +1,10 @@
+{
```
ethanfrey: :+1:

packages/iov-cli/src/tsrepl.ts
```diff
@@ -0,0 +1,195 @@
...
+export class TsRepl {
+  private readonly typeScriptService: Register;
+  private readonly initialTypeScript: string;
+  private readonly debuggingEnabled: boolean;
+  private readonly evalFilename = `[eval].ts`;
+  private readonly evalPath = join(process.cwd(), this.evalFilename);
+  private readonly evalData = { input: "", output: "", version: 0, lines: 0 };
+
+  constructor(tsconfigPath: string, initialTypeScript: string, debuggingEnabled: boolean) {
+    this.typeScriptService = register({ project: tsconfigPath });
```
ethanfrey: very nice

packages/iov-cli/src/tsrepl.ts
```diff
@@ -0,0 +1,195 @@
...
+    const resetEval = this.appendEval("");
+
+    const reset = (): void => {
+      resetEval();
+
+      // Hard fix for TypeScript forcing `Object.defineProperty(exports, ...)`.
+      executeJavaScript("exports = module.exports", this.evalFilename);
+
+      // Ensure code ends with "\n" due to implementation of replEval
+      this.replEval(this.initialTypeScript + "\n");
```
ethanfrey: Ahhh, very smart, this is where you plug in the init stuff :)

packages/iov-cli/src/tsrepl.ts
```diff
@@ -0,0 +1,195 @@
...
+    const lastResult = changes.reduce((result, change) => {
+      return change.added ? executeJavaScript(change.value, this.evalFilename) : result;
+    }, undefined);
+
+    return lastResult;
+  }
+
+  private async replEval(code: string): Promise<ReplEvalResult | undefined> {
+    // Delay eval execution – because we can
+    await sleep(3);
```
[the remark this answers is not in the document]
webmaster128: it is a proof of concept showing that repl evaluation is an async thing (which is great). But it can go, yes

webmaster128:
The DeprecationWarning is https://github.com/bluelovers/deasync/issues/2 but I don't see a way to set NODE_NO_WARNINGS=1 nefore starting node automatically at the moment

Issue #220, opened 2018-08-03
Self documenting REPL


ethanfrey:
Building on #219 

It would be nice for a developer to get some feedback and API insight directly in the REPL. We have this is the IDE, and it doesn't have to be so nice, but at least listing the methods on a class and arguments to a method....

I hacked something like this into the javascript weave-js repl:
  * https://github.com/confio/weave-js/blob/master/bin/cli.js#L36-L64
  * https://github.com/confio/weave-js/blob/master/bin/args.js

We should be able to do something at least as powerful with the typescript info

PR #260, opened 2018-08-10
Make .type work in the repl again

ethanfrey:
Testing out the repl, I discovered the useful `.type` command was broken due to some of the refactoring. This is the simplest change to get it to work again.

You can also reject this PR for a better change that addresses the underlying causes.
I was actually taking a short look for #220 to see if I could extend `.type` to show all methods on an object, but then found it wasn't working at all.

I would love `.type UserProfile` to show static methods, and `const foo = new UserProfile(); .type foo` to show all instance methods... but at least seeing `.type toHex` work is a nice start

### DIFF
--- a/packages/iov-cli/src/tsrepl.ts
+++ b/packages/iov-cli/src/tsrepl.ts
@@ -55,6 +55,7 @@ export class TsRepl {
 
     // Bookmark the point where we should reset the REPL state.
     const resetEval = this.appendEval("");
+    +"\n";
 
     const reset = (): void => {
       resetEval();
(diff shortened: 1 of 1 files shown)


packages/iov-cli/src/tsrepl.ts
```diff
@@ -55,6 +55,7 @@ export class TsRepl {
 
     // Bookmark the point where we should reset the REPL state.
     const resetEval = this.appendEval("");
+    +"\n";
```
webmaster128: Is this supposed to be `const resetEval = this.appendEval("\n");`?

webmaster128:
I'll have a look at that again to understand the underlying issue better

PR #265, opened 2018-08-15
[iov-cli] make top level await possible in repl

webmaster128:
based on #264

Wow, what a journey: Not the most beautiful piece of software but should do the job. Top level await works as long as the TS compiler does not think it is a variable.

ok

```ts
await 1
await new Array(1, 2)
await abc
await abc.xyz()
```

fail

```ts
await (1)
await [1]
await [1, 2]
```


packages/iov-cli/src/helpers.spec.ts
```diff
@@ -1,16 +1,147 @@
...
+      expect(await executeJavaScriptAsync(code, "myfile.js", context)).toEqual("job done");
+    });
+
+    it("can execute timeout code in multiple statements", async () => {
+      const context = createContext({ setTimeout: setTimeout });
+      const code = `
+        const promise = new Promise(resolve => {
+          setTimeout(() => resolve('job done'), 5);
+        });
+        await (promise);
```
rudi-cilibrasi: why are there parentheses around that `promise` ?

webmaster128: Because the default parser used by recast needed them. Switching to babylon as a JavaScript parser makes them unnecessary. So they could be refactored away now.

Isabello:
The wait helper function is still available, should it be removed or ?

The PR works as expected, though.

webmaster128:
> The wait helper function is still available, should it be removed or ?

It is obsolete but I'd like to keep it around for one release – just in case the top level await feature does not work sometimes. Tracked as https://github.com/iov-one/iov-core/issues/274

packages/iov-cli/src/async.ts
```diff
@@ -0,0 +1,48 @@
+import recast = require("recast");
+import babylon = require("babylon");
+
+export function wrapInAsyncFunction(code: string): string {
```
ethanfrey: All so magic. But glad to see nice test coverage and comments

packages/iov-cli/src/async.ts
```diff
@@ -0,0 +1,48 @@
...
+  const codeInAsyncFunction = `(async () => {
+    ${code}
+  })()`;
+
+  const ast = recast.parse(codeInAsyncFunction, { parser: babylon });
+  // console.log(recast.print(ast));
```
ethanfrey: Maybe these can go away though?

PR #277, opened 2018-08-17
Add AppVeyor config

webmaster128:


webmaster128:
Damn, AppVeyor runs on Windows Server 2016, that do not support Edge

PR #297, opened 2018-08-27
[iov-tendermint-rpc] Expose API to subscribe to events

webmaster128:


packages/iov-tendermint-rpc/src/client.ts
```diff
@@ -89,6 +91,24 @@ export class Client {
...
+    if (!instanceOfRpcStreamingClient(this.client)) {
+      throw new Error("This RPC client type cannot subscribe to events");
+    }
+
+    const request: requests.SubscribeRequest = { method: Method.SUBSCRIBE, type: eventType };
+    const req = this.p.encodeSubscribe(request, query || undefined);
+    const eventStream = this.client.listen(req);
+    return eventStream.map<responses.SubscriptionEvent>(event => {
+      // tslint:disable-next-line:no-console
+      // console.log(JSON.stringify(event));
```
webmaster128: uncomment me for debugging

packages/iov-tendermint-rpc/src/adaptor.ts
```diff
@@ -32,6 +32,10 @@ export interface Params {
   readonly encodeGenesis: (req: requests.GenesisRequest) => JsonRpcRequest;
   readonly encodeHealth: (req: requests.HealthRequest) => JsonRpcRequest;
   readonly encodeStatus: (req: requests.StatusRequest) => JsonRpcRequest;
+  readonly encodeSubscribe: (
```
ethanfrey: Hmmm... you have the following types:

```
export interface SubscribeRequest {
  readonly method: Method.SUBSCRIBE;
  readonly type: SubscriptionEventType;
}

export interface SubscribeRequestQuery {
  readonly tags: ReadonlyArray<Tag>;
}
```

This is different than all the others, that keep a format closer to the wire. I understand that there is some string building going on here, so it is not as clear cut, but why not something like:

```
export interface SubscribeRequest {
  readonly method: Method.SUBSCRIBE;
  readonly query: {
    readonly type: SubscriptionEventType;
    readonly tags?: ReadonlyArray<Tag>;
  }
}
...
readonly encodeSubscribe: (req: requests.SubscribeRequest) => JsonRpcRequest;
```

One is not "better", but the second seems more consistent with the rest

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -159,9 +157,104 @@ describe("Client", () => {
     expect(info3).toBeTruthy();
   });
 
-  // don't print out WebSocket errors if marked pending
-  const onError = skipTests() ? () => 0 : console.log;
+  describe("With HttpClient", () => {
```
ethanfrey: This is a nice twist on the code reuse.

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -159,9 +157,104 @@ describe("Client", () => {
...
+    kvTestSuite(() => new HttpClient(tendermintUrl));
+  });
+
+  describe("With WebsocketClient", () => {
```
ethanfrey: And I like how you embed the common tests in the websocket group.
👍

packages/iov-tendermint-rpc/src/client.ts
```diff
@@ -1,7 +1,9 @@
+import { Stream } from "xstream";
+
 import { Adaptor, Decoder, Encoder, findAdaptor, Params, Responses } from "./adaptor";
-import { default as requests, Method } from "./requests";
+import { default as requests, Method, SubscribeRequestQuery, SubscriptionEventType } from "./requests";
```
ethanfrey: There were a LOT of things in requests. So rather than have a huge import here, I just used `requests.*` everywhere. Only exception was `Method`, which was used so much.

I would prefer for consistency to either refer to `requests.SubscribeRequestQuery` and `requests.SubscriptionEventType` and remove from the import.
OR
Explicitly import all the types from `requests.*` and drop the default import here

packages/iov-tendermint-rpc/src/client.ts
ethanfrey: This was a very useful comment. Can probably go now that it all works with #298, right?

packages/iov-tendermint-rpc/src/common.ts
```diff
@@ -33,7 +33,7 @@ export interface JsonRpcEvent {
   readonly query: string;
   readonly data: {
     readonly type: string;
-    readonly value: {};
+    readonly value: any;
```
ethanfrey: Question: do you need this?

`{}` kind of "enforces" that value is not a number or string. `any` allows those as well. If we have those types, please do change this. Otherwise, I don't see it adds anything

packages/iov-tendermint-rpc/src/requests.ts
```diff
@@ -38,6 +39,17 @@ export type Request =
   | TxSearchRequest
   | ValidatorsRequest;
 
+/**
+ * Raw values must match the tendermint event name
+ *
+ * @see https://godoc.org/github.com/tendermint/tendermint/types#pkg-constants
```
ethanfrey: :+1: great reference, I should add more here as well

packages/iov-tendermint-rpc/src/requests.ts
```diff
@@ -102,6 +114,15 @@ export interface StatusRequest {
   readonly method: Method.STATUS;
 }
 
+export interface SubscribeRequest {
```
ethanfrey: See comment above in `adapter.ts`. I would like to merge these two interfaces into one, as they are not used in isolation, and this maintains consistency

packages/iov-tendermint-rpc/src/rpcclient.spec.ts
```diff
@@ -112,6 +120,41 @@ describe("RpcClient", () => {
       });
     });
 
+    it("can execute commands while listening to events", done => {
```
ethanfrey: Good test. Thank you for finding and fixing this issue

packages/iov-tendermint-rpc/src/rpcclient.ts
```diff
@@ -22,6 +22,10 @@ export interface RpcStreamingClient extends RpcClient {
   readonly listen: (request: JsonRpcRequest) => Stream<JsonRpcEvent>;
 }
 
+export function instanceOfRpcStreamingClient(client: RpcClient): client is RpcStreamingClient {
+  return typeof (client as any).listen === "function";
```
ethanfrey: Nice check... I should fix some of my guards to be like this. (I think I check for existence of the key, not even the type of the value)

packages/iov-tendermint-rpc/src/v0-20/requests.ts
```diff
@@ -31,6 +32,16 @@ export class Params extends requests.DefaultParams {
     return jsonRpcWith(req.method, req.params);
   }
 
+  public static encodeSubscribe(
+    req: requests.SubscribeRequest,
+    query: requests.SubscribeRequestQuery | undefined,
+  ): JsonRpcRequest {
+    const allTags: ReadonlyArray<Tag> = [{ key: "tm.event", value: req.type }, ...(query ? query.tags : [])];
+
+    const queryString = requests.buildTxQuery({ tags: allTags });
```
ethanfrey: I like how you make the `tm.event` fit alongside other tags. Nice unification.

Does `requests.buildTxQuery` make sense as is or needs to be modified?

packages/iov-tendermint-rpc/src/v0-20/responses.ts
```diff
@@ -76,6 +76,25 @@ export class Responses {
...
+      // block header
+      height = event.data.value.header.height;
+    } else if (event.data.value.TxResult) {
+      // transaction
+      height = event.data.value.TxResult.height;
+    } else {
+      throw new Error("Unknown rpc event:" + JSON.stringify(event));
+    }
+
+    return decodeSubscriptionEvent({
```
ethanfrey: Woah, you really throw away all the other info except for the height of the change? So you just say "something happened, now go query"?

I kind of feel we should add some more info to the events. What do you think?

This can be another PR, but I was wondering why you did this instead of a union type for example.

packages/iov-tendermint-rpc/src/v0-20/requests.ts
webmaster128: Makes perfect sense as the subscription type is just another tag (e.g. `tm.event='Tx'` or `tm.event='Tx' AND app.creator='jae'`). Also the Tendermint documentation states that those subscription queries are supposed to work like searches.

packages/iov-tendermint-rpc/src/v0-20/responses.ts
webmaster128: No I don't. This is just not ready

packages/iov-tendermint-rpc/src/common.ts
webmaster128: The problem is that with an interface definition like this, every user needs to explicitly cast to any.

Before:

`(events[i].data.value as any).header.height`

After:

`events[i].data.value.header.height`

The old version has no better type safety in the getter but is far harder to read. I see that this allows assigning numbers and strings, but as long as there is no "any interface" type, I think `any` describes the property best.

packages/iov-tendermint-rpc/src/client.ts
webmaster128: Hmm, did not think about that. This was added by the IDE. In general, I'd prefer to just list them all like we do in the other places as long as the symbols are unique enough to not have conflicts. But for now, I go with the existing style

webmaster128: But this import makes no sense, because there is no default export in `./requests`. This makes `requests.SubscriptionEventType` a type (maybe a TS bug) and `requests.SubscriptionEventType.NewBlock` unavailable, because `requests` is undefined at runtime.

I don't think there is syntax for importing everything from one module plus some separate symbols. You either have

```js
import * as requests from "./requests";
```

or

```js
import { all, the, things, Method } from "./requests";
```

webmaster128: The name `Request` from "./requests" caused conflicts with some DOM type. Thus I go with `import * as requests from "./requests";`

packages/iov-tendermint-rpc/src/requests.ts
webmaster128: When writing new docs, have a quick look at https://github.com/Microsoft/tsdoc. I am not using much of it yet but starting with `/**` and a on-liner description for the cases where some words are necessary is probably a good start

packages/iov-tendermint-rpc/src/responses.ts
```diff
@@ -115,6 +116,22 @@ export interface ValidatorsResponse {
   readonly results: ReadonlyArray<Validator>;
 }
 
+// Events
+
+export interface NewBlockEvent extends Block {}
+
+export interface NewBlockHeaderEvent extends Header {}
+
+export interface TxEvent {
```
ethanfrey: Hmmm... odd that `TxEvent` is different than `TxResponse` (subscribe vs. query).

I mean, I guess proof would always be empty here, but I would assume `txResult` could be present in `TxEvent` and `result` with tags present in `TxResponse`.

I should investigate tendermint more.

packages/iov-tendermint-rpc/src/v0-20/requests.ts
```diff
@@ -31,6 +32,16 @@ export class Params extends requests.DefaultParams {
     return jsonRpcWith(req.method, req.params);
   }
 
+  public static encodeSubscribe(req: requests.SubscribeRequest): JsonRpcRequest {
+    const allTags: ReadonlyArray<Tag> = [
+      { key: "tm.event", value: req.query.type },
+      ...(req.query.tags ? req.query.tags : []),
```
ethanfrey: Style question...

In such a case, i would usually write `...(req.query.tags || [])`
Both versions are fine. Wondering if there is a reason to use one over the other, or just personal preference.

packages/iov-tendermint-rpc/src/v0-20/responses.ts
```diff
@@ -76,6 +76,18 @@ export class Responses {
     return decodeStatus(response.result as RpcStatusResponse);
   }
 
+  public static decodeNewBlockEvent(event: JsonRpcEvent): responses.NewBlockEvent {
```
ethanfrey: I like this clean reuse

packages/iov-tendermint-rpc/src/v0-20/requests.ts
webmaster128: Both ways are fine. The ternary operator was the best thing that case to my mind.

I slightly prefer to keep the test and the return value independent allowing modifications to one or the other more easily, e.g. `(req.query.tags === null ? req.query.tags : [])`

packages/iov-tendermint-rpc/src/responses.ts
webmaster128: I did not have the experience and the motivation to dig into that. What I saw at a first glance was different structures so I built the type from my example return data. Maybe it can be uniform, maybe not.

ethanfrey: I think you are right the data is different. They really don't understand the need for a nice rpc api... 18 months later....

PR #298, opened 2018-08-27
[iov-tendermint-rpc] Fix bug in situations where listen and execute is used

webmaster128:


packages/iov-tendermint-rpc/src/rpcclient.spec.ts
```diff
@@ -112,6 +112,41 @@ describe("RpcClient", () => {
       });
     });
 
+    it("can execute commands while listening to events", done => {
+      pendingWithoutTendermint();
+
+      const client = new WebsocketClient(tendermintUrl);
+
+      const query = "tm.event='NewBlockHeader'";
```
ethanfrey: Nice low-level setup

packages/iov-tendermint-rpc/src/rpcclient.spec.ts
```diff
@@ -112,6 +112,41 @@ describe("RpcClient", () => {
...
+            // wait 1.5s and check we did not get more events
+            setTimeout(() => {
+              expect(events.length).toEqual(2);
+              done();
+            }, 1500);
+          }
+        },
+      });
+
+      const startusResponse = client.execute(jsonRpcWith(Method.STATUS));
```
ethanfrey: Good test, but i would almost put some delay here to make sure that the subscription finished first.

`setTimeout(100, () => ...)` ?

packages/iov-tendermint-rpc/src/rpcclient.ts
```diff
@@ -185,16 +185,31 @@ export class WebsocketClient implements RpcStreamingClient {
       // only one of the two listeners should fire, and it will
       // deregister the other so as not to cause a leak.
 
+      const good = {
+        eventName: id,
+        listener: (data: JsonRpcResponse) => {
+          unregisterGoodAndBad();
```
ethanfrey: maybe just `unregisterBad()`? `unregisterGood` would already happen via `once`

webmaster128: You mean having two different cleanup functions? Right now the idea is to cleanup both such that only one function is needed.

packages/iov-tendermint-rpc/src/rpcclient.spec.ts
webmaster128: could be done, yes. My idea was to keep the test as simple as possible while catching the bug. But I am open for future improvements

PR #300, opened 2018-08-28
Create @iov/faucets

### LINKED ISSUE #293: Add faucet functionality

webmaster128:
Closes #293

### DIFF
--- a/packages/iov-cli/README.md
+++ b/packages/iov-cli/README.md
@@ -205,7 +205,7 @@ When using a Testnet, you can use the BovFaucet to receive tokens:
 
 > const bovFaucet = new BovFaucet("https://faucet.friendnet-slow.iov.one/faucet");
 
-> await bovFaucet.open(meAddress)
+> await bovFaucet.credit(meAddress)
 > (await reader.getAccount({ address: meAddress })).data[0].balance
 [ { whole: 10,
     fractional: 0,
@@ -213,7 +213,7 @@ When using a Testnet, you can use the BovFaucet to receive tokens:
     tokenName: 'Main token of this chain',
     sigFigs: 6 } ]
 
-> await bovFaucet.open(meAddress, "PAJA" as TokenTicker)
+> await bovFaucet.credit(meAddress, "PAJA" as TokenTicker)
 > (await reader.getAccount({ address: meAddress })).data[0].balance
 [ { whole: 10,
     fractional: 0,
--- a/packages/iov-cli/README.md
+++ b/packages/iov-cli/README.md
@@ -185,6 +185,48 @@ Error: Socket was closed, so no data can be sent anymore.
     at ...
 ```
 
+## Faucet usage
+
+When using a Testnet, you can use the BovFaucet to receive tokens:
+
+```
+> const mnemonic = Bip39.encode(await Random.getBytes(16)).asString();
+> mnemonic
+'helmet album grow detail apology thank wire chef fame core private cargo'
+> const profile = new UserProfile();
+> profile.addEntry(Ed25519SimpleAddressKeyringEntry.fromMnemonic(mnemonic));
+> const me = await profile.createIdentity(0);
+
+> const knownChains = await withConnectors([await bnsConnector("https://bov.friendnet-slow.iov.one")]);
+> const writer = new IovWriter(profile, knownChains);
+> const chainId = writer.chainIds()[0];
+> const reader = writer.reader(chainId);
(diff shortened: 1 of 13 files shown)


webmaster128:
Travis fails because @iov/faucets is not on npm yet, which is required by the @iov/cli installation test. We could split the creation of the package and the usage but this adds a lot of overhead. I'd rather merge and release with red CI in this case.

packages/iov-cli/README.md
```diff
@@ -185,6 +185,48 @@ Error: Socket was closed, so no data can be sent anymore.
...
+
+```
+> const mnemonic = Bip39.encode(await Random.getBytes(16)).asString();
+> mnemonic
+'helmet album grow detail apology thank wire chef fame core private cargo'
+> const profile = new UserProfile();
+> profile.addEntry(Ed25519SimpleAddressKeyringEntry.fromMnemonic(mnemonic));
+> const me = await profile.createIdentity(0);
+
+> const knownChains = await withConnectors([await bnsConnector("https://bov.friendnet-slow.iov.one")]);
```
ethanfrey: yes, we need to clean this api up. big time.

packages/iov-cli/README.md
```diff
@@ -185,6 +185,48 @@ Error: Socket was closed, so no data can be sent anymore.
...
+> const writer = new IovWriter(profile, knownChains);
+> const chainId = writer.chainIds()[0];
+> const reader = writer.reader(chainId);
+> const meAddress = writer.keyToAddress(chainId, me.pubkey);
+
+> const bovFaucet = new BovFaucet("https://faucet.friendnet-slow.iov.one/faucet");
+
+> await bovFaucet.open(meAddress)
```
ethanfrey: Why is it `.open(...)`?

I like the optional ticker second argument, but what about a name like `bovFaucet.{distribute,grant,send,fill}(meAddress)`

packages/iov-faucets/README.md
```diff
@@ -0,0 +1,14 @@
+# @iov/faucets
+
+[![npm version](https://img.shields.io/npm/v/@iov/faucets.svg)](https://www.npmjs.com/package/@iov/faucets)
+
+Use some Testnet faucets from JavaScript.
+
+## API Documentation
```
ethanfrey: Maybe you can add a two liner example here as well, not just in cli.

```
const bovFaucet = new BovFaucet("https://faucet.friendnet-slow.iov.one/faucet");
await bovFaucet.open(addr)
```

Sure, you need to set an address and query and all that, but this shows main usage quite simply.

packages/iov-faucets/src/bovfaucet.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { Address, TokenTicker } from "@iov/bcp-types";
+import { Random } from "@iov/crypto";
+
+import { BovFaucet } from "./bovfaucet";
+
+describe("BovFaucet", () => {
```
ethanfrey: Seems a bit funny for unit tests to run against a live network, but there would be too much overhead to actually set up and run a local faucet, right?

We could start up https://hub.docker.com/r/iov1/faucet/tags/ but I will leave this as an open issue so we don't depend on friendnet so much.

packages/iov-faucets/src/bovfaucet.ts
```diff
@@ -0,0 +1,24 @@
...
+    this.url = url;
+  }
+
+  public async open(address: Address, ticker: TokenTicker = "IOV" as TokenTicker): Promise<void> {
+    const body = {
+      address: Encoding.toHex(address),
+      tokens: [{ ticker: ticker }],
+    };
+
+    await axios.post(this.url, body);
```
ethanfrey: Did you test this with karma (in browser). I mean, it would be great to have a button saying "get free tokens" in a webapp, so usecase makes sense to me.  I had issues with axios in tendermint rpc in browser. Can you add a test? You may need to make a fix like this (or improve/refactor my fix):

https://github.com/iov-one/iov-core/blob/master/packages/iov-tendermint-rpc/src/rpcclient.ts#L33-L55

webmaster128: Good point. Added browser tests but will keep them out of here because of https://github.com/iov-one/tools/issues/71

packages/iov-faucets/README.md
webmaster128: done

Issue #302, opened 2018-08-28
Better handling of paths in HD keyrings


ethanfrey:
Right now, the main keyring we use is `Ed25519SimpleAddressKeyringEntry` which hardcodes the beginning of a path and just increments the last value to produce multiple addresses. A `KeyringEntry` has just one command to create a new key, which accepts no arguments.

```
export interface KeyringEntry {
...
  // createIdentity will create one new identity
  readonly createIdentity: () => Promise<LocalIdentity>;
...
}
```

This is exposed in the `UserProfile` as:

```
  // creates an identity in the n-th keyring entry of the primary keyring
  public async createIdentity(n: number): Promise<LocalIdentity> {
    const entry = this.entryInPrimaryKeyring(n);
    return entry.createIdentity();
  }
```

This doesn't allow us to  specify the path to use for hd keys, or to import a private key from another key store. It also is not the cleanest way to specify the algorithm. I mean, n=0 could be ed25519 and n=1 secp256k1, but that is not so clear.

We should make it easy to either import existing private keys, or to do something like "Create a new secp256k1 key with path 44'/1'/2'/0/x where x is one higher that the last value you generated with that prefix".

Let's figure out the nicest API for the user, then work on implementing it. As all the pieces are there, it shouldn't be too much hard crypto work, just API polish.

ethanfrey:
What are the types of keyring entries we would want to support??

* HD generation in software (given bip39 mneumonic)
* HD generation in hardware (requires connection, mneumonic never exposed)
* Import private keys from other software
* Socket connection to process implementing one of the above (eg. extension connect to gateway app using node-ledger)

All should support both ed25519 and secp256k1

webmaster128:
In https://github.com/iov-one/iov-core/blob/master/docs/KeyBase.md#address-architecture we define two types of address schemes, the simple linear Simple Addresses and the 4-dimensional Bip-44-like Extended Addresses.

My last information was that arbitrary keypairs like in Ed25519KeyringEntry should not be supported.

Should we start by defining the use cases we want to support?

Isabello:
From: https://github.com/iov-one/iov-core/blob/master/docs/KeyBase.md#complete-object-definition

We are missing an easy way to use BIP44/32 paths
```
 "PublicIdentities": [
          "PublicIdentity": {
            "address": "0x6806ea1d9b2eb59DAc7fdcdf28bf8d5a12AD84Bc",
            "label": "Wells Fargo",
            "publicKey": {
[...]

PR #304, opened 2018-08-28
Refine keyring api

ethanfrey:
This is work to specify a desired API related to issue #302 

So far, I just have a first draft in https://github.com/iov-one/iov-core/blob/refine-keyring-api/packages/iov-keycontrol/src/proposal.ts

Please add comments, or make another PR if there are large changes, to better demonstrate with code.

### DIFF
--- a/packages/iov-keycontrol/src/proposal.ts
+++ b/packages/iov-keycontrol/src/proposal.ts
@@ -61,14 +61,14 @@ export interface KeyringEntry<T extends PublicIdentity = LocalIdentity> {
 }
 
 export const enum KeyringType {
-  HD = "HD",
+  Bip44Compatible = "Bip44",
   // Need a better name here, for a bunch of keys with no relation...
   // either randomly generated, imported, etc....
   Mixed = "Mixed",
 }
 
 export interface HDKeyringEntry extends KeyringEntry<LocalHDIdentity> {
-  readonly kind: KeyringType.HD;
+  readonly kind: KeyringType.Bip44Compatible;
 
   // generate a new HD keyring identity with a given path
   // a keyring entry may support many possible paths, or only one.
--- a/packages/iov-keycontrol/src/proposal.ts
+++ b/packages/iov-keycontrol/src/proposal.ts
@@ -74,15 +74,14 @@ export interface HDKeyringEntry extends KeyringEntry<LocalHDIdentity> {
   // a keyring entry may support many possible paths, or only one.
   // if path or algorithm is not supported by the implementation,
   // it will throw an error
-  //
-  // if autoincrement is true, it will append an auto-incrementing counter
-  // to the path as first unused space.
-  // eg. if path is 44'/13'/0'/0 and auto-increment is true, it will look
-  // see 44'/13'/0'/0/{0,1,2,3} all exist and create with path 44'/13'/0'/0/4
   readonly createIdentityWithPathPrefix: (
     algo: Algorithm,
-    path: ReadonlyArray<Slip0010RawIndex>,
-    autoincrement: boolean,
+    // purpose should be hard-coded as 44' ?
+    purpose: Slip0010RawIndex,
+    coinType: Slip0010RawIndex,
+    account: Slip0010RawIndex,
+    change: Slip0010RawIndex,
+    addressIndex: Slip0010RawIndex,
   ) => Promise<LocalHDIdentity>;
 }
 
--- a/packages/iov-keycontrol/src/proposal.ts
+++ b/packages/iov-keycontrol/src/proposal.ts
@@ -9,7 +9,14 @@ Once approved, this should be a template to update the other files in
 this package.
 */
 
-import { PrehashType, SignableBytes } from "@iov/bcp-types";
+import {
+  Nonce,
+  PrehashType,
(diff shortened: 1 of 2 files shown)


packages/iov-keycontrol/src/proposal.ts
```diff
@@ -0,0 +1,142 @@
...
+  readonly createTransactionSignature: (
+    identity: PublicIdentity,
+    transactionBytes: SignableBytes,
+    prehash: PrehashType,
+    chainId: ChainId,
+  ) => Promise<SignatureBytes>;
+}
+
+export const enum KeyringType {
+  HD = "HD",
```
Isabello: Rename as BIP44Compatible

packages/iov-keycontrol/src/proposal.ts
```diff
@@ -74,15 +74,14 @@ export interface HDKeyringEntry extends KeyringEntry<LocalHDIdentity> {
...
-  //
-  // if autoincrement is true, it will append an auto-incrementing counter
-  // to the path as first unused space.
-  // eg. if path is 44'/13'/0'/0 and auto-increment is true, it will look
-  // see 44'/13'/0'/0/{0,1,2,3} all exist and create with path 44'/13'/0'/0/4
   readonly createIdentityWithPathPrefix: (
     algo: Algorithm,
-    path: ReadonlyArray<Slip0010RawIndex>,
-    autoincrement: boolean,
+    // purpose should be hard-coded as 44' ?
```
Isabello: Yes.

Isabello: We can default this in functions before calling it, though. As we probably need to be flexible in the event that something changes for purpose in the future

webmaster128:
I like most of "Represent bip44 as wrapper over bip32 functionality".

One thing I noticed: if our "bip32" implementation supports curves other than secp256k1 it should be called "slip10", which is a superset of bip32.

Separation would be even better if the bip44 code is only responsible to create paths to be used, e.g.

```
const path = bip44path(hardened(1), hardened(2), 0, 4); // user input
const identity = entry.createSlip10Identity(path);
const nextPath = bip44IncrementAddress(identity.path);
// nextPath is now 44', 1', 2', 0, 5
const nextIdentity = entry.createSlip10Identity(nextPath);
```

packages/iov-keycontrol/src/proposal.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// some utilities like clone, serialize removed for simplicity just
+// to focus on key management api.
+// this is good enough for signing, for creating keys, you need
+// a refinement (HDKeyringEntry or MixedKeyringEntry)
+export interface KeyringEntry<T extends PublicIdentity = LocalIdentity> {
+  // these store labels locally outside of the key material
+  // do we want Promises for the case of proxy connections?
+  readonly setLabel: (label: string | undefined) => void;
```
rudi-cilibrasi: seems like  this should be `setKeyringLabel` to more clearly and in parallel separate for the average user from `setIdentityLabel` because these two concepts are very easy to confuse and not so clear for anybody yet.

packages/iov-keycontrol/src/proposal.ts
```diff
@@ -0,0 +1,176 @@
...
+  readonly setIdentityLabel: (identity: PublicIdentity, label: string | undefined) => void;
+
+  // all state observation should be based on streams (ValueAndUpdates)
+  readonly getLabel: () => ValueAndUpdates<string | undefined>;
```
rudi-cilibrasi: suggest `getKeyringLabel` here to increase learnability

rudi-cilibrasi: or maybe `getKeyringEntryLabel` ?

packages/iov-keycontrol/src/proposal.ts
```diff
@@ -0,0 +1,176 @@
...
+  // for changes on their state.
+  //
+  // Note that this assumes the caller is trusted as we expose the
+  // full entries (including ability to create identities).
+  //
+  // Also, it will not work well over proxies as it stands (which
+  // should be fixed shortly).
+  readonly getEntries: () => ValueAndUpdates<ReadonlyArray<KeyringEntry>>;
+
+  // setEntryLabel should be called on the Entry itself
```
rudi-cilibrasi: i think this name is clearer than the one in use (`setLabel`) but noticed that specific name also doesn't appear in the code so am confused by it.

ethanfrey: proposal.ts is just a syntax-correct suggestion of the api, no real code.
once the proposal is accepted, then another pr would be to update the code to match

packages/iov-keycontrol/src/proposal.ts
webmaster128: I don't think KeyringEntry.getLabel should be renamed to KeyringEntry.getKeyringEntryLabel. It is the simple label for the instance you are working with (a keyring entry). If there is confusion about the classes that are used, this should be addresses by documentation, examples and the type system rather than repetition.

ethanfrey:
This is made obsolete from all the other refactoring in the 0.7.0 release

Issue #301, opened 2018-08-28
Improve separation of concerns, public APIs


ethanfrey:
This is a mega-issue to work on redesigning how to expose the query/signing functionality in a safe, flexible, extensible way. We have a working architecture, and it is not used in a UI, but currently all code is in the same process. This can work for eg. an electron app, but if we wish to work in a browser, we need to the key management into an extension or native app.

A number of related issues have arisen, notably #222 and #223 which were a first attempt to define separation, and #291. Ledger support in #256 also seems to require placing the key management out of process from any chrome extension.

This issue attempts to close many of the referenced issues above with a clean and extensible architecture. The design work will be done in a PR for comments and collaboration. The main thought is to separate into three components (which can be in the same process or different ones).

* **Untrusted UI / App** - this can be a webapp, etc. where we would like to be correct, but cannot forge transactions if compromised. This part can then iterate quickly, import plugins, etc. This allows a very flexible "reader" API for use here.
* **Trusted UI / Signer** - this is a signed, downloaded executable that talks to the *Untrusted UI* and provides information for the user to approve transactions before signing them. It also provides an interface to create new accounts and expose identities to the *Untrusted API*. This part must be able to communicate with the blockchain to query nonces, submit transactions, check codecs, etc...
* **Keyring** - the keyring provides an interface to signing with key material. It trusts the app that connects to it (*Trusted UI*), meaning it will sign whatever it is requested to sign, but it will never leak private key material across this boundary. This should usually be on the same machine as the *Trusted UI* and may run as a background daemon.

We also assume that there may be a 1:N / N:N relationship at any of these stages (to be implemented in the future, but at least included in the design).  That is, an app may call (one of) multiple signers, something like https://walletconnect.org/. Also, a Signer may need multiple key stores to back it, such as a chrome extension that stores software keys internally, but calls out to a local node app to talk to the ledger, as extensions cannot talk to a ledger.... 

- [x] Accepted PR for new architecture design (markdown)
- [x] Accepted PR implementing new architecture
- [ ] Implement some examples for out-of-process use case.

ethanfrey:
Hi @lazaridiscom thanks for the comments.

Yeah, we wanted a library for our blockchain providing metamask-like functionality, as it is the most developer-friendly tooling I have seen. I read through metamask code first to see if we could fork it easily. Turns out it is quite tied to web3 all the way down to the signing layer, so we built this from the ground up. But I did grab some terms like Keyring from the metamask project.

[...]

PR #305, opened 2018-08-28
Start working out how a keyring proxy would look like

ethanfrey:
Just some toy code / strawman to help figure out good interfaces for #301 

This clearly doesn't work, but should give an idea of the direction I am thinking to go to.
It would be nice to get some feedback on the general design, and better yet, suggestions of better ways to organize and cut it.


ethanfrey:
This clearly is not fully working, and should be pulled out into another package and much more...

But I would love some quick feedback by @webmaster128 on the basic `message.ts`/`connection.ts`/`proxy.ts` design I wrote down.

I would refine this, make it it's own package, and then make much cleaner wrappers, but I figure this could be a nice way to allow many possible transports. As long as we can wrap them with some message stream interface, we get all the higher level functionality "for free" (implemented once for all connection types - extensions messages, websockets, unix pipes, etc....)

packages/iov-keycontrol/src/connection.ts
```diff
@@ -0,0 +1,232 @@
...
+      method: "subscribe",
+      params: {
+        query,
+        subscriptionId,
+      },
+    };
+    this.connection.send(sub);
+    const producer = new SendProducer<Event>();
+    // TODO: somehow store this to unsubscribe if subscription request
+    // above returns an error
```
webmaster128: Or make subscribe async, returning `Promise<Stream<Event>>`. This guarantees a `Stream<Event>` in case of success and a rejection in case of error. No complicated error handling in different places.

packages/iov-keycontrol/types/connection.d.ts
```diff
@@ -0,0 +1,30 @@
+import { Stream } from "xstream";
```
webmaster128: I think it is time to stop checking-in those types. Never used then for anything in weeks. However, this is independent of this PR

packages/iov-keycontrol/src/connection.ts
ethanfrey: `Promise<Stream<Event>>` seems like a beast, and I bet anyone using pure js will be bit by it, but it does make sense. The promise will reject on connection error (bad query), the stream on later error. I was wondering about using the stream error to signal bad query string.

Thanks for the suggestion.

packages/iov-keycontrol/types/connection.d.ts
ethanfrey: Yes, when you are back, let's take a few days to revisit the build processes, etc...
Not sure why the special `types` dir is needed outside of the `build` dir. I know there was an issue earlier, but not sure what.

webmaster128:
Closing as mainly outdated

ethanfrey:
Yes, outdated. Hope the code was useful to inspire some of the current design.

I like the current rpc implementation quite a bit

Issue #303, opened 2018-08-28
Run local faucet for @iov/faucet unit tests


ethanfrey:
Unit tests for this package run against friendnet, which seem much more like integration tests. I would prefer to pull up a local faucet in docker container, like we run local blockchains for CI. This is not urgent, but would make CI more reliable once friendnet has traffic.

https://github.com/iov-one/iov-core/pull/300#discussion_r213334401

webmaster128:
This is important for 2 reasons

* for every CI build, we use the faucet 8 times, burning a lot of tokens
* we're already having CI problems because of _Mismatch 257 != 258: Invalid sequence number_

webmaster128:
Current draft is here: https://github.com/iov-one/iov-core/pull/312

Now I am blocked by creating the `pk` file (https://github.com/iov-one/tools/issues/74)

webmaster128:
No further work will be put into the BovFaucet implemented in golang. We use the installation at https://faucet.friendnet-slow.iov.one/faucet until BovFaucet is removed in 0.9

PR #306, opened 2018-08-29
Add browsertests for faucets

webmaster128:
this currently fails because of https://github.com/iov-one/tools/issues/71

### DIFF
--- a/packages/iov-faucets/karma.conf.js
+++ b/packages/iov-faucets/karma.conf.js
@@ -1,0 +1,50 @@
+module.exports = function(config) {
+  config.set({
+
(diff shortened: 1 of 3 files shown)


ethanfrey:
Maybe this works now that lehajam deployed cors fix? Would be a great to have. :)

webmaster128:
Works in general but I want to do https://github.com/iov-one/iov-core/issues/303 first because otherwise we get a red CI all the time

webmaster128:
Depends on #312

webmaster128:
Depends on #474

webmaster128:
We can do that now. We just need to disable browser tests for the BovFaucet because it cannot handle many requests

PR #308, opened 2018-08-30
[iov-crypto] Fix keypair encoding bug in secp256k1

webmaster128:


ethanfrey:
Ah, yeah, the compressed secp format. I've seen there were different formats, and the compressed is like 63-67 bytes, but never looked deeper. Glad you did this update.

Issue #294, opened 2018-08-24
Add secp256k1 keyring entry


ethanfrey:
Ideally two classes to mirror `Ed25519HdKeyringEntry` and `Ed25519SimpleAddressKeyringEntry` from `iov-keycontrol/src/keyring-entries`.

I think secp support can be included directly in keycontrol and not as a external plugin (like ledger), as the implementation code is all in `@iov/crypto` which we already import.

This is a requirement to allow `iov-core` to work with ethereum based chains as well as the cosmos hub.

webmaster128:
We can easily generalize Ed25519HdKeyringEntry to Slip10KeyringEntry, allowing derived classes to use both curves. I think this should be step 1 for this ticket

webmaster128:
I am working on https://github.com/iov-one/iov-core/issues/294#issuecomment-416971109 now

ethanfrey:
Sounds good. That is also where we want to do, support multiple algorithms in one keyring

PR #310, opened 2018-08-30
[iov-keycontrol] Generalize to Slip10KeyringEntry

webmaster128:
This changes Ed25519HdKeyringEntry -> Slip10KeyringEntry, adding the ability to use ed25519 and secp256k1 within the same base class.

Slip10KeyringEntry is not meant to be used directly, so a secp256k1 equivalent of Ed25519SimpleAddressKeyringEntry is needed to utilize the new functionality.

Based on #308 

See #294


CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 ## 0.5.0
 
 * @iov/crypto: Rename all `Slip0010*` symbols to `Slip10*`
+* @iov/crypto: Fix keypair representation of Secp256k1.makeKeypair
+
+Breaking changes
+
+* Due to multi curve support in keyring entries, UserProfiles stored with
```
ethanfrey: Every breaking change breaks all keys :)
Good enough, we just need to be careful when we have actual users...

packages/iov-keycontrol/src/keyring-entries/ed25519simpleaddress.ts
```diff
@@ -1,16 +1,19 @@
...
 import { KeyringEntryImplementationIdString, LocalIdentity } from "../keyring";
-import { Ed25519HdKeyringEntry } from "./ed25519hd";
+import { Slip10KeyringEntry } from "./slip10";
 
-export class Ed25519SimpleAddressKeyringEntry extends Ed25519HdKeyringEntry {
+export class Ed25519SimpleAddressKeyringEntry extends Slip10KeyringEntry {
   // simple wrappers to cast return type
   public static fromEntropy(bip39Entropy: Uint8Array): Ed25519SimpleAddressKeyringEntry {
-    return super.fromEntropy(bip39Entropy) as Ed25519SimpleAddressKeyringEntry;
+    return super.fromEntropyWithCurve(Slip10Curve.Ed25519, bip39Entropy) as Ed25519SimpleAddressKeyringEntry;
```
ethanfrey: :+1:

packages/iov-keycontrol/src/keyring-entries/slip10.spec.ts
```diff
@@ -1,39 +1,55 @@
...
+      "secret": "rhythm they leave position crowd cart pilot student razor indoor gesture thrive",
+      "curve": "ed25519 seed",
+      "identities": []
+    }
+    ` as KeyringEntrySerializationString;
+
+  const emptySecp256k1Entry = `
+    {
+      "secret": "rhythm they leave position crowd cart pilot student razor indoor gesture thrive",
+      "curve": "Bitcoin seed",
```
ethanfrey: nice, add curve to serialization.

packages/iov-keycontrol/src/keyring-entries/slip10.spec.ts
```diff
@@ -71,8 +87,77 @@ describe("Ed25519HdKeyringEntry", () => {
...
+      expect(identities[2].pubkey.algo).toEqual(Algorithm.SECP256K1);
+      expect(identities[2].pubkey.data).toEqual(newIdentity3.pubkey.data);
+    }
+  });
+
+  it("generates secp256k1 keys compatible to other tools", async () => {
+    // Test data generated by the BIP44 tool of
+    // https://iancoleman.io/bip39/#english
+
+    const entry = Slip10KeyringEntry.fromMnemonicWithCurve(Slip10Curve.Secp256k1, "mushroom faint dumb venture million true skull grab pitch mesh share tortoise");
```
ethanfrey: Okay, so the same class can be used, just exposes one curve/mneumonic pair.
If I wanted the same mneumonic with two different curves, I would create two instances of this class, correct?

This is good. All the hard coding is done once, we can just wrap it up different ways.

packages/iov-keycontrol/src/keyring-entries/slip10.spec.ts
```diff
@@ -71,8 +87,77 @@ describe("Ed25519HdKeyringEntry", () => {
...
+    // $ node
+    // > Keypair.fromPublicKey("GDRXE2BQUC3AZNPVFSCEZ76NJ3WWL25FYFK6RGZGIEKWE4SOOHSUJUJ6").rawPublicKey().toString("hex")
+    const entry = Slip10KeyringEntry.fromMnemonicWithCurve(Slip10Curve.Ed25519, "illness spike retreat truth genius clock brain pass fit cave bargain toe");
+
+    // m/44'/148'/0'
+    const address0 = await entry.createIdentityWithPath([Slip10RawIndex.hardened(44), Slip10RawIndex.hardened(148), Slip10RawIndex.hardened(0)]);
+    expect(address0.pubkey.data).toEqual(fromHex("e3726830a0b60cb5f52c844cffcd4eed65eba5c155e89b26411562724e71e544"));
+
+    // m/44'/148'/1'
+    const address1 = await entry.createIdentityWithPath([Slip10RawIndex.hardened(44), Slip10RawIndex.hardened(148), Slip10RawIndex.hardened(1)]);
```
ethanfrey: Yes, like the bip32 curve idea from the proposal sketch earlier.
I think we can just keep this name.

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -96,14 +115,14 @@ export class Ed25519HdKeyringEntry implements KeyringEntry {
     for (const record of decodedData.identities) {
       const identity: LocalIdentity = {
         pubkey: {
-          algo: Ed25519HdKeyringEntry.algorithmFromString(record.localIdentity.pubkey.algo),
+          algo: Slip10KeyringEntry.algorithmFromString(record.localIdentity.pubkey.algo),
```
ethanfrey: This seems to allow multiple algorithms in one keyring, even if the above code uses only one curve.... Either way is fine, but if we only do one, I would read algorithm from top level.

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -117,22 +136,40 @@ export class Ed25519HdKeyringEntry implements KeyringEntry {
...
   public async createIdentity(): Promise<LocalIdentity> {
-    throw new Error("Ed25519HdKeyringEntry.createIdentity must not be called directly. Use derived type.");
+    throw new Error("Slip10KeyringEntry.createIdentity must not be called directly. Use derived type.");
   }
 
   public async createIdentityWithPath(path: ReadonlyArray<Slip10RawIndex>): Promise<LocalIdentity> {
     const seed = await Bip39.mnemonicToSeed(this.secret);
-    const derivationResult = Slip10.derivePath(Slip10Curve.Ed25519, seed, path);
-    const keypair = await Ed25519.makeKeypair(derivationResult.privkey);
+    const derivationResult = Slip10.derivePath(this.curve, seed, path);
```
ethanfrey: You make it look so easy :)

packages/iov-keycontrol/src/keyring-entries/slip10.spec.ts
webmaster128: Correct, every Slip10KeyringEntry has exactly one fixed mnemonic and exactly one fixed curve type

packages/iov-keycontrol/src/keyring-entries/slip10.ts
webmaster128: This is because data is duplicated. Each identity has an algorithm identifier, that can be used when the identity is copied out of the entry. This would allow us to create keyring entries in the future, that can produce identities with different curves. This is not necessarily a great thing but the possibility shows that our abstraction is good.

What we could do here in Slip10KeyringEntry is to verify that entry curve and identity algorithm match via an assertion. This would protect against some super crazy programming errors. However, we already cover that created identities are using the right algorithm.

packages/iov-keycontrol/src/keyring-entries/slip10.ts
webmaster128: :)

packages/iov-keycontrol/src/keyring-entries/slip10.ts
webmaster128: https://github.com/iov-one/iov-core/pull/310/commits/4af9a240da36482515cd80faba44c8666a8bf9f3

PR #312, opened 2018-08-30
Use local faucet in CI

webmaster128:
this is how far I got. Feel free to continue this work.


ethanfrey:
I can take a swing at the pk file if you wish... at least make the output match current faucet style.

ethanfrey:
Okay, there is the idea of just reimplementing the faucet in ts on top of iov-core. This came from backend guys.

If that is done, we can easily test in one process in CI, no docker needed. I'll just leave this pending until then.

webmaster128:
superseded by #472

PR #317, opened 2018-08-31
Expose subscriptions in @iov/bns.Client

ethanfrey:
We can subscribe to transactions with tendermint-rpc. Now, let us expose this with bcp types as Streams not just promises (to mirror the getXyz methods)

- [x] watch account balance (like getAccount)
- [x] watch account nonce? (like getNonce)
- [x] watch tx (parallel to searchTx, with tags)

Do we also need for tickers? blocks???
And we should add this to IovReader interface?

### DIFF
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -1,4 +1,6 @@
-import { ChainId, PostableBytes, PublicKeyBundle, TxQuery } from "@iov/tendermint-types";
+import { Stream } from "xstream";
+
+import { ChainId, PostableBytes, PublicKeyBundle, Tag, TxQuery } from "@iov/tendermint-types";
 
 import { Address, SignedTransaction } from "./signables";
 import { Nonce, TokenTicker } from "./transactions";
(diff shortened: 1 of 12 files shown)


Isabello:
Tickers probably dont need a stream. 
Nonce yes
Tx yes

All of this would be proper to put into reader, as the rest of account sampling stuff is in reader

packages/iov-bns/src/stream.ts
```diff
@@ -0,0 +1,80 @@
+import { InternalListener, InternalProducer, Stream } from "xstream";
+
+// streamPromise takes a Promsie that returns an array
```
rudi-cilibrasi: Promsie

packages/iov-bns/src/stream.ts
```diff
@@ -0,0 +1,80 @@
...
+// and emits one event for each element of the array
+export function streamPromise<T>(promise: Promise<ReadonlyArray<T>>): Stream<T> {
+  return new Stream<T>(new StreamPromise<T>(promise));
+}
+
+export async function readIntoArray<T>(stream: Stream<T>): Promise<ReadonlyArray<T>> {
+  const concat = (acc: ReadonlyArray<T>, val: T) => acc.concat(val);
+  const result = stream.fold(concat, []).last();
+  return new Promise<ReadonlyArray<T>>((resolve, reject) => {
```
rudi-cilibrasi: this seems weird to me.  when you call `addListener` it suggests that there can be zero or more calls later. importantly, two or more calls seems likely. yet, the resolve/reject parameters passed in would only be referring to the same promise instance over and over. so i think the `Promise` property that they only ever resolve or reject once and not again means that this is a first-event-only type filter that just "eats" events after the first? just guessing from inspection i have not tested it please let me know if i misunderstood it.

packages/iov-bns/src/stream.ts
```diff
@@ -0,0 +1,80 @@
...
+export async function countStream<T>(stream: Stream<T>): Promise<number> {
+  // tslint:disable-next-line:no-let
+  let count = 0;
+  return new Promise<number>((resolve, reject) => {
+    stream.addListener({
+      next: () => {
+        count++;
+      },
+      error: (err: any) => reject(err),
+      complete: () => resolve(count),
```
rudi-cilibrasi: feels like this can only count streams (in the future) of length 0 or 1?  i noticed in the test that counts events indeed one is historical... so an easy way to test the hypothesis is to make 2 events in the future.  if that is already tested and i missed it please let me know otherwise i think this is another case where it looks like it eats all future events after the first to me.

packages/iov-bns/src/stream.ts
ethanfrey: The `.last()` in result means it will only return one when it completes.
If no elements ever get sent, then the `last` output of `fold` will be the initial value, `[]`

ethanfrey: Weird that last returns a stream of one event, but I guess they like consistency. Thus my little helper functions to marshal into and out of Promises

packages/iov-bns/src/stream.ts
```diff
@@ -0,0 +1,80 @@
...
+          }
+        },
+        (e: any) => {
+          out._e(e);
+        },
+      )
+      .then(
+        () => 0,
+        (err: any) => {
+          setTimeout(() => {
```
rudi-cilibrasi: is this like `nextTick`?  I was wondering if it would be safer with a small number there instead of none to avoid a lockup or busy loop perhaps?  this part of the code is a big mystery to me... but i can't really tell if it's helping or not.

ethanfrey: this part is copy-paste from xstream library, note the comment above the function.
works and tested there.

ethanfrey: may be nice to refactor, but "if it ain't broke, don't fix it"

packages/iov-bns/src/stream.ts
ethanfrey: A stream will call next() N times.
It will end with one call to either error or complete.

If there are 5 events, they will call next 5 times, then complete upon closing the stream (why i have an explicit disconnect at the end of the watch tests).

Test case here: https://github.com/iov-one/iov-core/blob/bns-subscriptions/packages/iov-bns/src/stream.spec.ts#L23-L29

packages/iov-bns/src/stream.ts
rudi-cilibrasi: i don't know what works means here

packages/iov-bns/src/stream.ts
rudi-cilibrasi: yeah i think you are right in this explanation you gave now that I reread the code @ethanfrey  however the problem is that the countStream test still seems to return `2` when i expect `3` in the other pr based on this one with a new event added. so until that is explained we can block this to figure out where the mistake is.

ethanfrey:
Sorry @rudi, it is your example code from #320 that is broken:

```
    expect(afterSearch.length).toEqual(3);

    // disconnect the client, so all the live streams complete,
    // and promises resolve
    await sleep(50);
    client.disconnect();
    // this should grab the tx before it started, as well as the one after
    expect(await countLive).toEqual(2);
```

the error was cuz you didn't update countLive to expect 3.

`liveTx` will return a stream of all transactions regardless of when it was called. before the first tx, after the last tx, in the middle... it just asserts the total number is the same as if we do a search at the end.
And it does. You're code is asserting that `liveTx` only got 2 of the 3 transactions, while it does get all of them. Error in test.

`listenTx` just gets from that point on, `searchTx` gets the history, `liveTx` gets it all.
maybe the confusion is due to the name `liveTx`, I would be happy for some name that captures this use.
But I think the comment on the method does explain it (and please add suggestions for how to improve that as well):

```
  // liveTx does a search and then subscribes to all future changes.
  // It returns a stream starting the array of all existing transactions
  // and then continuing with live feeds
  public liveTx(txQuery: TxQuery): Stream<ConfirmedTransaction> {
```

rudi-cilibrasi:
yeah i think you defended the correctness but not the clarity because it seems like 1-2 names are still a bit misleading

ethanfrey:
Yes, I will update the helpers and the test code, as it is both unclear, and not a good example of how one would use the library (as tests should also serve as documentation).

Do you have any suggestions for improving names in the non-test part (searchTx/listenTx/liveTx, etc)?

packages/iov-bcp-types/src/bcp.ts
```diff
@@ -90,30 +93,42 @@ export interface IovReader {
...
+  readonly listenTx: (tags: ReadonlyArray<Tag>) => Stream<ConfirmedTransaction>;
+
+  // liveTx returns a stream for all historical transactions that match
+  // the query, along with all new transactions arriving from listenTx
+  readonly liveTx: (txQuery: TxQuery) => Stream<ConfirmedTransaction>;
 }
 
 export interface ConfirmedTransaction extends SignedTransaction {
-  readonly height: number; // the block it was write on
+  readonly height: number; // the block it was writen to
```
rudi-cilibrasi: `written`

PR #320, opened 2018-08-31
WIP bns-subscriptions

annalissac:


ethanfrey:
Test code in this pr was not correct. But the point was made that the tests I wrote did not properly convey the intention. I will update #317

PR #321, opened 2018-09-01
Fix node pregyp version so ledger-bns will build

Isabello:


Isabello:
Do a fresh clone and install it should work. Master branch fails.

ethanfrey:
Huh, test failure is due to https://github.com/yarnpkg/yarn/issues/3600 but only on osx... seems like downloading yarn install script every single time is not a good idea.

Suggestion just to rerun the build if it fails due to yarn not installing (different than `yarn install` failing). We should consider a better build process later. Let me know if you set up a better way in the frontend land

README.md
```diff
@@ -61,6 +61,17 @@ To build the documentation locally, run `yarn install && yarn build && yarn docs
 in this repository. This will generate a `./docs` directory in each package that you
 can browse locally to see API docs on the various packages.
 
+## FAQ For Potential Issues
```
ethanfrey: Thank you for the FAQ.
This will be very useful to jog our memory in a few months (or weeks)

PR #325, opened 2018-09-03
Pull out stream helpers into @iov/stream package

ethanfrey:
I wrote them to help with @iov/bns development, but they are far more general and it seems silly to import them there. Pull them into their own package to encourage reuse and to collect other stream helpers.

Maybe ValueAndUpdates belongs there as well?


webmaster128:
Nice, I was thinking in that direction too for some time now.

> Maybe ValueAndUpdates belongs there as well?

Definitely, yes

ethanfrey:
Cool, love to see a PR on for ValueAndUpdates... and we can use it a lot more. Very nice functionality

PR #327, opened 2018-09-04
Expose tx id

ethanfrey:
Let's show this in the results of IovReader (specifically bns.Client)

[x] Expose tx id in searchTx
[x] Expose tx id postTx
[x] Expose tx in listenTx (oof, tricky, had to calculate hash client side...)

This should help for blockchain explorer

### DIFF
--- a/packages/iov-tendermint-rpc/src/v0-20/hasher.ts
+++ b/packages/iov-tendermint-rpc/src/v0-20/hasher.ts
@@ -1,0 +1,10 @@
+// import Buffer from "buffer";
+import hasher from "ripemd160";
+
+import { PostableBytes, TxId } from "@iov/tendermint-types";
+
+// a bit ugly as this expects buffer, while the rest of code uint8array
+export const hashTx = (tx: PostableBytes): TxId => {
+  const hash = new hasher().update(Buffer.from(tx));
+  return Uint8Array.from(hash.digest()) as TxId;
+};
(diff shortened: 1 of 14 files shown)


packages/iov-tendermint-rpc/src/v0-20/hasher.ts
```diff
@@ -0,0 +1,22 @@
+// import Buffer from "buffer";
+import hasher from "ripemd160";
+
+import { PostableBytes, TxId } from "@iov/tendermint-types";
+
+// a bit ugly as this expects buffer, while the rest of code uint8array
```
webmaster128: Yeah, should be wrapped and tested by @iov/crypto

packages/iov-tendermint-rpc/src/v0-20/hasher.ts
```diff
@@ -0,0 +1,22 @@
...
+export const hashTx = (tx: PostableBytes): TxId => {
+  const hash = new hasher().update(Buffer.from(prefix(tx))).update(Buffer.from(tx));
+  return Uint8Array.from(hash.digest()) as TxId;
+};
+
+// prefix will create a varint prefix for any number < 2^14
+// should be a good enough estimation for tx hashing
+const prefix = (tx: Uint8Array): Uint8Array => {
```
webmaster128: This algorithm is defined by Tendermint, right? Otherwise we don't need it when we hash the whole transaction anyway. Can we add a reference to the source of this magic algorithm?

ethanfrey: The algorithm is in go-amino, which is not properly defined.... In v0.22 and beyond they use more normal hash, but for v0.21 and below, we need something like this.

Source:
https://github.com/tendermint/tendermint/blob/v0.20.0/types/tx.go#L19-L22
https://github.com/tendermint/tendermint/blob/v0.20.0/types/block.go#L648-L675

Which calls into `cdc.MarshalBinaryBare`, using go-aminos marshalling algorithm....

PR #332, opened 2018-09-12
Add identifiers to many exported objects

### LINKED ISSUE #331: add natural names to mnemonic phrases
rudi [1:22 PM]
On mnemonic natural names:

It's well known that the BIP standards specify a standard way to use English (or other language) words to encode a binary blob used as a sort of secret key material. In our iov-core api we used to use keyringIndex and have been adjusting the api to handle these key mnemonics differently lately. It strikes me that a fundamental weakness in the approach is that we don't provide a secure, safe, and convenient way to check mnemonic phrases for equality.

I propose we define the natural name of a BIP type mnemonic phrase to be something like triple SHA256 of the binary key it encodes.

Then we can use the hex encoded natural name of mnemonic phrases as safe keys in hashes stored in both
```purseNonSense```
as well as perhaps
```purseSense```
for the purpose of automatically updating balance and transaction histories for accounts that are owned by the user on wallet startup but not yet unlocked. In other words, we can securely store the hashes of the mnemonics safely as hash keys and consider them references to the "real" keys that can be unlocked (perhaps from the `purseSense` area only as needed) for signing.

Once we have natural names available (either as a utility function in iov-backend or if leaders prefer as a value-added feature in `iov-core`) for mnemonics then it becomes more convenient to think of groups of secrets (groups of mnemonic phrases) as _sets_ instead of _sequences_ of secrets. And that natural name allows us to deduplicate to avoid confusion for the user without adding additional security risk through the use of a one-way trapdoor function that makes the same security assumptions as many of our underlying signature schemes. (edited)

Importantly, knowing the _name_ of the secret is not the same as knowing the _secret itself_ in this naming scheme.

Ethan Frey [1:33 PM]
I like the way of adding an immutible id to a keyring entry that is unique, reproduceable by any other keyring
entry with the same secret, and doesn't leak info.
However, sometimes we do not have the mneumonic (eg. ledger).
I would propose something like the `sha256(publickey from bip32(mneumonic, <fixed path>))`.

Eg. calculate the public key at `0'/1'/2'` and hash it, leaking no info really, except allowing for uniqueness. And the computation doesn't even need access to the mneumonic
I would second a refinement of this proposal as an addition to iov-core

Rudi agrees that the enhancement suggested by Ethan is appropriate.

### PR DESCRIPTION
**Pull request by ethanfrey**
Resolved #331 

This is useful when a consumer wants to check for identity and de-duplicate indexes without needed understanding of the embedded types. It is also import for any object that is exposed by react in a list.

Add `.id: string` to the following objects:

- [x] LocalIdentity
- [x] KeyringEntry (and all implementations in iov-keycontrol)
- [x] Ledger-bns keyring entry
- [ ] UserProfile
- [ ] IovWriter (?)

After implementing the LocalIdentity and Keyring parts, I realize there is no clear way to provide a unique idenitity to a UserProfile.

If we add a KeyringEntry it must remain the same id.
However, if id is related to the secrets contained, then two UserProfiles both containing a KeyringEntry with id `x` should have the same id, right?

```
const a = new UserProfile();
const b = new UserProfile();

a.addKeyringEntry(entry);
expect(a.id).not.toEqual(b.id);
b.addKeyringEntry(entry);
expect(a.id).toEqual(b.id);
```

Or should I just give them all some pseudo-random id?
At least the slip10 based keyring entries and the local identities have a immuitible id tied to their data.

### DIFF
--- a/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
+++ b/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
@@ -1,3 +1,5 @@
+import PseudoRandom from "random-js";
+
 import { PrehashType, SignableBytes } from "@iov/bcp-types";
 import { Ed25519, Ed25519Keypair, Random } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
@@ -35,6 +37,8 @@ interface Ed25519KeyringEntrySerialization {
 }
 
 export class Ed25519KeyringEntry implements KeyringEntry {
+  private static readonly prng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().seed(12345678);
+
   private static identityId(identity: PublicIdentity): string {
     return identity.pubkey.algo + "|" + Encoding.toHex(identity.pubkey.data);
   }
@@ -196,8 +200,7 @@ export class Ed25519KeyringEntry implements KeyringEntry {
 
   private randomId(): string {
     // this can be pseudo-random, just used for internal book-keeping
-    const index = Math.random() * 1000000000;
-    // appends random 9-digit number
-    return "ed25519:" + Math.floor(index).toString(10);
+    const code = PseudoRandom.string()(Ed25519KeyringEntry.prng, 10);
+    return "ed25519:" + code;
   }
 }
--- a/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
+++ b/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
@@ -1,7 +1,7 @@
 import { PrehashType, SignableBytes } from "@iov/bcp-types";
 import { Ed25519, Ed25519Keypair, Random } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
-import { Algorithm, ChainId, PublicKeyBytes, SignatureBytes } from "@iov/tendermint-types";
+import { Algorithm, ChainId, PublicKeyBundle, PublicKeyBytes, SignatureBytes } from "@iov/tendermint-types";
 
 import {
   KeyringEntry,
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: packages/iov-keycontrol/src/keyring-entries/ed25519.ts [hunk 1]
```diff
@@ -177,4 +181,23 @@ export class Ed25519KeyringEntry implements KeyringEntry {
...
+    return {
+      pubkey,
+      label,
+      id: Ed25519KeyringEntry.identityId({ pubkey }),
+    };
+  }
+
+  private randomId(): string {
+    // this can be pseudo-random, just used for internal book-keeping
+    const index = Math.random() * 1000000000;
```
rudi-cilibrasi: this doesn't seem safe since `Math` seems like it would have a global seed so could easily be reseeded to accidentally create surprising keyring collisions e.g. in scientific simulations that reseed themselves. i would suggest we avoid the global random context and instead have a private random context that cannot be externally reseeded accidentally.

rudi-cilibrasi: can use `new` form without seed parameter to get automatic seeding https://github.com/davidbau/seedrandom

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -261,4 +280,29 @@ export class Slip10KeyringEntry implements KeyringEntry {
...
+      data: bytes,
+    };
+    return {
+      pubkey,
+      label,
+      id: Slip10KeyringEntry.identityId({ pubkey }),
+    };
+  }
+
+  // calculate id returns the tripple sha256 hash of the bip39 entropy as hex-string
```
webmaster128: secret has enough entropy to be brute-force resistant. So the triple hash is not necessary (and would not help). But it would be nice to have a use case identifier in the hash to avoid any kind of collisions with results in other applications. Using HMAC-SHA256 we can add something like "IOV entry ID" to the secret.

ethanfrey: sounds good.

packages/iov-keycontrol/src/keyring-entries/ed25519.ts
```diff
@@ -29,11 +31,14 @@ interface IdentitySerialization {
 }
 
 interface Ed25519KeyringEntrySerialization {
+  readonly id: string;
   readonly label: string | undefined;
   readonly identities: ReadonlyArray<IdentitySerialization>;
 }
 
 export class Ed25519KeyringEntry implements KeyringEntry {
+  private static readonly prng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().seed(12345678);
```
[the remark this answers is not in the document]
ethanfrey: good call, yes, please do adjust

PR #333, opened 2018-09-12
Add synchronous, isolated semi-pseudo-random generator

ethanfrey:
Crypto/random is secure but async. It is also overkill for identifiers for test cases.
Math.random is vulnerable to reseeding.
I didn't feel comfortable adding another dep https://github.com/davidbau/seedrandom especially as it globally modifies the Math variable.

Rudi's comment https://github.com/iov-one/iov-core/pull/332#pullrequestreview-154697853 and insistence on slack encouraged me to add something myself.

Hope this is okay....


rudi-cilibrasi:
i think it's better to use a solution that has received peer review for many years

https://www.npmjs.com/package/random-js

rudi-cilibrasi:
the problems with semirandom are: 1. non-standard api 2. not well documented 3. hard to find seeding (as it's not standard) 4. hard to believe in (not widely tested/reviewed) 5. hard to use elsewhere. if we include a non-offensive prng module from npm we will gain a lot more functionality in addition to points 1-4.

rudi-cilibrasi:
we also can reduce risk for our users by not making a(nother) unsafe prng ```For anything important, please use Random.getBytes() from ./libsodium.ts.```

rudi-cilibrasi:
credible docs here https://github.com/ckknight/random-js

rudi-cilibrasi:
`semi-random` is a real technical term also already but with different meaning i believe https://arxiv.org/abs/1805.02259

ethanfrey:
Closing due to popular demand.

webmaster128:
:+1: for using a lib here

PR #337, opened 2018-09-14
Force republish v0.5.3

ethanfrey:
I forgot to `yarn build` before `lerna publish` and pushed the old javascript, even when tagging v0.5.2 with the typescript. This now builds automatically on `yarn release` and I update packages to allow a proper v0.5.3 release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test": "lerna run test",
     "build": "lerna run build",
     "docs": "lerna run docs",
-    "release": "lerna publish",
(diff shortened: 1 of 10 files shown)


webmaster128:
An even safer command for releasing is documented here: https://github.com/iov-one/iov-core/wiki/Publishing#how-to-publish-to-npm This also ensures that no files are published that idle in the working directory for whatever reason (e.g. deleted .js files)

PR #340, opened 2018-09-18
Add @iov/cli module paths to REPL context

### LINKED ISSUE #328: Docs to run with @iov/cli don't work out of the box
Roman tried to install from npm and that didn't work, he had to install a few more packages manually.

Let's reproduce this on a clean vm (not dev machine) and verify what commands are needed and update documentation to make this process smoother for those not cloning the repo and building it.

Please assign to Ethan when there is a clear way to reproduce problem.

webmaster128:
Resolves  #328

### DIFF
--- a/packages/iov-cli/src/cli.ts
+++ b/packages/iov-cli/src/cli.ts
@@ -117,7 +117,9 @@ export const main = (originalArgs: string[]): void => {
     `;
   }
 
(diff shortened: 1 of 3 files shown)


webmaster128:
okay, found a solution :)

webmaster128:
Linked the issue this PR resolves. I think this is sufficient to get the idea.

ethanfrey:
Ahh, i missed that. Thanks.

webmaster128:
> Ahh, i missed that. Thanks.

I added it after the review. It was not there before

PR #341, opened 2018-09-18
Test global @iov/cli installation

webmaster128:


webmaster128:
Test reproducing the bug is integrated into #340

PR #345, opened 2018-09-19
Upgrade dependencies

webmaster128:


ethanfrey:
Maybe need to rebase and do again after the DefinitelyTyped change? Which also modified yarn.lock.

I give my approval as long as it is just yarn.lock and CI passes

PR #349, opened 2018-09-21
Refactor keyring entry ID generation

webmaster128:
Using a static seed, the sequence of IDs was always identical, leading to the same first ID for every new process. E.g.

```
// process 1
>> new Ed25519KeyringEntry().id
'ed25519:ZIFeXqRVH4'
>> new Ed25519KeyringEntry().id
'ed25519:VclMgX__bi'
>> new Ed25519KeyringEntry().id
'ed25519:NZ6HCT22pU'

// process 2
>> new Ed25519KeyringEntry().id
'ed25519:ZIFeXqRVH4'
>> new Ed25519KeyringEntry().id
'ed25519:VclMgX__bi'
>> new Ed25519KeyringEntry().id
'ed25519:NZ6HCT22pU'
```

This was fixed by using auto seeding (implemented as time in milliseconds + 16 calls of Math.random()).

Also, it was not possible before to create two different entries with the same mnemonic because both got the same ID. This led to an incomplete and misleading idea of equality. Now HD entries also get random IDs that are serialized and preserved when cloning and storing.

### DIFF
--- a/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
+++ b/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
@@ -39,7 +39,7 @@ interface Ed25519KeyringEntrySerialization {
 }
 
 export class Ed25519KeyringEntry implements KeyringEntry {
-  private static readonly prng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().seed(12345678);
+  private static readonly prng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().autoSeed();
 
   private static identityId(identity: PublicIdentity): LocalIdentityId {
     const id = identity.pubkey.algo + "|" + Encoding.toHex(identity.pubkey.data);
--- a/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
+++ b/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
@@ -72,12 +72,11 @@ export class Ed25519KeyringEntry implements KeyringEntry {
 
   constructor(data?: KeyringEntrySerializationString) {
     // tslint:disable-next-line:no-let
+    let id: KeyringEntryId;
+    // tslint:disable-next-line:no-let
     let label: string | undefined;
     const identities: LocalIdentity[] = [];
     const privkeys = new Map<string, Ed25519Keypair>();
-
-    // tslint:disable-next-line:no-let
-    let id = this.randomId();
 
     if (data) {
       const decodedData: Ed25519KeyringEntrySerialization = JSON.parse(data);
@@ -102,6 +101,8 @@ export class Ed25519KeyringEntry implements KeyringEntry {
         identities.push(identity);
         privkeys.set(identity.id, keypair);
       }
+    } else {
+      id = this.randomId();
     }
 
     this.identities = identities;
--- a/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
+++ b/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
@@ -39,7 +39,7 @@ interface Ed25519KeyringEntrySerialization {
 }
 
 export class Ed25519KeyringEntry implements KeyringEntry {
-  private static readonly prng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().autoSeed();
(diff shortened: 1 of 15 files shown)


ethanfrey:
I used Math.random first until I got criticism for it in the original PR, then modified it to use this local copy, seeded version.

packages/iov-keycontrol/src/keyring-entries/ed25519.ts
```diff
@@ -39,7 +39,13 @@ interface Ed25519KeyringEntrySerialization {
 }
 
 export class Ed25519KeyringEntry implements KeyringEntry {
-  private static readonly prng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().seed(12345678);
+  private static readonly idsPrng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().autoSeed();
```
ethanfrey: ahhh, didn't see the `autoSeed` method. I do like it

packages/iov-keycontrol/src/keyring-entries/ed25519.ts
```diff
@@ -39,7 +39,13 @@ interface Ed25519KeyringEntrySerialization {
...
+
+  private static generateId(): KeyringEntryId {
+    // this can be pseudo-random, just used for internal book-keeping
+    const code = PseudoRandom.string()(Ed25519KeyringEntry.idsPrng, 16);
```
ethanfrey: :+1:

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -76,6 +79,14 @@ export class Slip10KeyringEntry implements KeyringEntry {
     return new cls(JSON.stringify(data) as KeyringEntrySerializationString);
   }
 
+  private static readonly idsPrng: PseudoRandom.Engine = PseudoRandom.engines.mt19937().autoSeed();
+
+  private static generateId(): KeyringEntryId {
+    // this can be pseudo-random, just used for internal book-keeping
+    const code = PseudoRandom.string()(Slip10KeyringEntry.idsPrng, 16);
+    return `${code}` as KeyringEntryId;
```
ethanfrey: why ` ${}` ? this is different than implementation in ed25519.ts

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -68,6 +70,7 @@ export class Slip10KeyringEntry implements KeyringEntry {
     cls: Slip10KeyringEntryConstructor = Slip10KeyringEntry,
   ): Slip10KeyringEntry {
     const data: Slip10KeyringEntrySerialization = {
+      id: Slip10KeyringEntry.generateId(),
```
ethanfrey: ahh, here we generate, not in constructor.
a bit confusing at review... but i think it is correct

packages/iov-keycontrol/src/keyring-entries/slip10.ts
webmaster128: A leftover from before, will clean up

webmaster128:
> I used Math.random first until I got criticism for it in the original PR, then modified it to use this local copy, seeded version.

I think it is good to [not implement this ourselves](https://github.com/iov-one/iov-core/pull/333/files#diff-841d0e04889cd1bf442114b47a6a9e8dR13) but use the well known Mersenne-Twister.

Not trusting Math.random because of reseeding … well. Don't know how valid that is. When someone is able to access your global JS scope, you are doomed anyway.

packages/iov-keycontrol/src/keyring-entries/slip10.ts
webmaster128: Yeah, this is a bit unintuitive as we do construction via deserialization. One could clean this up by using a constructor that takes all the fields and have a separate deserializer.

rudi-cilibrasi:
it is not for protection against hackers it is to prevent accidental identical reseeding causing accidental collision in the same millisecond. simple point is that we don't need to use a global random number state and so choosing any (such as MT) that support instantiation is fine. that's the 'deficiency' of Math.random that it is not 'encapsulatable' but it is ony a deficiency in the sense of protection against future mistakes.

PR #350, opened 2018-09-21
Fix beginning and end in toMatch expressions

webmaster128:
Because `expect("öa~").toMatch(/[a-z]+/)` passes


ethanfrey:
Looks good to merge, what's up with Travis?

webmaster128:
> what's up with Travis

Nice, this change found a bug in the tests. I test base64 format against hex pattern

PR #355, opened 2018-09-25
[iov-encoding] create class Uint64

webmaster128:
this will be required to handle Lisk addresses


webmaster128:
Very good point. I was not expecting **unsigned** int64 to be covered by the Long type. I try to refactor my caller to use Long instead of Uint64 and if it does the job, throw this away.

webmaster128:
obsolete, thanks for feedback

PR #358, opened 2018-09-27
[iov-lisk] Implement Lisk connector

### LINKED ISSUE #292: Add lisk connector support

webmaster128:
This implements everything for doing stuff like

```ts
import { liskCodec, liskConnector, LiskKeyringEntry } from "@iov/lisk";

const entry = new LiskKeyringEntry();
const mainIdentity = await entry.createIdentity("oxygen fall sure lava energy veteran enroll frown question detail include maximum");

profile.addEntry(entry);

const writer = new IovWriter(profile);
await writer.addChain(liskConnector("https://testnet.lisk.io"));
const chainId = writer.chainIds()[0];
const reader = writer.reader(chainId);

const mainAddress = writer.keyToAddress(chainId, mainIdentity.pubkey);
console.log((await reader.getAccount({ address: mainAddress })).data[0].balance);

const recipientAddress = Encoding.toAscii("6076671634347365051L") as Address;

const sendTx: SendTx = {
  kind: TransactionKind.Send,
  chainId: chainId,
  signer: mainIdentity.pubkey,
  recipient: recipientAddress,
  memo: "We ❤️ developers – iov.one",
  amount: {
    whole: 1,
    fractional: 44550000,
    tokenTicker: "LSK" as TokenTicker,
  }
};

console.log("Writing to blockchain. This may take a while …");
await writer.signAndCommit(sendTx, 0);
console.log((await reader.getAccount({ address: recipientAddress })).data[0].balance);
```

Closes #292 

~Based on #356~

- [x] Extract nonce generation into function
- [x] Write transaction timestamp into nonce (in parseBytes())

### DIFF
--- a/packages/iov-bns/src/client.ts
+++ b/packages/iov-bns/src/client.ts
@@ -34,9 +34,10 @@ import * as codecImpl from "./codecimpl";
 import { InitData, Normalize } from "./normalize";
 import { Decoder, Keyed, Result } from "./types";
 
-// queryByAddress is a type guard to use in the account-based queries
-const queryByAddress = (query: BcpAccountQuery): query is BcpAddressQuery =>
-  (query as BcpAddressQuery).address !== undefined;
+// a type checker to use in the account-based queries
+function isAddressQuery(query: BcpAccountQuery): query is BcpAddressQuery {
+  return (query as BcpAddressQuery).address !== undefined;
+}
 
 // onChange returns a filter than only passes when the
 // value is different than the last one
@@ -142,7 +143,7 @@ export class Client implements IovReader {
   }
 
   public async getAccount(account: BcpAccountQuery): Promise<BcpQueryEnvelope<BcpAccount>> {
-    const res = queryByAddress(account)
+    const res = isAddressQuery(account)
       ? this.query("/wallets", account.address)
       : this.query("/wallets/name", Encoding.toAscii(account.name));
     const parser = parseMap(codecImpl.namecoin.Wallet, 5);
@@ -156,7 +157,7 @@ export class Client implements IovReader {
     // getAddress will do a lookup from name -> address if needed
     // make this an async function so easier to switch on return value
     const getAddress = async (): Promise<Uint8Array | undefined> => {
-      if (queryByAddress(account)) {
+      if (isAddressQuery(account)) {
         return account.address;
       }
       const addrRes = await this.getAccount(account);
@@ -244,7 +245,7 @@ export class Client implements IovReader {
   // watchAccount gets current balance and emits an update every time
   // it changes
   public watchAccount(account: BcpAccountQuery): Stream<BcpAccount | undefined> {
-    if (!queryByAddress(account)) {
+    if (!isAddressQuery(account)) {
       throw new Error("watchAccount requires an address, not name, to watch");
     }
     // oneAccount normalizes the BcpEnvelope to just get the
@@ -265,7 +266,7 @@ export class Client implements IovReader {
   // watchNonce gets current nonce and emits an update every time
   // it changes
   public watchNonce(account: BcpAccountQuery): Stream<BcpNonce | undefined> {
-    if (!queryByAddress(account)) {
+    if (!isAddressQuery(account)) {
       throw new Error("watchNonce requires an address, not name, to watch");
     }
     // oneNonce normalizes the BcpEnvelope to just get the
--- a/packages/iov-lisk/README.md
+++ b/packages/iov-lisk/README.md
@@ -4,11 +4,58 @@
 
 ## Getting started
 
+The primary way to use @iov/lisk is together with @iov/core. Alternatively,
+you can use @iov/lisk to create offline transactions which can be posted manually.
+
+All examples are made for use in @iov/cli and you may need to include some
+missing symbols if used in a different environment.
+
+### Using with @iov/core
+
+You can use @iov/lisk as an extension of @iov/core to interact with the
+Lisk blockchain as follows.
+
+```ts
+import { liskCodec, liskConnector, LiskKeyringEntry } from "@iov/lisk";
+
+const entry = new LiskKeyringEntry();
+const mainIdentity = await entry.createIdentity("oxygen fall sure lava energy veteran enroll frown question detail include maximum");
+
+const profile = new UserProfile();
+profile.addEntry(entry);
+
+const writer = new IovWriter(profile);
+await writer.addChain(liskConnector("https://testnet.lisk.io"));
+const chainId = writer.chainIds()[0];
+const reader = writer.reader(chainId);
+
+const mainAddress = writer.keyToAddress(chainId, mainIdentity.pubkey);
+console.log((await reader.getAccount({ address: mainAddress })).data[0].balance);
+
+const recipientAddress = Encoding.toAscii("6076671634347365051L") as Address;
+
+const sendTx: SendTx = {
+  kind: TransactionKind.Send,
+  chainId: chainId,
+  signer: mainIdentity.pubkey,
+  recipient: recipientAddress,
+  memo: "We ❤️ developers – iov.one",
+  amount: {
+    whole: 1,
+    fractional: 44550000,
+    tokenTicker: "LSK" as TokenTicker,
+  }
+};
+
+console.log("Writing to blockchain. This may take a while …");
+await writer.signAndCommit(sendTx, 0);
+console.log((await reader.getAccount({ address: recipientAddress })).data[0].balance);
+```
+
+### The manual way
+
 This is how you use `liskCodec` and `LiskKeyringEntry` to generate send transactions
 for Lisk manually, i.e. without the help of @iov/core.
-
-The following example is made for use in @iov/cli and you may need to include some
-missing symbols if used in a different environment.
 
 ```ts
 import { liskCodec, LiskKeyringEntry } from "@iov/lisk";
--- a/packages/iov-lisk/README.md
+++ b/packages/iov-lisk/README.md
@@ -27,7 +27,7 @@ const sendTx: SendTx = {
   recipient: recipientAddress,
   memo: "We ❤️ developers – iov.one",
   amount: {
-    whole: 2,
+    whole: 1,
     fractional: 44550000,
     tokenTicker: "LSK" as TokenTicker,
   }
--- a/packages/iov-lisk/README.md
+++ b/packages/iov-lisk/README.md
@@ -58,7 +58,7 @@ This is how you use `liskCodec` and `LiskKeyringEntry` to generate send transact
 for Lisk manually, i.e. without the help of @iov/core.
 
 ```ts
-import { liskCodec, LiskKeyringEntry } from "@iov/lisk";
+import { generateNonce, liskCodec, LiskKeyringEntry } from "@iov/lisk";
 
 const liskTestnet = "da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba" as ChainId;
 
@@ -80,8 +80,7 @@ const sendTx: SendTx = {
   }
 };
 
-// Encode creation timestamp into nonce
-const nonce = Long.fromNumber(Math.floor(Date.now() / 1000)) as Nonce;
+const nonce = generateNonce();
 const signingJob = liskCodec.bytesToSign(sendTx, nonce);
 const signature = await entry.createTransactionSignature(mainIdentity, signingJob.bytes, signingJob.prehashType, liskTestnet);
 
--- a/packages/iov-lisk/src/liskclient.spec.ts
+++ b/packages/iov-lisk/src/liskclient.spec.ts
@@ -59,7 +59,7 @@ describe("LiskClient", () => {
     expect(() => new LiskClient("http://[2001::0370:7344]/")).toThrowError(/invalid api url/i);
     expect(() => new LiskClient("http://[2001::0370:7344]:8080/")).toThrowError(/invalid api url/i);
 
-    //wrong path
+    // wrong path
     expect(() => new LiskClient("http://localhost/api")).toThrowError(/invalid api url/i);
   });
 
--- a/packages/iov-lisk/src/liskclient.spec.ts
+++ b/packages/iov-lisk/src/liskclient.spec.ts
@@ -1,3 +1,6 @@
+import { Address, BcpAccountQuery } from "@iov/bcp-types";
+import { Encoding } from "@iov/encoding";
+
 import { LiskClient } from "./liskclient";
 
 describe("LiskClient", () => {
@@ -24,5 +27,18 @@ describe("LiskClient", () => {
     const height = await client.height();
     expect(height).toBeGreaterThan(6000000);
     expect(height).toBeLessThan(8000000);
+  });
+
+  it("can get account", async () => {
+    // Generate dead target address:
+    // python3 -c 'import random; print("{}L".format(random.randint(0, 18446744073709551615)))'
+    const client = new LiskClient(base);
+    const query: BcpAccountQuery = { address: Encoding.toAscii("15683599531721344316L") as Address };
+    const account = await client.getAccount(query);
+    expect(account.data[0].address).toEqual(Encoding.toAscii("15683599531721344316L"));
+    expect(account.data[0].balance[0].tokenTicker).toEqual("LSK");
+    expect(account.data[0].balance[0].sigFigs).toEqual(8);
+    expect(account.data[0].balance[0].whole).toEqual(15);
+    expect(account.data[0].balance[0].fractional).toEqual(48687542);
   });
 });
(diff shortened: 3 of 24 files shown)


webmaster128:
Rebased onto master after merging of Lisk codec. Added a small TODO to the PR description of stuff I am currently aware of.

packages/iov-bns/src/client.ts
```diff
@@ -34,9 +34,10 @@ import * as codecImpl from "./codecimpl";
 import { InitData, Normalize } from "./normalize";
 import { Decoder, Keyed, Result } from "./types";
 
-// queryByAddress is a type guard to use in the account-based queries
-const queryByAddress = (query: BcpAccountQuery): query is BcpAddressQuery =>
-  (query as BcpAddressQuery).address !== undefined;
+// a type checker to use in the account-based queries
```
ethanfrey: yes, better name

packages/iov-lisk/README.md
```diff
@@ -4,14 +4,61 @@
...
-The following example is made for use in @iov/cli and you may need to include some
+All examples are made for use in @iov/cli and you may need to include some
 missing symbols if used in a different environment.
 
+### Using with @iov/core
+
+You can use @iov/lisk as an extension of @iov/core to interact with the
+Lisk blockchain as follows.
+
+```ts
```
ethanfrey: very nice that it works this way.
ideally also with other keyring entry, but cool to see the standard interfaces holding up.

packages/iov-lisk/src/liskclient.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const client = new LiskClient(base);
+    const chainId = await client.chainId();
+    expect(chainId).toEqual("da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba");
+  });
+
+  it("can get height", async () => {
+    const client = new LiskClient(base);
+    const height = await client.height();
+    expect(height).toBeGreaterThan(6000000);
+    expect(height).toBeLessThan(8000000);
```
ethanfrey: and in some months this will fail.... and we will have to update.
but fair enough, real testnets

packages/iov-lisk/src/liskclient.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  it("can get account", async () => {
+    // Generate dead target address:
+    // python3 -c 'import random; print("{}L".format(random.randint(0, 18446744073709551615)))'
+    const client = new LiskClient(base);
+    const query: BcpAccountQuery = { address: Encoding.toAscii("15683599531721344316L") as Address };
+    const account = await client.getAccount(query);
+    expect(account.data[0].address).toEqual(Encoding.toAscii("15683599531721344316L"));
+    expect(account.data[0].balance[0].tokenTicker).toEqual("LSK");
+    expect(account.data[0].balance[0].sigFigs).toEqual(8);
+    expect(account.data[0].balance[0].whole).toEqual(15);
```
ethanfrey: no one uses this account? this will not change?

packages/iov-lisk/src/liskclient.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    expect(account.data[0].balance[0].fractional).toEqual(48687542);
+  });
+
+  it("can get nonce", async () => {
+    const client = new LiskClient(base);
+    const query: BcpAccountQuery = { address: Encoding.toAscii("15683599531721344316L") as Address };
+    const nonce = await client.getNonce(query);
+
+    expect(nonce.data[0].address).toEqual(Encoding.toAscii("15683599531721344316L"));
+    // nonce is current timestamp +/- one second
```
ethanfrey: :+1:

packages/iov-lisk/src/liskclient.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      const bytesToPost = liskCodec.bytesToPost(signedTransaction);
+
+      const client = new LiskClient(base);
+      const result = await client.postTx(bytesToPost);
+      expect(result).toBeTruthy();
+      expect(result.metadata.height).toBeDefined();
+      expect(result.metadata.height).toBeGreaterThan(6000000);
+      expect(result.metadata.height).toBeLessThan(8000000);
+    },
+    40 * 1000,
```
ethanfrey: oof, this can take 40 seconds?

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+import { Parse } from "./parse";
+
+function isAddressQuery(query: BcpAccountQuery): query is BcpAddressQuery {
+  return (query as BcpAddressQuery).address !== undefined;
+}
+
+/**
+ * Encodes the current date and time as a nonce
+ */
+export function generateNonce(): Nonce {
```
ethanfrey: :+1:

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+  constructor(baseUrl: string) {
+    this.baseUrl = baseUrl;
+  }
+
+  public disconnect(): void {
+    // no-op
+  }
+
+  public async chainId(): Promise<ChainId> {
+    const url = this.baseUrl + "/api/node/constants";
```
ethanfrey: Just a note, I had some issues with tendermint rpc, when accidentally entering a baseurl that ended with `/`, so the combination looked like `http://host//path`.

Maybe handle that here as well, eg. trimming trailing '/' in the constructor. Or something safer

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  public async height(): Promise<number> {
+    const url = this.baseUrl + "/api/node/status";
+    const result = await axios.get(url);
+    const responseBody = result.data;
+    return responseBody.data.height;
+  }
+
+  public async postTx(bytes: PostableBytes): Promise<BcpTransactionResponse> {
+    const transactionId = JSON.parse(Encoding.fromUtf8(bytes)).id as string;
```
ethanfrey: nice check, that this is really a lisk transaction. I see JSON.parse throwing an error on lots of bad data

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+    if (!transactionId.match(/^[0-9]+$/)) {
+      throw new Error("Invalid transaction ID");
+    }
+
+    await axios.post(this.baseUrl + "/api/transactions", bytes, {
```
ethanfrey: question: did you test this in the browser as well (webpack/karma)?
axios should work there as well but i had some odd issue with post in tendermint-rpc and default to fetch in the browser. i should investigate further if this works, but after 3 hours, i just wanted to get it to work

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+      headers: {
+        "Content-Type": "application/json",
+      },
+    });
+
+    // Sleep some seconds to ensure transaction will be found.
+    // Sleep duration determined by trial and error. 15 seconds was not enough.
+    await new Promise(resolve => setTimeout(resolve, 20 * 1000));
```
ethanfrey: wait, so post always succeeds? just that it is in mempool, not that is it commited?
and you have to check the trasnactionid in the search to see if it made it...

hmmm... this code is fine for now, I can see a refinement coming up

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    return {
+      metadata: {
+        status: true, // commit failures should throw
+        height: height,
+      },
+      data: {
+        message: "",
+        txid: Encoding.toAscii(transactionId) as TxId,
```
ethanfrey: nice that these fields also have meaning for lisk

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+        result: transactionResultBytes || new Uint8Array([]),
+      },
+    };
+  }
+
+  public getTicker(_: TokenTicker): Promise<BcpQueryEnvelope<BcpTicker>> {
+    throw new Error("Not implemented");
+  }
+
+  public getAllTickers(): Promise<BcpQueryEnvelope<BcpTicker>> {
```
ethanfrey: Can be hardcoded to just return LSK?
But we should talk to isabella first on the design there.
So we can differentiate testnet from mainnet

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+      const result = await axios.get(url);
+      const responseBody = result.data;
+
+      const account: BcpAccount = {
+        address: address,
+        name: undefined,
+        balance: [
+          {
+            sigFigs: 8,
+            tokenName: undefined,
```
ethanfrey: please: `tokenName: "LSK" as TokenTicker`

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+            ...Parse.liskAmount(responseBody.data[0].balance),
+          },
+        ],
+      };
+
+      const wrapped: BcpQueryEnvelope<BcpAccount> = {
```
ethanfrey: I think there is a helper function `dummyEnvelope` that does it. Nice to reuse (and extend if needed)

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  public getNonce(query: BcpAccountQuery): Promise<BcpQueryEnvelope<BcpNonce>> {
+    if (isAddressQuery(query)) {
+      const address = query.address;
+
+      const nonce: BcpNonce = {
+        address: address,
+        // fake pubkey, we cannot always know this
```
ethanfrey: do we ever get to see the pubkey?
i mean, after they sign one transaction it should be visible somehow?

empty is fine if they never signed

packages/iov-lisk/src/liskclient.ts
```diff
@@ -0,0 +1,200 @@
...
+        },
+        data: [nonce],
+      };
+      return Promise.resolve(out);
+    } else {
+      throw new Error("Query type not supported");
+    }
+  }
+
+  public changeBalance(_: Address): Stream<number> {
```
ethanfrey: yup, let's do this later.
curious how this can work on lisk... gives a good test of the api and interfaces, maybe we need to adjust

packages/iov-lisk/src/liskcodec.ts
```diff
@@ -95,33 +94,31 @@ export const liskCodec: TxCodec = {
   parseBytes: (bytes: PostableBytes, chainId: ChainId): SignedTransaction => {
     const json = JSON.parse(Encoding.fromUtf8(bytes));
 
-    const fee = Int53.fromString(json.fee).asNumber();
-    const amount = Int53.fromString(json.amount).asNumber();
+    let kind: TransactionKind;
+    switch (json.type) {
```
ethanfrey: fair enough, we only need send for now

packages/iov-lisk/src/liskclient.ts
ethanfrey: hmmm... i guess this codec also needs to support lisk-compatible forks as well...
if the ticker is not in the api anywhere, maybe we hardcode a chainId->ticker mapping.
maybe later read that mapping in from bns

packages/iov-lisk/src/liskclient.spec.ts
webmaster128: Yeah, something like 10 months to go

packages/iov-lisk/src/liskclient.spec.ts
webmaster128: It cannot send as there is no keypair for that address. But it can receive. Pssst, don't tell anyone ;)

packages/iov-lisk/src/liskclient.spec.ts
webmaster128: If I want to have feedback like height, I need to sleep 20 seconds between submission and status query. So it should not take longer than ~22 seconds

packages/iov-lisk/src/liskclient.ts
webmaster128: `http://host//path` is not an issue and should work exactly like `http://host/path`. But we should validate that it is actually a baseUrl including the scheme and not a host like `testnet.list.io` or `testnet.list.io:8000`

packages/iov-lisk/src/liskclient.ts
webmaster128: I'm no having any issues with axios in browser tests for both @iov/lisk and @iov/faucets

packages/iov-lisk/src/liskclient.ts
webmaster128: It succeeds except for
- malformed requests
- invalid transaction signatures
- insufficient funds

No idea what other checks are done. But yeah, we cannot know if the transaction makes it into the blockchain at this point. At a blocktime of 10 seconds and 25 transactions per block, it can take minutes until the transaction is in the chain if the transaction pool of full.

packages/iov-lisk/src/liskclient.ts
webmaster128: `tokenName` != `tokenTicker`. tokenTicker is set in the line after that: `...Parse.liskAmount(responseBody.data[0].balance),`

I don't know what tokenName is supposed to be, but it is of type `string | undefined`, not `TokenTicker`

packages/iov-lisk/src/liskclient.ts
webmaster128: We can query the API and will get a pubkey if and only of the address did a transaction before. However, this adds another network query, will not work for the first outgoing transaction of an address and the result is thrown away anyway.

packages/iov-lisk/src/liskclient.ts
Isabello: Its the expanded full name. So BTC is a ticker, Bitcoin is a token name.

In this case, it should be `Lisk`

Isabello: This differentiation is needed because there are duplicate tickers in this world, so the name is a separate identifier that would be displayable to a user.

packages/iov-lisk/src/liskclient.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  it(
+    "can post transaction",
+    async () => {
+      pendingWithoutLongRunning();
+
+      const liskTestnet = "da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba" as ChainId;
+
+      const entry = new LiskKeyringEntry();
+      const mainIdentity = await entry.createIdentity(
+        "oxygen fall sure lava energy veteran enroll frown question detail include maximum",
```
Isabello: this could have been read in from an environment variable instead... oh well! :)

packages/iov-lisk/src/liskclient.ts
Isabello: there are many checks around vote list, multisig and a few other cases.

packages/iov-lisk/src/liskclient.ts
webmaster128: Oh I was too quick here. `http://host//path` is indeed 404 in the Lisk API. I'l normalize that in addition to the format check

packages/iov-lisk/src/liskclient.ts
webmaster128: fixed. And I think we can make the field non-optional. But this will go in the next PR https://github.com/iov-one/iov-core/pull/358/commits/1e82ca7b06747a9b22a6d46dd4f8fb1cd1aa8ebd

packages/iov-lisk/src/liskclient.ts
ethanfrey: btw, i moved `dummyEnvelope` into `bcp-types` with pr https://github.com/iov-one/iov-core/pull/361 to help reuse...

webmaster128:
There is an error in @iov/lisk on macOS+Safari and it looks like `lerna run` cuts some of the error message after

```
lerna ERR! yarn run test-safari exited 1 in '@iov/lisk'
lerna ERR! yarn run test-safari stdout:
```

Will debug locally when I have my mac available.

ethanfrey:
This may be with axios in the browser... if it is troublesome, you can switch axios/fetch based on the environment like: https://github.com/iov-one/iov-core/blob/master/packages/iov-tendermint-rpc/src/rpcclient.ts#L33-L55

(I added this due to failing CI in browser)

webmaster128:
The HTTP 409 is because the same transaction is created and submitted twice. This is because for some tests, two browsers or browser tabs run at the same time: https://screenshots.firefox.com/sFQyp5GVReEkraV8/travis-ci.com

The issue of multiple browser tests running at the same time exists on master as well but only causes problems here (and maybe in tendermint-rpc as well)

webmaster128:
Here we go – all systems green

Issue #342, opened 2018-09-19
Remove types directories from the repo


webmaster128:
In the JS world, there is a tendency to check-in distribution files into a repo.

Advantages:

* you can download a distribution file from github and copy it into your project
* you can npm install from a git URL
* the current state of the repo is what you upload to npm, no temporary files involved

Disadvantages:

* When shipping a distributions file that includes dependencies, it is very hard to comply to licenses since you need to keep track of and ship all dependencies' licenses
* Code and distribution files are out of sync most of the time
* Annoying "regenerate" commits pollute the git history
* Depending on the degree of bundling and minimization, small source changes can lead to huge changes in distribution files
* diffs (especially between two version tags) get very hard to review as most of the diff is distribution files
* Maintaining different branches with distribution files can never merge automatically

For those reasons we decided early to not check-in TS -> JS build output.

Using a monorepo, this lead to the issue that before building the entire monorepo, types of different packages are not available. For example, removing the types of @iov/encoding gives you an editor full of errors and "any" types when reviewing code in @iov/crypto:

![bildschirmfoto von 2018-09-19 10-49-16](https://user-images.githubusercontent.com/2603011/45742256-dafc0380-bbf9-11e8-88bf-f540a3b56c14.png)

This problem only exists for local dependencies, not for npm dependencies and is monorepo specific. Until now we worked around it by checking-in type definitions (which increase PR diffs and occasionally create build problems).

| Feature | no build files checked-in | .d.ts checked-in |  .d.ts and .js checked-in |
|------------|----------------|----------------|----------------|
| A. Types available for review, coding and building before full build | ✕ | ✓ | ✓ |
| B. Running tests before full build | ✕ | ✕ | ✓ |

A full build runs for 2:18 min on my machine, so it would bearable to require a build before any kind of work. If you assume that no work can be done without running tests, the full build step is mandatory anyway.

I leave it for discussion if feature A. is something we want to have.

rudi-cilibrasi:
This is why we standardized on `.mjs` as theoretically it can cache / store safely deps and precompiled artifacts before linking. Our expectation is that `.mjs` should keep getting faster than regular `.js` as project size increases due to the increased strictness. The increased strictness from the extension configuration constraints from `.mjs` should allow compilers to jam with static types better than what is possible in other extensions.

webmaster128:
[...]

PR #360, opened 2018-09-29
Compile on windows

### LINKED ISSUE #319: disallow cygwin on windows builds
we should check for it and disable it since it's subsumed and not worth maintaining nor confusing the other better way under windows using the official subsystem nowadays

ethanfrey:
Resolves #319 

Most of the scripts in package.json have errors on the normal window shell. Since this is also causing issues with my vscode integration, I finally set out to fix it. Inspired by the approach from rxjs: https://github.com/ReactiveX/rxjs/pull/1681/files I went ahead and added shx and cross-env to handle the cross-platform issues, while allowing us to retain the same functionality (mainly just prefixing the commands).

I verified on a normal windows shell that the following works:

- [x] yarn build (in root directory, using lerna)
- [x] yarn build-or-skip (doesn't like `[ -n \"$SKIP_BUILD\" ]` but always executes `yarn build`)
- [x] yarn test
- [x] yarn test-chrome
- [x] yarn test-edge

I think `find-proto` in `iov-bns` doesn't work, but this is really a preprocessor that no one except bns protobuf developers need to worry about.

### DIFF
--- a/.gitattibutes
+++ b/.gitattibutes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.gitattibutes
+++ b/.gitattibutes
@@ -1,1 +1,0 @@
-* text=auto
--- a/README.md
+++ b/README.md
@@ -58,6 +58,27 @@ To build the documentation locally, run `yarn install && yarn build && yarn docs
 in this repository. This will generate a `./docs` directory in each package that you
 can browse locally to see API docs on the various packages.
 
+## Developing under Windows 10
+
+Most of the developers working on this project in windows are also using Windows Subsystem for Linux
+(WSL), which should have maximum compatibility, as CI is linux and osx. However, we do attempt to ensure
+that this code also compiles on the normal windows shell, if you have set up git, node, yarn etc. correctly.
+(Note this has only be verified under windows 10, no guarantees for older versions).
+
+### Line endings (CRLF vs LF)
+
+All the code in the repo should use LF not the windows-specific CRLF as a line ending.
+`tsc` is currently set up to output properly. However, you should also make sure your editor
+saves with `LF` line endings rather than `CRLF`.
+
+A bigger issue is `git` changing the endings upon commit. Here is a short workaround,
+adapted from a [stackoverflow discussion](https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows):
+
+```
+git config --global core.autocrlf false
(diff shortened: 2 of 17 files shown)


ethanfrey:
One issue is that the generated files in `types/*` all end with CRLF instead of CR, so be careful not to check them in. 

Then again, I think these need to be rethought anyway as part of #342 
Just a heads-up not to check in these types files generated from yarn build on normal windows shell (not wsl)

Isabello:
Its possible to configure a windows git installation to checking/out as UNIX line styling and ignore m$ funny business.

ethanfrey:
Great tip. Can you write / link to how to do it.
I will add that to the README somewhere (maybe a link to "windows notes")

Isabello:
Example from stack overflow: https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows
``` 
The proper way to get LF endings in Windows is to first set core.autocrlf to false:

git config --global core.autocrlf false
You need to do this if you are using msysgit, because it sets it to true in its system settings.

Now git won’t do any line ending normalization. If you want files you check in to be normalized, do this: Set text=auto in your .gitattributes for all files:

* text=auto
And set core.eol to lf:

git config --global core.eol lf
Now you can also switch single repos to crlf (in the working directory!) by running

git config core.eol crlf
After you have done the configuration, you might want git to normalize all the files in the repo. To do this, go to to the root of your repo and run these commands:

git rm --cached -rf .
git diff --cached --name-only -z | xargs -n 50 -0 git add -f
If you now want git to also normalize the files in your working directory, run these commands:

git ls-files -z | xargs -0 rm
git checkout .
```

ethanfrey:
Updated some config and documented the two commands `git config --global ...` you need to run so the line endings work properly on windows.

I can actually do basic development workflow in the default windows command prompt. Crazy!

ethanfrey:
C'mon @Isabello does this get your approval now?

.gitattibutes
```diff
@@ -0,0 +1 @@
+* text=auto
```
webmaster128: `.gitattibutes` -> `.gitattributes`. However, if everything worked with `.gitattibutes`, this file is probably obsolete

ethanfrey: Afk. Can you push a commit deleting and then merge? 

Too many open prs and I think we need to get good ones in before more conflicts

PR #359, opened 2018-09-29
Move start and stop blockchain out of .travis.yml into reusable scripts

ethanfrey:
This allows for much easier setup/teardown of any tendermint/bcpd/bnsd related tests on the local system.

As I am working on query code against bcpd, this saves me a lot of trouble to verify locally it is passing without waiting for full ci run.

Issue #347, opened 2018-09-21
Add queries for escrow / atomic swap to the IovReader


ethanfrey:
Isabello can build atomic swap tx in the gui, but cannot see the state to properly get ids or verify on the second party. Let's extend this information like we do with queryAccount (and watchAccount)

Issue #291, opened 2018-08-24
Improve IovReader/IovWriter achitecture


webmaster128:
## Dependencies

The currently documented way to create an IovReader is complicated:

```js
const knownChains = await withConnectors([await bnsConnector("ws://localhost:22345")]);
// reader is already there: knownChains[0].client, no writer required

const writer = new IovWriter(profile, knownChains);
const chainId = writer.chainIds()[0];
const reader = writer.reader(chainId);
```

So that the moment we need a reader to make a writer but need a writer to make a reader.

## IovReader is responsible for writing to the blockchain

The term "reader" is irritating as IovReader is not only reading from the blockchain but also write signed transactions to it. It serves more as a networking component.

This becomes important when thinking about `disconnect()`. Right now, in order to disconnect a writer, you need to disconnect all the readers and as soon as you disconnect a reader, the writer does not work anymore. This behavior would make more sense, if it was clear, that IovReader is IovWriter's networking component.

ethanfrey:
The uglyness commented on above was fixed with #323 
The naming clarity part is still outstanding.

I would suggest `IovReader` -> `ChainConnection` or `BcpConnection` or something similar

`IovWriter` is a multi-chain signer. If we look at the separation of components in #301, we consider when this is out of process (eg. in extension, mobile phone), and the client using the `IovWriter` would maintain it's own set of `Connections` for queries.

I would also suggest naming the concrete class `IovWriter` -> `MultiChainSigner` or such, and defining an interface (with `chainId()`, but without `getReader()` functionality) that can refer to the in-process or out of process key management and transaction signing portion.

Maybe `Wallet`?

In short, my suggestion:

* `IovReader` interface -> `BcpConnection` (or similar)
* `IovWriter` concrete class -> `MultiChainSigner` (with a disconnect that handles all connections)
* `MultiChainSigner` implementes new`Wallet` interface, which we use in all the signing code. This `Wallet` does the signing can asks for user-approval, designed for the case where the `TransactionRequest` and the private key management are performed by different processes.

webmaster128:
From my point of view, this issue is mostly done. The current usage is much simpler and consistent

```ts
const writer = new IovWriter(profile);
await writer.addChain(bnsConnector("ws://localhost:22345"));
const chainId = writer.chainIds()[0];
const connection = writer.reader(chainId); // a BCP connection
```

[...]

PR #361, opened 2018-09-29
Define types, reader for atomic swap 

ethanfrey:
This defines the interfaces for #347 
Also addresses the naming issue for IovReader from #291 

Builds on top of #359 which should be merged first

I changed `IovReader` interface to `BcpConnection` and added a new `BcpAtomicSwapConnection` which extends the basic connection with information on atomic swap for those chains that support it.

I defined the types for atomic swap in [@iov/bcp-types](https://github.com/iov-one/iov-core/blob/separate-reader-types/packages/iov-bcp-types/src/atomicswap.ts)

I pulled some common functionality related to bcp into the bcp-types packages (as it is now ts and not just .d.ts), such as dummyEnvelope and the type guards for queries.

If the interfaces are clean, the next step for #347 is to actually implement the logic in @iov/bns.

### DIFF
--- a/packages/iov-bcp-types/src/atomicswap.ts
+++ b/packages/iov-bcp-types/src/atomicswap.ts
@@ -50,7 +50,7 @@ export interface BcpSwapSenderQuery {
 }
 
 export interface BcpSwapIdQuery {
-  readonly id: SwapIdBytes;
+  readonly swapid: SwapIdBytes;
 }
 
 // can we implement this as well? maybe id should equal swap
@@ -59,6 +59,19 @@ export interface BcpSwapIdQuery {
 // }
 
 export type BcpSwapQuery = BcpSwapRecipientQuery | BcpSwapSenderQuery | BcpSwapIdQuery;
+
+// isQueryBySwapRecipient is a type guard to use in the swap-based queries
+export function isQueryBySwapRecipient(query: BcpSwapQuery): query is BcpSwapRecipientQuery {
+  return (query as BcpSwapRecipientQuery).recipient !== undefined;
+}
+// isQueryBySwapSender is a type guard to use in the swap-based queries
+export function isQueryBySwapSender(query: BcpSwapQuery): query is BcpSwapSenderQuery {
+  return (query as BcpSwapSenderQuery).sender !== undefined;
+}
+// isQueryBySwapId is a type guard to use in the swap-based queries
+export function isQueryBySwapId(query: BcpSwapQuery): query is BcpSwapIdQuery {
+  return (query as BcpSwapIdQuery).swapid !== undefined;
+}
 
 export interface BcpSwapEvent {
   readonly height: number;
@@ -79,4 +92,13 @@ export interface BcpAtomicSwapConnection extends BcpConnection {
   // changeSwap triggers an event when a matching swap event occurs.
   // it emits the new height as well as the swap that was modified
   readonly changeSwap: (swap: BcpSwapQuery) => Stream<BcpSwapEvent>;
+}
+
+export function isAtomicSwapConnection(conn: BcpConnection): conn is BcpAtomicSwapConnection {
+  const check = conn as BcpAtomicSwapConnection;
+  return (
+    typeof check.getSwap === "function" &&
+    typeof check.watchSwap === "function" &&
+    typeof check.changeSwap === "function"
+  );
 }
--- a/packages/iov-bcp-types/src/atomicswap.ts
+++ b/packages/iov-bcp-types/src/atomicswap.ts
@@ -1,0 +1,82 @@
+import { Stream } from "xstream";
+
+import { BcpCoin, BcpConnection, BcpQueryEnvelope } from "./bcp";
+import { Address, TransactionIdBytes } from "./signables";
+import { SwapIdBytes } from "./transactions";
+
+export enum SwapState {
+  OPEN = "open",
+  CLAIMED = "claimed",
+  EXPIRED = "expired",
+}
+
+export interface SwapData {
+  readonly id: SwapIdBytes; // this is used as an unique identitier to locate the swap
+  readonly sender: Address;
+  readonly recipient: Address;
+  readonly hashlock: Uint8Array; // this is the hash, whose preimage releases the swap
+  readonly amount: ReadonlyArray<BcpCoin>;
+  readonly timeout: number; // blockheight where the swap expires (TODO: alternatively support Date?)
+  readonly memo?: string;
+}
+
+// OpenSwap is an offer that has not yet been claimed
+export interface OpenSwap {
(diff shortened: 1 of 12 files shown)


packages/iov-bcp-types/src/bcp.ts
```diff
@@ -70,15 +88,25 @@ export interface BcpValueNameQuery {
...
-// the connection, subscription, and such.
-export interface IovReader {
+// BcpConnection is the minimal interface needed to be supported by any blockchain
+// that is compatible with the bcp spec and iov-core library. This supports
+// getting account balances, sending tokens, and observing the blockchain state.
+//
+// There are other optional interfaces that extend this functionality with
+// features like atomic swap, NFTs, etc which may be implemented by any connector
+// to enable enhanced features in the clients
+export interface BcpConnection {
```
Isabello: This should not break anything in client side if its still referred to as a `writer.reader(chainId)`

ethanfrey: So far I just changed interfaces, not method names and not class names.

I would like to properly update writer, but that is another pr as it is far more breaking that this feature

Isabello: Ok, I will wait for the writer update so I can be prepared to update my end to match when the time comes.

ethanfrey: I would do that after swap and lisk are merged, so we can tag something useful for iov-wallet before breaking stuff.

I would also like to update redux to typescript, as it is much easier to handle renamings with typescript, as you get compile errors or warnings in the ide if the methods don't match. Otherwise it is a bunch of hunting around

packages/iov-bcp-types/src/atomicswap.ts
```diff
@@ -0,0 +1,90 @@
...
+// can we implement this as well? maybe id should equal swap
+// export interface SwapHashQuery {
+//   readonly hashlock: Uint8Array;
+// }
+
+export type BcpSwapQuery = BcpSwapRecipientQuery | BcpSwapSenderQuery | BcpSwapIdQuery;
+
+// isQueryBySwapRecipient is a type guard to use in the swap-based queries
+export function isQueryBySwapRecipient(query: BcpSwapQuery): query is BcpSwapRecipientQuery {
+  return (query as BcpSwapRecipientQuery).recipient !== undefined;
```
webmaster128: `query` should be cased to `any` to avoid mixing up compile time and runtime properties. Testing an BcpSwapRecipientQuery instance for `.recipient !== undefined` makes no sense: it is always defined.

Is I was a compiler, I would optimize out that check.

packages/iov-bcp-types/src/atomicswap.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+// isQueryBySwapSender is a type guard to use in the swap-based queries
+export function isQueryBySwapSender(query: BcpSwapQuery): query is BcpSwapSenderQuery {
+  return (query as BcpSwapSenderQuery).sender !== undefined;
```
webmaster128: same here

packages/iov-bcp-types/src/atomicswap.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+// isQueryBySwapId is a type guard to use in the swap-based queries
+export function isQueryBySwapId(query: BcpSwapQuery): query is BcpSwapIdQuery {
+  return (query as BcpSwapIdQuery).swapid !== undefined;
```
webmaster128: same here

packages/iov-bcp-types/src/atomicswap.ts
```diff
@@ -0,0 +1,90 @@
...
+  // getSwap returns all matching swaps in their current state
+  readonly getSwap: (swap: BcpSwapQuery) => Promise<BcpQueryEnvelope<BcpAtomicSwap>>;
+
+  // watchSwap emits currentState (getSwap) as a stream, then sends updates for any matching swap
+  // this includes an open swap beind claimed/expired as well as a new matching swap being offered
+  readonly watchSwap: (swap: BcpSwapQuery) => Stream<BcpAtomicSwap>;
+}
+
+export function isAtomicSwapConnection(conn: BcpConnection): conn is BcpAtomicSwapConnection {
+  const check = conn as BcpAtomicSwapConnection;
```
webmaster128: same here

packages/iov-bcp-types/src/bcp.ts
```diff
@@ -70,15 +88,25 @@ export interface BcpValueNameQuery {
 
 export type BcpAccountQuery = BcpAddressQuery | BcpValueNameQuery;
 
-// IovReader is a high-level interface to a blockchain node,
+// isQueryByAddress is a type guard to use in the account-based queries
+export function isQueryByAddress(query: BcpAccountQuery): query is BcpAddressQuery {
+  return (query as BcpAddressQuery).address !== undefined;
```
webmaster128: same here

packages/iov-bcp-types/src/bcp.ts
```diff
@@ -99,6 +127,10 @@ export interface IovReader {
   readonly chainId: () => Promise<ChainId>;
   readonly height: () => Promise<number>;
 
+  // these emits the new blockHeight on every block,
+  // so you can trigger a custom response
+  readonly changeBlock: () => Stream<number>;
```
webmaster128: Not for this PR but as I am just seeing this: wouldn't `blockChanges()` be a better name? Method names starting with a verb usually indicate an action. When I implemented Lisk, I didn't have any idea what the three mothods `change*` were supposed to do

packages/iov-bcp-types/src/atomicswap.ts
```diff
@@ -0,0 +1,90 @@
+import { Stream } from "xstream";
+
+import { BcpCoin, BcpConnection, BcpQueryEnvelope } from "./bcp";
+import { Address } from "./signables";
+import { SwapIdBytes } from "./transactions";
+
+export enum SwapState {
+  OPEN = "open",
```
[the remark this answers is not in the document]
ethanfrey: Okay, then we should fix up Algorithm as well, maybe a few more in basic types that I used as templates

packages/iov-bcp-types/src/bcp.ts
ethanfrey: Nice point. We should make a new pr

ethanfrey:
I will fix comments and rebase on lisk, where I think there will be a conflict.

packages/iov-bcp-types/src/atomicswap.ts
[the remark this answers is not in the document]
webmaster128: Indeed. https://github.com/iov-one/iov-core/issues/368

PR #362, opened 2018-09-29
Bns atomic swap

### LINKED ISSUE #347: Add queries for escrow / atomic swap to the IovReader

ethanfrey:
Resolves #347 
Depends on #361 

This adds support for atomic swap queries to the bns client.

- [x] query open atomic swap offers from abci state
- [x] query closed atomic swaps from transaction history
- [x] watch atomic swaps
- [x] cleanup api


packages/iov-bcp-types/src/atomicswap.ts
```diff
@@ -0,0 +1,94 @@
...
+  EXPIRED = "expired",
+}
+
+export interface SwapData {
+  readonly id: SwapIdBytes; // this is used as an unique identitier to locate the swap
+  readonly sender: Address;
+  readonly recipient: Address;
+  readonly hashlock: Uint8Array; // this is the hash, whose preimage releases the swap
+  readonly amount: ReadonlyArray<BcpCoin>;
+  readonly timeout: number; // blockheight where the swap expires (TODO: alternatively support Date?)
```
Isabello: I am not sure how doing it by date would work. We can simulate it on the frontend already for user inputs, as we have a block time estimator.

ethanfrey: The date would have to be implemented on the blockchain. Since it enforces the expiration.

Right now, we just use blockheight in weave. A note to reconsider if we encounter chains (or update weave), so they use wall time to control expiration.

ethanfrey:
I think the current implementation is technically done.

I would like a review, both on code quality and on the updated interfaces. I'm sure there are a few things for me to clean up in here.

PR #364, opened 2018-09-30
Ensure consecutive Safari sessions don't re-open old tabs

webmaster128:


webmaster128:
Oh this looks better @ethanfrey

PR #367, opened 2018-10-02
BCP/BNS cleanups

webmaster128:
A handful of minor changes that I came across while working on Lisk integration

### DIFF
--- a/packages/iov-bns/src/index.ts
+++ b/packages/iov-bns/src/index.ts
@@ -1,2 +1,3 @@
 export { bnsCodec } from "./bnscodec";
 export { Client } from "./client";
+export { bnsFromOrToTag, bnsNonceTag } from "./tags";
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -52,7 +52,14 @@ export interface BcpNonce {
 export interface BcpTicker {
   readonly tokenTicker: TokenTicker;
   readonly sigFigs: number;
-  readonly tokenName?: string;
+  /**
+   * A name to be displayed to the user which allows differentiation
+   * of multiple tokens that use the same ticker.
+   *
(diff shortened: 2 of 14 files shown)


ethanfrey:
I like the idea, and yeah, chain connector should be bcp types and the Bns implementation in Bns package, like lisk implementation in lisk package.

Not sure if weave makes token names mandatory... Maybe need to fix on the backend before requiring them. Or just default to ticker if no response from server?

Good call moving out the Bns tags stuff, but I think those are also duplicated somewhere... Yet another cleanup

webmaster128:
> Not sure if weave makes token names mandatory... Maybe need to fix on the backend before requiring them. Or just default to ticker if no response from server?

In all the cases I saw, the field exists. In Lisk I set it to the one and only constant value and in BNS we have

```ts
export class Normalize {
  public static token(data: codecImpl.namecoin.IToken & Keyed): BcpTicker {
    return {
      tokenTicker: Encoding.fromAscii(data._id) as TokenTicker,
      tokenName: ensure(data.name),
      sigFigs: ensure(data.sigFigs),
    };
  }
```

So I think every new client can ensure to set a sensible value

packages/iov-bns/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export { bnsCodec } from "./bnscodec";
 export { Client } from "./client";
+export { bnsFromOrToTag, bnsNonceTag } from "./tags";
```
ethanfrey: This is a good call.
It didn't belong in writer.ts at all, just wanted it to be exposed to the outside world through iov-core.

We can rethink how tags are built later, but this is a nice cleanup. step by step

ndooreck:
Can you please unsubscribe me from this list? Tried to do it several times on Girhub but doesn’t seem to work!

Isabello:
@ndooreck you have to unsubscribe yourself by changing your watch settings.
image

PR #372, opened 2018-10-02
Let Ed25519SimpleAddressKeyringEntry.createIdentity take an index 

### LINKED ISSUE #371: Let Ed25519SimpleAddressKeyringEntry.createIdentity take an index argument
A simple address takes an index `i` as defined in https://github.com/iov-one/iov-core/blob/master/docs/KeyBase.md#simple-addresses

This is in intermediate step for #302

webmaster128:
Resolves #371


packages/iov-bns/src/client.spec.ts
```diff
@@ -52,12 +52,12 @@ describe("Integration tests with bov+tendermint", () => {
   // max open connections??? (but 900 by default)
   const tendermintUrl = "ws://localhost:22345";
 
-  const userProfile = async (): Promise<UserProfile> => {
+  async function userProfileWithFaucet(): Promise<UserProfile> {
```
ethanfrey: you really don't like my `async () =>` abuse do you?

ethanfrey: better name in any case

packages/iov-core/src/writer.spec.ts
```diff
@@ -46,15 +46,6 @@ describe("IovWriter", () => {
       return profile;
     };
 
-    // will make identities if needed.
```
ethanfrey: yeah, you just got rid of this nasty logic. 🎉

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -170,7 +170,7 @@ export class Slip10KeyringEntry implements KeyringEntry {
     this.labelProducer.update(label);
   }
 
-  public async createIdentity(): Promise<LocalIdentity> {
+  public async createIdentity(_: number): Promise<LocalIdentity> {
```
ethanfrey: hmm... didn't you make this take optional keypair in other pr?

packages/iov-bns/src/client.spec.ts
webmaster128: I am trying hard to get used to it but it is still taking me considerably longer to parse that code when reading

ethanfrey: The only time I think it really shines are either inline functions or partial ones...

```
const parse = (config: Config) => (item: Item) => {
  // do somethng with item based on config values
}
const parsed = items.map(parse(myConfig));
```

type def on top line reads almost like Haskell ;)

PR #376, opened 2018-10-04
Move ChainConnector into bcp-types

### LINKED ISSUE #366: Move interface ChainConnector into bcp-types
Right now, @iov/lisk depends on @iov/core just because of ChainConnector, which is an unnecessary dependency.

ChainConnector is basically a BCP client and a BCP codec, so it can easily move to bcp-types.

Depends on #361 to avoid merge conflicts

webmaster128:
Closes #366

### DIFF
--- a/packages/iov-core/src/index.ts
+++ b/packages/iov-core/src/index.ts
@@ -16,4 +16,4 @@ export {
 } from "@iov/keycontrol";
 export { ChainId } from "@iov/tendermint-types";
 
-export { bnsConnector, ChainConnector, IovWriter } from "./writer";
+export { bnsConnector, IovWriter } from "./writer";
--- a/packages/iov-core/src/index.ts
+++ b/packages/iov-core/src/index.ts
@@ -4,7 +4,7 @@
 // this should serve as an entry point into the whole monorepo.
 
 export { Address, Nonce, SendTx, SetNameTx, TokenTicker, TransactionKind } from "@iov/bcp-types";
-export { bnsFromOrToTag, bnsNonceTag, bnsSwapQueryTags } from "@iov/bns";
+export { bnsConnector, bnsFromOrToTag, bnsNonceTag, bnsSwapQueryTags } from "@iov/bns";
 export {
   Ed25519KeyringEntry,
   Ed25519SimpleAddressKeyringEntry,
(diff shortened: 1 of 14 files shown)


packages/iov-core/src/index.ts
```diff
@@ -16,4 +16,4 @@ export {
 } from "@iov/keycontrol";
 export { ChainId } from "@iov/tendermint-types";
 
-export { bnsConnector, ChainConnector, IovWriter } from "./writer";
+export { IovWriter } from "./writer";
```
ethanfrey: We no longer export ChainConnector anywhere?
I guess we only want concrete implementations, but if someone imports iov-core from npm to add support for another blockchain, they may want that type.

packages/iov-core/src/writer.ts
```diff
@@ -34,12 +52,12 @@ export class IovWriter {
...
 
   public reader(chainId: ChainId): BcpConnection {
-    return this.getChain(chainId).client;
+    return this.getChain(chainId).connection;
   }
 
   public async addChain(connector: ChainConnector): Promise<void> {
     const connection = await connectChain(connector);
-    const chainId = connection.client.chainId();
+    const chainId = connection.connection.chainId();
```
ethanfrey: Hmmm... a bit of stutter. I like the renaming, but maybe this could be clearer.
Just a thought, no need to change this for the pr

packages/iov-core/src/writer.ts
```diff
@@ -66,7 +84,7 @@ export class IovWriter {
     keyring: number | KeyringEntryId,
   ): Promise<BcpTransactionResponse> {
     const chainId = tx.chainId;
-    const { client, codec } = this.getChain(chainId);
+    const { connection, codec } = this.getChain(chainId);
```
ethanfrey: yes, here this is a nice change of name in the struct

packages/iov-core/src/index.ts
webmaster128: > if someone imports iov-core from npm to add support for another blockchain, they may want that type.

yes, they get the type from @iov/bcp-types, which now includes everything you need to create a custom connector. As you can see in https://github.com/iov-one/iov-core/pull/376/commits/ed1338c1e9349cf2a82b41a97235a931c5693e32, @iov/lisk gets rid of the @iov/core dependency by this change.

ethanfrey: cool. :+1:

webmaster128:
Note to myself: this change was backwards compatible for JavaScript users, because the term ChainConnector does not exist in .js output. However, it was breaking for TypeScript users since `import { ChainConnector } from "@iov/core";` is not available anymore.

PR #379, opened 2018-10-04
Create Secp256k1HdWallet

### LINKED ISSUE #294: Add secp256k1 keyring entry

### LINKED ISSUE #384: Rename Slip10KeyringEntry -> Slip10Wallet
as discussed in #302. Depends on #379 and #380 to avoid unnecessary merge conflicts.

This is non-breaking since Slip10KeyringEntry is not part of the public API.

webmaster128:
Closes #294
Closes #384


packages/iov-keycontrol/src/keyring-entries/secp256k1hdwallet.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+describe("Secp256k1HdWallet", () => {
+  it("returns the concrete type when creating from entropy", () => {
+    const wallet = Secp256k1HdWallet.fromEntropy(fromHex("51385c41df88cbe7c579e99de04259b1aa264d8e2416f1885228a4d069629fad"));
+    expect(wallet).toEqual(jasmine.any(Secp256k1HdWallet));
+    expect(wallet.implementationId).toEqual("secp256k1-hd");
+  });
+
+  it("returns the concrete type when creating from mnemonic", () => {
+    const wallet = Secp256k1HdWallet.fromMnemonic("execute wheel pupil bachelor crystal short domain faculty shrimp focus swap hazard");
+    expect(wallet).toEqual(jasmine.any(Secp256k1HdWallet));
```
ethanfrey: nice check... this is like `instanceOf` in curried form, right?

packages/iov-keycontrol/src/keyring-entries/secp256k1hdwallet.ts
```diff
@@ -0,0 +1,35 @@
+import { Slip10Curve, Slip10RawIndex } from "@iov/crypto";
+
+import { KeyringEntryImplementationIdString, LocalIdentity } from "../keyring";
+import { Slip10KeyringEntry } from "./slip10";
+
+export class Secp256k1HdWallet extends Slip10KeyringEntry {
```
ethanfrey: Amazing how small this class gets....

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -274,12 +285,11 @@ export class Slip10KeyringEntry implements KeyringEntry {
   }
 
   // This throws an exception when private key is missing
-  private async privkeyForIdentity(identity: PublicIdentity): Promise<Ed25519Keypair> {
+  private async privkeyForIdentity(identity: PublicIdentity): Promise<Uint8Array> {
```
ethanfrey: This is all private stuff, so it is clear enough between where it is created and where it is used.
On one side, we need to `makeKeypair`, the other just pass it. But you can just scroll down a few lines to understand.

packages/iov-keycontrol/src/keyring-entries/slip10.ts
```diff
@@ -227,8 +226,20 @@ export class Slip10KeyringEntry implements KeyringEntry {
...
     _: ChainId,
   ): Promise<SignatureBytes> {
-    const keypair = await this.privkeyForIdentity(identity);
-    const signature = await Ed25519.createSignature(prehash(transactionBytes, prehashType), keypair);
+    const privkey = await this.privkeyForIdentity(identity);
+    const message = prehash(transactionBytes, prehashType);
+
+    let signature: Uint8Array;
+    switch (this.curve) {
+      case Slip10Curve.Ed25519:
```
ethanfrey: I think whether we want to pull out this switch into protected method from subclass, but honestly overkill if we only add a new curve once every blue moon....

webmaster128: Only one more curve [is defined in SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation): NIST P-256 curve. If we want to have it, no problem. But I don't think there is more coming in the foreseeable future

packages/iov-keycontrol/src/keyring-entries/secp256k1hdwallet.spec.ts
webmaster128: jap

PR #381, opened 2018-10-04
Don't use SignableBytes as output type of prehash()

webmaster128:
this is confusing since SignableBytes should either be prehashed or plain

### DIFF
--- a/packages/iov-keycontrol/src/prehashing.ts
+++ b/packages/iov-keycontrol/src/prehashing.ts
@@ -1,14 +1,14 @@
 import { PrehashType, SignableBytes } from "@iov/bcp-types";
 import { Sha256, Sha512 } from "@iov/crypto";
 
-export function prehash(bytes: SignableBytes, type: PrehashType): SignableBytes {
+export function prehash(bytes: SignableBytes, type: PrehashType): Uint8Array {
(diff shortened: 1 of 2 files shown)


packages/iov-keycontrol/src/prehashing.ts
```diff
@@ -1,14 +1,14 @@
 import { PrehashType, SignableBytes } from "@iov/bcp-types";
 import { Sha256, Sha512 } from "@iov/crypto";
 
-export function prehash(bytes: SignableBytes, type: PrehashType): SignableBytes {
+export function prehash(bytes: SignableBytes, type: PrehashType): Uint8Array {
```
ethanfrey: Maybe add more info rather than losing it?
Like `PrehashedSignableBytes`
Then we could enforce that prehash was run before signing later on.

webmaster128:
>  Why not add a second type... they have no runtime overhead.

I don't think it is worth adding a custom type for a scope of ~ 10 lines of code.

Tagged types are problematic when used with JavaScript, because they don't exists and it is hard to understand that e.g. an Address needs to be created with `new Uint8Array()`. This is not an argument here but I am developing the tendency to not use that many new tagged types.

PR #394, opened 2018-10-05
Convert some comments to TSDoc format

webmaster128:
This is mostly to show the idea:

- [tsdocs](https://github.com/Microsoft/tsdoc) are preserved in declaration files
- tsdocs [are preserved in our documentation](https://iov-one.github.io/iov-core-docs/latest/iov-crypto/enums/slip10curve.html)
- IDEs use that to show function/class descriptions at the caller side

oneliner

Additionally, I am removing the duplicate class/function name from the comment to

1. make comments shorter
2. avoid mismatch between comment and descrition after simple renamings

### DIFF
--- a/packages/iov-bns/src/client.ts
+++ b/packages/iov-bns/src/client.ts
@@ -45,8 +45,10 @@ import { bnsFromOrToTag, bnsNonceTag, bnsSwapQueryTags } from "./tags";
 import { Decoder, Keyed, Result } from "./types";
 import { arraysEqual, hashIdentifier, isSwapOffer, isSwapRelease } from "./util";
 
-// onChange returns a filter than only passes when the
-// value is different than the last one
+/**
+ * Returns a filter that only passes when the
+ * value is different than the last one
+ */
 function onChange<T>(): (val: T) => boolean {
   let oldVal: T | undefined;
   return (val: T): boolean => {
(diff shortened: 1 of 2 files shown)


packages/iov-bns/src/client.ts
```diff
@@ -58,9 +60,12 @@ function onChange<T>(): (val: T) => boolean {
...
 
-// Client talks directly to the BNS blockchain and exposes the
-// same interface we have with the BCP protocol.
-// We can embed in iov-core process or use this in a BCP-relay
+/**
+ * Talks directly to the BNS blockchain and exposes the
+ * same interface we have with the BCP protocol.
+ *
+ * We can embed in iov-core process or use this in a BCP-relay
+ */
```
ethanfrey: Did we want to change this name to Connection as well, or other pr I guess?

packages/iov-bns/src/client.ts
```diff
@@ -181,7 +190,9 @@ export class Client implements BcpAtomicSwapConnection {
     return dummyEnvelope(data);
   }
 
-  // getSwapFromState returns all matching swaps that are open (from app state)
+  /**
+   * All matching swaps that are open (from app state)
```
ethanfrey: Yes, this format does look cleaner.
Always prefix with function name is golang standard, not needed here

packages/iov-bns/src/client.ts
webmaster128: Yes, we want. Incompatible change so I create a ticket

PR #412, opened 2018-10-05
Improve shell scripts to start/stop blockchain

webmaster128:
further improvements found by recent shellchack have been applied

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ before_script:
 # Use Docker if available (currently Linux only)
 # macOS/Linux compatible temp dir deleted on reboot (https://stackoverflow.com/a/31397073)
 - |
-  if which docker > /dev/null ; then
+  if command -v docker > /dev/null ; then
     source ./scripts/blockchain_start.sh
   fi
 
--- a/scripts/blockchain_start.sh
+++ b/scripts/blockchain_start.sh
@@ -1,35 +1,41 @@
 # blockchain_start is called in .travis.yml or also locally before running tests
 # to ensure we have all the blockchains set up for the full integration tests
 
-# get this files directory regardless of pwd when we run it
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-# this is root of all the lerna packages to find package-specific scripts
-PACKAGES="${DIR}/../packages"
+if [ -z "${BASH_SOURCE[0]}" ]; then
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -47,7 +47,7 @@ before_script:
 # Use Docker if available (currently Linux only)
 # macOS/Linux compatible temp dir deleted on reboot (https://stackoverflow.com/a/31397073)
 - |
-  if which docker > /dev/null ; then
+  if command -v docker > /dev/null ; then
```
ethanfrey: interesting, i did not know about `command`

scripts/blockchain_start.sh
```diff
@@ -1,35 +1,45 @@
...
 # to ensure we have all the blockchains set up for the full integration tests
 
-# get this files directory regardless of pwd when we run it
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-# this is root of all the lerna packages to find package-specific scripts
-PACKAGES="${DIR}/../packages"
+if [ -z "${BASH_SOURCE[0]}" ]; then
+  echo "\${BASH_SOURCE[0]} is unset or empty. This usually means your current shell does not support it.";
+  echo "Please use bash do start and stop the blockchain or use your shell scripting superhero powers to show us a better solution at"
+  echo "https://github.com/iov-one/iov-core"
```
ethanfrey: how about `exit 1` here, and then the rest of the else statement top-level?

I would like to be able to abort a script if the blockchain didn't start, here we only write text but there is no programmatic feedback. (error code)

webmaster128: this is not possible, since exit will close the user's console

.travis.yml
webmaster128: same here. Thanks shellcheck!

scripts/blockchain_start.sh
Isabello: It shouldn't. It should only exit the script with an error code., not the console itself.

webmaster128: the problem is that "the script" is your console if you `source` source code

ethanfrey: kudos to @webmaster128 
that is very true....

maybe the last line before the `else` can be `false`

```
$ echo true > true.sh
$ echo false > false.sh
$ source true.sh
$ echo $?
0
$ source false.sh
$ echo $?
1
```

PR #421, opened 2018-10-09
Convert type Address to string

### LINKED ISSUE #415: Make address representation string
The encoding is chain-dependent
All external APIs use string
The BnsConnector/TxCodec must know how to parse this string and create the correct format to post a transaction or query the blockchain.

Expose BnsConnection as hex
Expose LiskConnection as lisk format

This is a pre-requisite for #339 which will update the external string encoding for bns

webmaster128:
Closes #415

### DIFF
--- a/packages/iov-bns/src/encode.ts
+++ b/packages/iov-bns/src/encode.ts
@@ -13,7 +13,7 @@ import {
 
 import * as codecImpl from "./codecimpl";
 import { encodeFullSig, encodeToken } from "./types";
-import { keyToAddress, preimageIdentifier } from "./util";
+import { decodeBnsAddress, keyToAddress, preimageIdentifier } from "./util";
 
 export const buildSignedTx = (tx: SignedTransaction): codecImpl.app.ITx => {
   const sigs: ReadonlyArray<FullSignature> = [tx.primarySignature, ...tx.otherSignatures];
@@ -48,8 +48,8 @@ export const buildMsg = (tx: UnsignedTransaction): codecImpl.app.ITx => {
 
 const buildSendTx = (tx: SendTx): codecImpl.app.ITx => ({
   sendMsg: codecImpl.cash.SendMsg.create({
-    src: keyToAddress(tx.signer),
-    dest: tx.recipient,
+    src: decodeBnsAddress(keyToAddress(tx.signer)),
+    dest: decodeBnsAddress(tx.recipient),
     amount: encodeToken(tx.amount),
     memo: tx.memo,
   }),
@@ -57,7 +57,7 @@ const buildSendTx = (tx: SendTx): codecImpl.app.ITx => ({
 
 const buildSetNameTx = (tx: SetNameTx): codecImpl.app.ITx => ({
   setNameMsg: codecImpl.namecoin.SetWalletNameMsg.create({
-    address: keyToAddress(tx.signer),
+    address: decodeBnsAddress(keyToAddress(tx.signer)),
     name: tx.name,
   }),
 });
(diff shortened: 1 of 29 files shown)


packages/iov-bns/src/client.spec.ts
```diff
@@ -45,7 +45,7 @@ describe("Integration tests with bov+tendermint", () => {
   // the first key generated from this mneumonic produces the given address
   // this account has money in the genesis file (setup in docker)
   const mnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
-  const expectedFaucetAddress = Encoding.fromHex("b1ca7e78f74423ae01da3b51e676934d9105f282") as Address;
+  const expectedFaucetAddress = "B1CA7E78F74423AE01DA3B51E676934D9105F282" as Address;
```
ethanfrey: :+1:

packages/iov-bns/src/encode.ts
```diff
@@ -48,16 +48,16 @@ export const buildMsg = (tx: UnsignedTransaction): codecImpl.app.ITx => {
 
 const buildSendTx = (tx: SendTx): codecImpl.app.ITx => ({
   sendMsg: codecImpl.cash.SendMsg.create({
-    src: keyToAddress(tx.signer),
-    dest: tx.recipient,
+    src: decodeBnsAddress(keyToAddress(tx.signer)),
```
ethanfrey: Love all these encodeBnsAddress/decodeBnsAddress... thank you types for highlighting what to change

packages/iov-bns/src/normalize.ts
```diff
@@ -53,7 +52,7 @@ export class Normalize {
     return (acct: codecImpl.namecoin.IWallet & Keyed): BcpAccount => {
       return {
         name: typeof acct.name === "string" ? acct.name : undefined,
-        address: acct._id as Address,
+        address: encodeBnsAddress(acct._id),
```
ethanfrey: :+1: this was actually one of my first thoughts to check

packages/iov-bns/src/util.ts
```diff
@@ -14,8 +14,20 @@ import { Sha256 } from "@iov/crypto";
...
+export function encodeBnsAddress(bytes: Uint8Array): Address {
+  return Encoding.toHex(bytes).toUpperCase() as Address;
+}
+
+/** Decodes a printable address into raw bytes */
+export function decodeBnsAddress(address: Address): Uint8Array {
+  return Encoding.fromHex(address);
+}
+
+export function keyToAddress(key: PublicKeyBundle): Address {
```
ethanfrey: nice. we never need to export the Uint8Array -> Address encoder explicitly to other packages, do we? That is done in the BcpConnection.

Cool, even easier than I thought

packages/iov-lisk/src/liskcodec.spec.ts
```diff
@@ -27,7 +27,7 @@ describe("liskCodec", () => {
       algo: Algorithm.Ed25519,
       data: fromHex("f4852b270f76dc8b49bfa88de5906e81d3b001d23852f0e74ba60cac7180a184") as PublicKeyBytes,
     };
-    expect(liskCodec.keyToAddress(pubkey)).toEqual(toAscii("6076671634347365051L"));
+    expect(liskCodec.keyToAddress(pubkey)).toEqual("6076671634347365051L");
```
ethanfrey: looks cleaner the new way

webmaster128:
> less invasive to the interfaces than I thought possible

same here.

> nice. we never need to export the Uint8Array -> Address encoder explicitly to other packages, do we? That is done in the BcpConnection.

At least not for the UI side. Right now the code makes the assumption that the UI facing string address is the same format than the facet call expects. When we change the format, we must hack the benc32->hex conversion into the faucet code as this is not compatible with the new style addresses. But for now, this is fine.

ethanfrey:
> At least not for the UI side. Right now the code makes the assumption that the UI facing string address is the same format as the faucet call expects. When we change the format, we must hack the bench32->hex conversion into the faucet code as the faucet is not compatible with the new style addresses. But for now, this is fine.


okay, port faucet before we do bech32.
i'll quiet down and wait.

PR #422, opened 2018-10-09
Remove Ed25519SimpleAddressKeyringEntry

### LINKED ISSUE #420: Remove Ed25519SimpleAddressKeringEntry
All of its functionality exists in the Ed25519HdWallet now. All we need is a helper

```ts
function iovSimpleAddressPath(index: number): ReadonlyArray<Slip10RawIndex> {
  const iovPurpose = 4804438;
  return [Slip10RawIndex.hardened(iovPurpose), Slip10RawIndex.hardened(index)];
}
```

webmaster128:
Closes #420


packages/iov-bns/src/client.spec.ts
```diff
@@ -54,8 +49,8 @@ describe("Integration tests with bov+tendermint", () => {
 
   async function userProfileWithFaucet(): Promise<UserProfile> {
     const profile = new UserProfile();
-    profile.addEntry(Ed25519SimpleAddressKeyringEntry.fromMnemonic(mnemonic));
-    await profile.createIdentity(0, 0);
+    profile.addEntry(Ed25519HdWallet.fromMnemonic(mnemonic));
```
ethanfrey: Yes, this is clean usage

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -0,0 +1,13 @@
+import { Slip10RawIndex } from "@iov/crypto";
+
+export class HdPaths {
+  /**
+   * IOV's SimpleAddress derivation path
+   *
+   * @see https://github.com/iov-one/iov-core/blob/v0.6.1/docs/KeyBase.md#simple-addresses
+   */
+  public static simpleAddress(index: number): ReadonlyArray<Slip10RawIndex> {
```
ethanfrey: Okay, nice. So this is the first of various path derivations.

Later bip44, etc... or use raw for full bip32

ethanfrey: It would be nice to see a test (later once implemented) with different path derivations in the same keyring.

one simpleAddress, one bip44, one bip44 with another chain, etc.

webmaster128:
> Very nice. I think you just missed an export in @iov/cli

Thanks! There is one issue with @iov/cli imports: since the CI tests a global installation of @iov/cli, I can only import stuff that is available in the last version released on npm. Not cool but the other option is to remove the global installation test again

PR #431, opened 2018-10-10
Test identity ID collision in Slip10Wallet

webmaster128:


webmaster128:
Oh, nice – looks like this catches a bug in MultiChainSigner/BNS compatibility tests. Will investigate

PR #437, opened 2018-10-11
Add and enforce optional expectedChainId in ChainConnector

ethanfrey:
If we know the chainId ahead of time, we can pass it to `addChain` for extra security


ethanfrey:
Let's see what the CI has to say. I think this is done.

packages/iov-core/src/multichainsigner.spec.ts
```diff
@@ -132,4 +133,37 @@ describe("MultiChainSigner", () => {
       expect(acct.data[0].balance.length).toBe(1);
     });
   });
+
+  // transforms promise so resolved->rejected and rejected->resolved
+  const expectRejected = (prom: Promise<any>): Promise<boolean> =>
```
webmaster128: Could you make this either return `Promise<void>` or `Promise<any>` (with the error from the catch handler returned)?

In the later case we could potentially check what kind of error we got

ethanfrey: okay, Promise<any> it is

PR #438, opened 2018-10-12
WIP need help expanding iov-ledger-bns cleanly

rudi-cilibrasi:
these are too confusing for me to expand cleanly but i thought i should push my (broken) work in case there is a line or two you guys can use when you redo this path enhancement part.

### DIFF
--- a/packages/iov-ledger-bns/src/app.ts
+++ b/packages/iov-ledger-bns/src/app.ts
@@ -13,16 +13,19 @@ const cmdPubkeyWithPath = 5;
 const cmdAppVersion = 0xca;
 
 export function getPublicKey(transport: TransportNodeHid): Promise<Uint8Array> {
-  return sendChunks(transport, appCode, cmdPubkey, new Uint8Array([]));
+  return sendChunks(transport, appCode, cmdPubkey, new Uint8Array([0]));
 }
 
 export function getPublicKeyWithIndex(transport: TransportNodeHid, i: number): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  return sendChunks(transport, appCode, cmdPubkeyWithPath, encodeUint32(pathComponent));
+  let pathPart = encodeUint32(pathComponent);
+  pathPart = [pathComponent.length, ...pathPart];
+  return sendChunks(transport, appCode, cmdPubkeyWithPath, pathPart);
 }
 
 export function signTransaction(transport: TransportNodeHid, transaction: Uint8Array): Promise<Uint8Array> {
-  return sendChunks(transport, appCode, cmdSign, transaction);
+  const paddedTx: ReadOnlyArray<Uint8> = [0, ...transaction];
+  return sendChunks(transport, appCode, cmdSign, paddedTx);
 }
 
 export function signTransactionWithIndex(
@@ -31,7 +34,9 @@ export function signTransactionWithIndex(
   i: number,
 ): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  const data = new Uint8Array([...encodeUint32(pathComponent), ...transaction]);
+  let pathPart = encodeUint32(pathComponent);
+  pathPart = [pathComponent.length, ...pathPart];
+  const data = new Uint8Array([...pathPart, ...transaction]);
   return sendChunks(transport, appCode, cmdSignWithPath, data);
 }
 
--- a/packages/iov-ledger-bns/src/exchange.ts
+++ b/packages/iov-ledger-bns/src/exchange.ts
@@ -112,6 +112,7 @@ export async function sendChunks(
 ): Promise<Uint8Array> {
   let offset = 0;
   // loop over the non-end chunks
+  // add length indicator for testing
   while (offset + 255 < payload.length) {
     const chunk = payload.slice(offset, offset + 255);
     offset += 255;


packages/iov-ledger-bns/src/app.ts
```diff
@@ -13,16 +13,19 @@ const cmdPubkeyWithPath = 5;
...
 export function getPublicKey(transport: TransportNodeHid): Promise<Uint8Array> {
-  return sendChunks(transport, appCode, cmdPubkey, new Uint8Array([]));
+  return sendChunks(transport, appCode, cmdPubkey, new Uint8Array([0]));
 }
 
 export function getPublicKeyWithIndex(transport: TransportNodeHid, i: number): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  return sendChunks(transport, appCode, cmdPubkeyWithPath, encodeUint32(pathComponent));
+  let pathPart = encodeUint32(pathComponent);
+  pathPart = [pathComponent.length, ...pathPart];
```
webmaster128: Does this compile? What you want is

`const pathPart = new Uint8Array([pathComponent.length, ...encodeUint32(pathComponent)]);`

packages/iov-ledger-bns/src/app.ts
```diff
@@ -13,16 +13,19 @@ const cmdPubkeyWithPath = 5;
 const cmdAppVersion = 0xca;
 
 export function getPublicKey(transport: TransportNodeHid): Promise<Uint8Array> {
-  return sendChunks(transport, appCode, cmdPubkey, new Uint8Array([]));
+  return sendChunks(transport, appCode, cmdPubkey, new Uint8Array([0]));
```
webmaster128: 👍

packages/iov-ledger-bns/src/app.ts
```diff
@@ -13,16 +13,19 @@ const cmdPubkeyWithPath = 5;
...
   const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  return sendChunks(transport, appCode, cmdPubkeyWithPath, encodeUint32(pathComponent));
+  let pathPart = encodeUint32(pathComponent);
+  pathPart = [pathComponent.length, ...pathPart];
+  return sendChunks(transport, appCode, cmdPubkeyWithPath, pathPart);
 }
 
 export function signTransaction(transport: TransportNodeHid, transaction: Uint8Array): Promise<Uint8Array> {
-  return sendChunks(transport, appCode, cmdSign, transaction);
+  const paddedTx: ReadOnlyArray<Uint8> = [0, ...transaction];
```
webmaster128: Does this compile?

Simpler:

`const paddedTx = new Uin8Array([0, ...transaction]);`

Independent of the change: the account for signing is encoded in the transaction? Do we really need to prepend the path length for `cmdSign`?

packages/iov-ledger-bns/src/app.ts
```diff
@@ -31,7 +34,9 @@ export function signTransactionWithIndex(
   i: number,
 ): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  const data = new Uint8Array([...encodeUint32(pathComponent), ...transaction]);
+  let pathPart = encodeUint32(pathComponent);
+  pathPart = [pathComponent.length, ...pathPart];
+  const data = new Uint8Array([...pathPart, ...transaction]);
```
webmaster128: Can be simplified a bit (see above) but looks correct

packages/iov-ledger-bns/src/app.ts
webmaster128: Ahahaa, here we go: `pathComponent` is a number, the index `i` from https://github.com/iov-one/iov-core/blob/v0.6.1/docs/KeyBase.md#simple-address-derivation, so `pathComponent.length` cannot compile. length is always 1.

packages/iov-ledger-bns/src/app.ts
```diff
@@ -31,7 +34,9 @@ export function signTransactionWithIndex(
   i: number,
 ): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  const data = new Uint8Array([...encodeUint32(pathComponent), ...transaction]);
+  let pathPart = encodeUint32(pathComponent);
+  pathPart = [pathComponent.length, ...pathPart];
```
webmaster128: Same here: `pathComponent.length` -> `1`

packages/iov-ledger-bns/src/app.ts
rudi-cilibrasi: yeah this part i started to change and realized i was going a bit too quickly and probably need to switch to more explicit verbose form to allow easier reading / modification... thanks for help webmaster128 lets just try next time we catch each other online at the same time to finish this quick together. i have the c code pretty well studied but the JS side is slowing me down now on compile errors in TS etc.

packages/iov-ledger-bns/src/app.ts
rudi-cilibrasi: no i gave up after some TS compile struggling figured you would know better as you know the type system that is working pretty well in iov-core but i have not learned it well yet.

ethanfrey:
What is the status? It seems like a good review from Simon. Did you make those fixes and rebase?

Or do you want one of us to fix this up and test it?
I'm blocked on https://github.com/iov-one/ledger-bns/issues/62 for testing it locally. So if you want to spend 30 minutes to document it, I can probably get this to build and pass.

rudi-cilibrasi:
i think it's better if one of you throws this away and restarts. the basic idea is simple and i'm sure we've discussed it in details before. it is just to add a single byte to indicate the number of path components. We want to add a "1" specifically on the first test so that the result is supposed to be unchanged. Only an input APDU format change for the first agreement test. It is logically necessary and I think obvious after a moment's thought that we need to prepend (not append) the length byte.  It is also pretty obvious to me that it should be literally 0x01 and not, say 0x04 (the number of bytes). This was the format enhancement we decided upon and what I tried to go for in the C but currently having a hard time testing in iov-core. Some of Simon's suggestions worked but not well enough yet for me to get anything close to passing IMO. I've switched to trying to see if the C works by modifying the old "nodeconnect" in the ledger-bns repo where I was assigned and where I made the first tests before the first typescript iov-core integration. unfortunately the old "nodeconnect" version was also a bit out of date so no good news yet but still trying.

ethanfrey:
Hi Rudi, I just saw this note.
Don't struggle with node-connect. I can take a stab at it tomorrow.
At least the code

webmaster128:
Superseded by #445

PR #443, opened 2018-10-15
Update createIdentity interface

### LINKED ISSUE #302: Better handling of paths in HD keyrings

webmaster128:
Closes #302 

* The `options` argument is non-optional everywhere now
* `KeyringEntry` implementations use type`unknown` to avoid requiring the knowledge of every possible type
* `UserProfile` and `KeyringEntry` use the stricter type `Ed25519KeyringEntry | ReadonlyArray<Slip10RawIndex> | number` to support developers at the cost of having a centralized and exhaustive list of possible types

### DIFF
--- a/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
+++ b/packages/iov-keycontrol/src/keyring-entries/ed25519.ts
@@ -120,10 +120,11 @@ export class Ed25519KeyringEntry implements KeyringEntry {
     this.labelProducer.update(label);
   }
 
-  public async createIdentity(keypair?: Ed25519Keypair): Promise<LocalIdentity> {
-    if (!keypair) {
+  public async createIdentity(options: unknown): Promise<LocalIdentity> {
+    if (!(options instanceof Ed25519Keypair)) {
       throw new Error("Ed25519.createIdentity requires a keypair argument");
     }
+    const keypair = options;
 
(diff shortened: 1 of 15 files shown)


webmaster128:
CI fails are unrelated to this PR. Will re-run after review

PR #444, opened 2018-10-16
Add pubkey to BcpConnection.getAccount

### LINKED ISSUE #436: Let BcpConnection.getAccount accept a pubkey
The interface `getAccount: (account: BcpAccountQuery) => Promise<BcpQueryEnvelope<BcpAccount>>;` remains unchanged but `BcpAccountQuery` gets an additional type, like:

```ts
declare type BcpAccountQuery = BcpAddressQuery | BcpPubkeyQuery | BcpValueNameQuery;
```

`BcpPubkeyQuery` must be created, see `BcpAddressQuery`. Pubkey should be of type `PublicKeyBundle`.

BNS, Lisk and Rise should support query by pubkey too (either by conversion to address or by querying by pubkey directly)

LucasIsasmendi:
Closes #436

### DIFF
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -92,7 +92,7 @@ export interface BcpValueNameQuery {
 }
 
 export interface BcpPubkeyQuery {
-  readonly publicKey: PublicKeyBundle;
+  readonly pubkey: PublicKeyBundle;
 }
 
 export type BcpAccountQuery = BcpAddressQuery | BcpPubkeyQuery | BcpValueNameQuery;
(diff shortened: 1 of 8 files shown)


packages/iov-bcp-types/src/bcp.ts
```diff
@@ -91,13 +91,25 @@ export interface BcpValueNameQuery {
   readonly name: string;
 }
 
-export type BcpAccountQuery = BcpAddressQuery | BcpValueNameQuery;
+export interface BcpPubkeyQuery {
+  readonly publicKey: PublicKeyBundle;
```
webmaster128: Can we try to use "pubkey" as a single work? This makes so much stuff simpler and shorter in the crypto world

packages/iov-bcp-types/src/bcp.ts
```diff
@@ -91,13 +91,25 @@ export interface BcpValueNameQuery {
...
+}
+
+export type BcpAccountQuery = BcpAddressQuery | BcpPubkeyQuery | BcpValueNameQuery;
 
 // a type checker to use in the account-based queries
 export function isAddressQuery(query: BcpAccountQuery): query is BcpAddressQuery {
   return (query as BcpAddressQuery).address !== undefined;
 }
 
+export function isPublicKeyQuery(query: BcpAccountQuery): query is BcpPubkeyQuery {
```
webmaster128: isPubkeyQuery

packages/iov-bcp-types/src/bcp.ts
LucasIsasmendi: We should unified this concept, because we are using multiple names in the same repo

webmaster128: As discussed in Slack, we make "pubkey" an atomic word and use it at least in all new code and try to convert old code step by step

PR #445, opened 2018-10-16
Update ledger path

ethanfrey:
Replaces #438

The first commit is a super-minimal change that should work. Please try that against the ledger. It just prepends the byte '0x1' before the 4 byte path component. If this doesn't work, I understood the protocol wrong (it is 1 byte length prefix, not 2, right?)

The second commit is a refactor to abstract this to encode arbitrary paths, even if we only use one segment now, it makes it easy to expand later.

@rudi-cilibrasi i haven't run this against the newly compiled ledger app to test yet, please validate or wait some time for me to get set up.

### DIFF
--- a/packages/iov-ledger-bns/src/app.ts
+++ b/packages/iov-ledger-bns/src/app.ts
@@ -12,7 +12,7 @@ const cmdAppVersion = 0xca;
 
 export function getPublicKeyWithIndex(transport: TransportNodeHid, i: number): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i);
-  const chunk = buildPathPrefix([pathComponent]);
+  const chunk = buildPrefixedPath([pathComponent]);
   return sendChunks(transport, appCode, cmdPubkeyWithPath, chunk);
 }
 
@@ -22,17 +22,17 @@ export function signTransactionWithIndex(
   i: number,
 ): Promise<Uint8Array> {
   const pathComponent = Slip10RawIndex.hardened(i);
-  const data = new Uint8Array([...buildPathPrefix([pathComponent]), ...transaction]);
+  const data = new Uint8Array([...buildPrefixedPath([pathComponent]), ...transaction]);
   return sendChunks(transport, appCode, cmdSignWithPath, data);
 }
 
 // construct a binary length-prefixed path to send to the ledger
-function buildPathPrefix(path: ReadonlyArray<Slip10RawIndex>): Uint8Array {
+function buildPrefixedPath(path: ReadonlyArray<Slip10RawIndex>): Uint8Array {
   // I'm sure there is a better way to do this, but...
   let res = new Uint8Array([path.length]);
   // note: as an example how to combine Uint8Arrays, look at appendSignBytes in iov-bns/src/util.ts
-  for (const step of path.map(slip => slip.toBytesBigEndian())) {
-    res = Uint8Array.from([...res, ...step]);
+  for (const componentBytes of path.map(component => component.toBytesBigEndian())) {
+    res = Uint8Array.from([...res, ...componentBytes]);
   }
   return res;
 }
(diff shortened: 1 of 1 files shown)


packages/iov-ledger-bns/src/app.ts
```diff
@@ -11,20 +11,32 @@ const cmdPubkeyWithPath = 5;
...
-  const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  const data = new Uint8Array([...encodeUint32(pathComponent), ...transaction]);
+  const pathComponent = Slip10RawIndex.hardened(i);
+  const data = new Uint8Array([...buildPathPrefix([pathComponent]), ...transaction]);
   return sendChunks(transport, appCode, cmdSignWithPath, data);
 }
 
+// construct a binary length-prefixed path to send to the ledger
+function buildPathPrefix(path: ReadonlyArray<Slip10RawIndex>): Uint8Array {
+  // I'm sure there is a better way to do this, but...
```
webmaster128: haha, I see what you mean. I don't see an obvious way to improve and it looks correct.

packages/iov-ledger-bns/src/app.ts
```diff
@@ -11,20 +11,32 @@ const cmdPubkeyWithPath = 5;
...
+  let res = new Uint8Array([path.length]);
+  // note: as an example how to combine Uint8Arrays, look at appendSignBytes in iov-bns/src/util.ts
+  for (const step of path.map(slip => slip.toBytesBigEndian())) {
```
webmaster128: I suggest `step` -> `componentBytes` and `slip` -> `component` where "component" is short for "path component"

packages/iov-ledger-bns/src/app.ts
```diff
@@ -11,20 +11,32 @@ const cmdPubkeyWithPath = 5;
...
 ): Promise<Uint8Array> {
-  const pathComponent = Slip10RawIndex.hardened(i).asNumber();
-  const data = new Uint8Array([...encodeUint32(pathComponent), ...transaction]);
+  const pathComponent = Slip10RawIndex.hardened(i);
+  const data = new Uint8Array([...buildPathPrefix([pathComponent]), ...transaction]);
   return sendChunks(transport, appCode, cmdSignWithPath, data);
 }
 
+// construct a binary length-prefixed path to send to the ledger
+function buildPathPrefix(path: ReadonlyArray<Slip10RawIndex>): Uint8Array {
```
webmaster128: `buildPathPrefix` -> `buildPrefixedPath`

rudi-cilibrasi:
well it built and i tested it. seems to not work. it did not make any signing attempts that i noticed but maybe it fetched a public key i am not sure. therefore i attach the entire test log in this comment.  my guess is the C side i did needs a bit more adjustment as it's never been tested at all before i bet it's still got an off by one or sth somewhere.
[testlog.txt](https://github.com/iov-one/iov-core/files/2483882/testlog.txt)

ethanfrey:
Rudi, what flags did you run it with?

LEDGER_ALL or only a subset?

It seems like all the public key queries worked fine, just an error on signing. Nothing appeared on the screen I guess. Did the app freeze? Or just ignore it?

Since PubKey works, I assume that you read the length prefix slightly different in the two forms. Or maybe you miscalculated the offset for the beginning of transaction. Let me know if you find anything, I can look with a magnifying glass later, but tomorrow is workshop day

webmaster128:
I did not yet merge because I wait for feedback if there is a stable Ledger app that is compatible with this change. Please let me know when the app is ready.

ethanfrey:
Okay, I was having build trouble and then just focused on frontend this week. More than enough to do there. I was hoping Rudi would give more feedback with the updated ledger app.

webmaster128:
Closing as next path upgrade will still be single account index (`m/44'/234'/i`) and multi variables will not be necessary for the foreseeable future

PR #448, opened 2018-10-19
Remove Long imports from definition files

### LINKED ISSUE #446: Compiler option esModuleInterop: true is required for users of IOV-Core
For users of IOV-Core, we should not make assumptions on the compiler settings.

```
$ yarn init --yes
$ yarn add typescript --dev
$ yarn add @iov/core
$ echo '{ "compilerOptions": { "target": "es6", "module": "commonjs" }}' > tsconfig.json
$ echo "import { ChainId } from '@iov/core';" > index.ts
$ ./node_modules/.bin/tsc
node_modules/@iov/bcp-types/types/transactions.d.ts:1:8 - error TS1192: Module '"/private/tmp/bla/node_modules/@types/long/index"' has no default export.

1 import Long from "long";
         ~~~~

node_modules/@iov/bns/types/types.d.ts:1:8 - error TS1192: Module '"/private/tmp/bla/node_modules/@types/long/index"' has no default export.

1 import Long from "long";
         ~~~~

```

can be fixed by

```
$ echo '{ "compilerOptions": { "target": "es6", "module": "commonjs", "esModuleInterop": true }}' > tsconfig.json
$ ./node_modules/.bin/tsc
```

webmaster128:
This ensures users of IOV-Core can set `esModuleInterop` to any value

Closes #446


packages/iov-encoding/src/integers.spec.ts
```diff
@@ -128,40 +128,40 @@ describe("Integers", () => {
...
-      expect(Int53.fromString("1").asString()).toEqual("1");
-      expect(Int53.fromString("9007199254740991").asString()).toEqual("9007199254740991");
+      expect(Int53.fromString("0").toString()).toEqual("0");
+      expect(Int53.fromString("1").toString()).toEqual("1");
+      expect(Int53.fromString("9007199254740991").toString()).toEqual("9007199254740991");
 
-      expect(Int53.fromString("-1").asString()).toEqual("-1");
-      expect(Int53.fromString("-9007199254740991").asString()).toEqual("-9007199254740991");
+      expect(Int53.fromString("-1").toString()).toEqual("-1");
+      expect(Int53.fromString("-9007199254740991").toString()).toEqual("-9007199254740991");
```
LucasIsasmendi: We should test also "Input not in int53 range"

webmaster128: Jap, test added. This was only tested in for the dirct constructor from number before

PR #449, opened 2018-10-19
Add bech32 support

### LINKED ISSUE #339: Provide support for bech32 encoding of address bytes (encode/decode)
Bech32 is the future: https://en.bitcoin.it/wiki/Bech32
"Readable" addresses with error correction.

This would replace displaying hex-strings for addresses, not the human-names
Moved here from: https://github.com/iov-one/weave/issues/27

LucasIsasmendi:
Closes #339


packages/iov-encoding/src/encoding.spec.ts
```diff
@@ -313,3 +313,21 @@ describe("Encoding", () => {
     });
   });
 });
+
+describe("Bech32", () => {
+  const lskAddressUtf8 = Encoding.toUtf8("6472030874529564639L");
+  const riseAddressUtf8 = Encoding.toUtf8("6472030874529564639R");
+  const ethAddressUtf8 = Encoding.toUtf8("0x9D4e856E572E442f0A4b2763e72d08A0E99D8deD");
```
webmaster128: I think this example is very irritating because you don't encode the 20 bytes Ethereum address but the 42 bytes string. Better go for `const ethAddressRaw = Encoding.fromHex("9D4e856E572E442f0A4b2763e72d08A0E99D8deD")`

packages/iov-encoding/src/encoding.spec.ts
```diff
@@ -313,3 +313,21 @@ describe("Encoding", () => {
     });
   });
 });
+
+describe("Bech32", () => {
+  const lskAddressUtf8 = Encoding.toUtf8("6472030874529564639L");
+  const riseAddressUtf8 = Encoding.toUtf8("6472030874529564639R");
```
webmaster128: The Lisk and RISE tests are irritating because there is no proper binary representation for those, so using them as example data is not very helpful. We will also not use Bech32 to encode Lisk and RISE addresses

PR #450, opened 2018-10-19
Add return value to UserProfile.addEntry() for ease of use

ethanfrey:
A small enhancement to aid usability from our discussions on Thursday

### DIFF
--- a/packages/iov-keycontrol/src/userprofile.spec.ts
+++ b/packages/iov-keycontrol/src/userprofile.spec.ts
@@ -119,6 +119,21 @@ describe("UserProfile", () => {
     expect(profile.getIdentities(entry3.id)).toBeTruthy();
   });
 
+  it("addEntry returns wallet info", () => {
+    const profile = new UserProfile();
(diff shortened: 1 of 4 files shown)


packages/iov-keycontrol/src/userprofile.spec.ts
```diff
@@ -120,6 +120,21 @@ describe("UserProfile", () => {
     expect(profile.getIdentities(entry3.id)).toBeTruthy();
   });
 
+  it("addEntry returns wallet info", () => {
+    const profile = new UserProfile();
+    const entry1 = Ed25519HdWallet.fromMnemonic("melt wisdom mesh wash item catalog talk enjoy gaze hat brush wash");
+    const entry2 = Ed25519HdWallet.fromMnemonic("perfect clump orphan margin memory amazing morning use snap skate erosion civil");
```
webmaster128: Time to get rid of the term "entry". But this is no this PR's problem

ethanfrey:
Okay, think it is done

PR #432, opened 2018-10-10
Use local devnet for Lisk in tests

webmaster128:
Let's see if Travis is happy with the docker-compose setup from https://github.com/LiskHQ/lisk/tree/1.2.0/docker

PR #457, opened 2018-10-23
Eth chain id and height

### LINKED ISSUE #402: Connect to ethereum node and get chainId and height
Beginning of implementation of a BcpConnection.

Feel free to use any libraries that are compatible with Apache license. If there is great code that is not, please discuss, maybe we more ethereum adapter to separate repo.

This story should include a test case that runs in CI against a local ethereum node. Please use https://github.com/trufflesuite/ganache-cli unless you have a really good alternative (I had good experience with the old version https://www.npmjs.com/package/ethereumjs-testrpc)

LucasIsasmendi:
Closes #402


packages/iov-lisk/tsconfig.json
```diff
@@ -6,6 +6,6 @@
     "rootDir": "src"
   },
   "include": [
-    "src/**/*"
+    "src/**/*", "../iov-ethereum/src/utils.ts"
```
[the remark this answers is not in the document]
LucasIsasmendi: Yes, I ran some code to generate type folder and changes this

packages/iov-ethereum/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+To run Ethereum tests locally, you need to install ganache and set as follow:
+- RPC Server: `http://127.0.0.1:7545`
+- Network ID: `5777`
+- Mnemonic: `oxygen fall sure lava energy veteran enroll frown question detail include maximum`
+
+Or run the followin script:
+```
+cd packages/iov-ethereum
+source startGanache.sh
```
webmaster128: Using source for scripts is very problematic. I know we use it for BOV and tendermint, but that is only because we set environment variables that we need later. Problems with source:
- you run in the shell of the users, not the one you set after `#!`. So e.g. I cannot use bash features in script because I use zsh
- you cannot stop the script in case of an error in one command, because `exit 1` does not end the script but your current console

In https://github.com/iov-one/iov-core/pull/432/files#diff-a09db41a73f7fd39327cf6a25e928626R45 I run a script and set the env variable explicitly, which is better. In this case, we don't need env variables as far as I can see because the stop script uses the process ID from the process list.

packages/iov-ethereum/startGanache.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
webmaster128: Please use this battle tested bash header for all shell scripts:

```
#!/bin/bash
set -o errexit -o nounset -o pipefail
```

packages/iov-ethereum/startGanache.sh
```diff
@@ -0,0 +1,23 @@
...
+# Helper script for starting/stopping ganache and running tests
+#
+
+function startupGanache() {
+  GANACHE_SEARCH="$(ps -ef | grep "[g]anache" | awk '{print $2}')"
+  export GANACHE_PID="${GANACHE_SEARCH}"
```
webmaster128: Why is GANACHE_PID explorted? As far as I can see, we don't need it outside

packages/iov-ethereum/src/utils.ts
```diff
@@ -0,0 +1,8 @@
+export function decodeHexQuantity(hexString: string): number {
+  if (hexString === "0x0" || hexString === "0x00") {
```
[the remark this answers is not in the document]
LucasIsasmendi: YEs, json rpc is confusing regarding that, they say `WRONG: 0x0400 (no leading zeroes allowed)` and then they provide the following example: `"result": "0x09184e72a000" // 10000000000000`

packages/iov-ethereum/webpack.web.config.js
```diff
@@ -0,0 +1,21 @@
...
+  {
+    // bundle used for Karma tests
+    target: target,
+    entry: glob.sync("./build/**/*.spec.js"),
+    output: {
+      path: distdir,
+      filename: "tests.js",
+    },
+    plugins: [
+      new webpack.EnvironmentPlugin(['LONG_RUNNING_ENABLED']),
```
webmaster128: The environment variable `LONG_RUNNING_ENABLED` is not used in the new tests, so we don't need to pass it from environment to Karma tests. But what we probably want to have is `ETH_ENV`

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { EthereumConnection } from "./ethereumconnection";
+import { TestConfig } from "./testconfig";
+
+describe("EthereumConnection", () => {
+  const base = TestConfig.base;
+  const nodeChainId = TestConfig.chainId;
+  const minHeight = TestConfig.minHeight;
+
+  it(`can be constructed for ${base}`, () => {
+    const connection = new EthereumConnection(base, nodeChainId);
```
webmaster128: Those tests should be pending if `ETH_ENV` is not set. See bns tests for example code

.travis.yml
```diff
@@ -37,6 +37,7 @@ install:
 before_script:
 - export TSLINT_FLAGS='-c ./tslint_ci.json'
 - export LONG_RUNNING_ENABLED=1
+- ./packages/iov-ethereum/startGanache.sh
```
webmaster128: This is a good place to set ETH_ENV as well. When unset, no ethereum is available (like on Windows CI)

LucasIsasmendi:
It works with ganache-cli from node_modules, should we change it to docker?

packages/iov-ethereum/package-lock.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@iov/ethereum",
```
webmaster128: Please don't check in npm lockfiles

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { EthereumConnection } from "./ethereumconnection";
+import { TestConfig } from "./testconfig";
+
+const skipTests = (): boolean => !process.env.ETHEREUM_ENABLED;
+
+const pendingWithoutEthereum = () => {
```
webmaster128: Could you use real functions instead of arrow function? I know this is how existing code is written, but we should use classic `function`s that follow a consistent style with classes and enums and are way easier to parse when reading a lot of code

webmaster128:
> It works with ganache-cli from node_modules, should we change it to docker?

I think it is great if we only need a single executable in node_modules. No need to make it more complex

packages/iov-ethereum/webpack.web.config.js
```diff
@@ -15,7 +15,7 @@ module.exports = [
       filename: "tests.js",
     },
     plugins: [
-      new webpack.EnvironmentPlugin(['LONG_RUNNING_ENABLED']),
+      new webpack.EnvironmentPlugin(['ETH_ENV']),
```
webmaster128: We need ETHEREUM_ENABLED as well. Otherwise the browser tests do not see the environment variable

PR #461, opened 2018-10-24
[iov-faucets] Let BovFaucet take Bech32 addresses

webmaster128:


webmaster128:
> One question, there is a prefix on bech32 (eg. tiov), but that doesn't seem to be checked anywhere? Not just in this module, but also in the other bns code.

I'm waiting for you validator function to include this check

PR #462, opened 2018-10-24
Add isValidAddress to TxReadCodec and all implementations

### LINKED ISSUE #459: Add isValidAddress to TxCodec
Isabella mentioned this when working on the faucet.
I came across it today when working on the wallet.

An easy-to-implement and valuable addition to the API.
I would be happy enough to implement tomorrow if this could squeeze into 0.7.0.

In `iov-bcp-types`, define:

```
export interface TxReadCodec {
    readonly parseBytes: (bytes: PostableBytes, chainId: ChainId) => SignedTransaction;
    readonly keyToAddress: (key: PublicKeyBundle) => Address;
    // apply chain-specific validation rules
    readonly isValidAddress: (address: Address) => boolean;
}
```

or more explicitly (but perhaps too complex)

```
export interface TxReadCodec {
    readonly parseBytes: (bytes: PostableBytes, chainId: ChainId) => SignedTransaction;
    readonly keyToAddress: (key: PublicKeyBundle) => Address;
    readonly isValidAddress: (address: string) => address is Address;
}
```

ethanfrey:
Closes #459

### DIFF
--- a/packages/iov-lisk/src/derivation.spec.ts
+++ b/packages/iov-lisk/src/derivation.spec.ts
@@ -1,6 +1,6 @@
 import { Encoding } from "@iov/encoding";
 
-import { passphraseToKeypair, pubkeyToAddress } from "./derivation";
+import { isValidAddress, passphraseToKeypair, pubkeyToAddress } from "./derivation";
 
 const { fromHex } = Encoding;
 
@@ -30,6 +30,30 @@ describe("Derivation", () => {
       // https://testnet-explorer.lisk.io/address/10176009299933723198L
       const pubkey = fromHex("06ad4341a609af2de837e1156f81849b05bf3c280940a9f45db76d09a3a3f2fa");
       expect(pubkeyToAddress(pubkey)).toEqual("10176009299933723198L");
+    });
+  });
+
+  describe("isValidAddress", () => {
+    it("works for valid numbers within unsigned int64 range", () => {
+      expect(isValidAddress("1234567890L")).toBeTruthy();
+      expect(isValidAddress("6076671634347365051L")).toBeTruthy();
+      expect(isValidAddress("10176009299933723198L")).toBeTruthy();
+      // 2**64-1
+      expect(isValidAddress("18446744073709551615L")).toBeTruthy();
+    });
+
+    it("rejects malformed addresses", () => {
+      // invalid ending
+      expect(isValidAddress("1234567890R")).toBeFalsy();
+      // leading 0
+      expect(isValidAddress("01234567821L")).toBeFalsy();
+      // decimal
+      expect(isValidAddress("12345.6788L")).toBeFalsy();
+      // string values
+      expect(isValidAddress("12some45L")).toBeFalsy();
+      // 2**64, 2**64 + 1
+      expect(isValidAddress("18446744073709551616L")).toBeFalsy();
+      expect(isValidAddress("18446744073709551617L")).toBeFalsy();
     });
   });
 });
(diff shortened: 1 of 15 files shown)


packages/iov-lisk/src/derivation.spec.ts
```diff
@@ -32,4 +32,28 @@ describe("Derivation", () => {
       expect(pubkeyToAddress(pubkey)).toEqual("10176009299933723198L");
     });
   });
+
+  describe("isValidAddress", () => {
+    it("works for valid numbers within unsigned int64 range", () => {
+      expect(isValidAddress("1234567890L")).toBeTruthy();
```
webmaster128: `toBeTruthy()` -> `toEqual(true)` because we don't want 1 or "blabla" to pass

packages/iov-lisk/src/derivation.spec.ts
```diff
@@ -32,4 +32,28 @@ describe("Derivation", () => {
...
+      expect(isValidAddress("6076671634347365051L")).toBeTruthy();
+      expect(isValidAddress("10176009299933723198L")).toBeTruthy();
+      // 2**64-1
+      expect(isValidAddress("18446744073709551615L")).toBeTruthy();
+    });
+
+    it("rejects malformed addresses", () => {
+      // invalid ending
+      expect(isValidAddress("1234567890R")).toBeFalsy();
```
webmaster128: `toBeFalsy()` -> `toEqual(false)` because we don't want 0, null, "" or undefined to pass

packages/iov-lisk/src/derivation.ts
```diff
@@ -25,3 +25,19 @@ export function pubkeyToAddress(pubkey: Uint8Array): Address {
...
+
+export function isValidAddressWithEnding(address: string, ending: string): boolean {
+  const tail = address.slice(-ending.length);
+  if (tail !== ending) {
+    return false;
+  }
+  const addressString = address.slice(0, -ending.length);
+  // parse as unsigned base 10 number and verify re-encoding matches encoding
+  // this is no leading zeros, no decimals, within 0 <= addressString <= 2**64-1
+  const addressNumber = Long.fromString(addressString, true, 10); // true => unsigned
```
webmaster128: Long behaves unreliable wenn passing string values > 64 bit. See e.g.

> Long.fromString("18446744073709551616", true).greaterThan(Long.fromString("18446744073709551615", true))
false

So we need to use bn.js

packages/iov-rise/src/derivation.ts
```diff
@@ -25,3 +25,20 @@ export function pubkeyToAddress(pubkey: Uint8Array): Address {
   const addressString = Long.fromBytesLE(firstEightBytes, true).toString(10) + "R";
   return addressString as Address;
 }
+
+export function isValidAddress(address: string): boolean {
+  return isValidAddressWithEnding(address, "R");
+}
+
+// copied from iov-lisk, maybe just import this instead?
+export function isValidAddressWithEnding(address: string, ending: string): boolean {
```
webmaster128: Nice duplication, like it. It is a perfect first step to extract into a shared dpos package. I'd avoid adding cross-dependencies between the chains.

packages/iov-lisk/src/derivation.ts
```diff
@@ -25,3 +25,19 @@ export function pubkeyToAddress(pubkey: Uint8Array): Address {
...
+  return addressNumber.toString(10) === addressString;
```
webmaster128: Explicit uint64 range check is missing here.

ethanfrey: Long will only return up to 2**64 -1.
That means any number larger will be truncated somehow and return a different value.

packages/iov-lisk/src/derivation.ts
ethanfrey: I see your point, but since we care about the equality of `parse with long -> encode to string`, we know this only works with integers between 0 and 2**64-1 with no leading zeros and no decimal points. Otherwise the output is somehow different.

Since this is just an equality test, I think Long is fine here.

PR #469, opened 2018-10-26
Eth query address and nonce

### LINKED ISSUE #403: Query address and nonce for ethereum address
This should extend the BcpConnection started in #402 and be covered with test cases (using the addresses returned on startup).

LucasIsasmendi:
Closes #403

### DIFF
--- a/packages/iov-ethereum/src/parse.ts
+++ b/packages/iov-ethereum/src/parse.ts
@@ -1,0 +1,21 @@
+import { TokenTicker } from "@iov/bcp-types";
+
+import { constants } from "./constants";
+
+export interface AmountFields {
+  readonly whole: number;
+  readonly fractional: number;
+  readonly tokenTicker: TokenTicker;
+}
+
+export class Parse {
+  public static ethereumAmount(total: string): AmountFields {
+    const totalLenght = total.length;
+    const sigFigs = constants.primaryTokenSigFigs;
+    return {
+      whole: Number(total.substring(0, totalLenght - sigFigs).replace(/^0+/, "")) || 0,
+      fractional: Number(total.substring(totalLenght - sigFigs, totalLenght).replace(/^0+/, "")) || 0,
+      tokenTicker: constants.primaryTokenTicker,
+    };
+  }
+}
--- a/packages/iov-ethereum/src/parse.ts
+++ b/packages/iov-ethereum/src/parse.ts
@@ -10,11 +10,10 @@ export interface AmountFields {
 
 export class Parse {
   public static ethereumAmount(total: string): AmountFields {
-    const totalLenght = total.length;
     const sigFigs = constants.primaryTokenSigFigs;
     return {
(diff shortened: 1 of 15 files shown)


packages/iov-ethereum/src/parse.ts
```diff
@@ -0,0 +1,21 @@
...
+  readonly tokenTicker: TokenTicker;
+}
+
+export class Parse {
+  public static ethereumAmount(total: string): AmountFields {
+    const totalLenght = total.length;
+    const sigFigs = constants.primaryTokenSigFigs;
+    return {
+      whole: Number(total.substring(0, totalLenght - sigFigs).replace(/^0+/, "")) || 0,
+      fractional: Number(total.substring(totalLenght - sigFigs, totalLenght).replace(/^0+/, "")) || 0,
```
webmaster128: When you use slice() instead of substring, whis becomes much more readable. slice() supports negative or unset ends, e.g.

`total.substring(0, -sigFigs)`, `total.substring(-sigFigs)`

webmaster128: Number() does not check for integer. Could you wrap string->number into `Int53.fromString(x).toNumber()`? ...which does not work for numbers > 53 bit

webmaster128: Then better `Number.parseInt(x, 10)`

packages/iov-ethereum/src/parse.ts
```diff
@@ -0,0 +1,21 @@
...
+  readonly fractional: number;
+  readonly tokenTicker: TokenTicker;
+}
+
+export class Parse {
+  public static ethereumAmount(total: string): AmountFields {
+    const totalLenght = total.length;
+    const sigFigs = constants.primaryTokenSigFigs;
+    return {
+      whole: Number(total.substring(0, totalLenght - sigFigs).replace(/^0+/, "")) || 0,
```
Isabello:
```suggestion
      whole: Number(total.substring(0, totalLength - sigFigs).replace(/^0+/, "")) || 0,
```

packages/iov-ethereum/src/parse.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface AmountFields {
+  readonly whole: number;
+  readonly fractional: number;
+  readonly tokenTicker: TokenTicker;
+}
+
+export class Parse {
+  public static ethereumAmount(total: string): AmountFields {
+    const totalLenght = total.length;
```
Isabello:
```suggestion
    const totalLength = total.length;
```

packages/iov-ethereum/src/parse.ts
Isabello:
```suggestion
      fractional: Number(total.substring(totalLength - sigFigs, totalLength).replace(/^0+/, "")) || 0,
```

PR #472, opened 2018-10-31
Start and test IovFaucet

### LINKED ISSUE #451: Build proof of concept faucet
# Architecture:
 
The faucet will run as a node application on a server. The initialization and configuration will happen via command-line by an admin, who also has access to some large number of tokens to seed the faucet. The faucet will expose a simple http server with at least two endpoints - `/status` to check liveness and `/faucet` to request tokens for a provided address. 

Ideally, we would like a faucet to be able to provide tokens for many chains simultaneously. However, a simpler iteration, and plenty for starting, is to allow it to provide tokens to any one blockchain. Thus, both the url for the node, as well as the type of blockchain (bns, bcpd, lisk, rise), should somehow be configurable on initialization.

## Endpoints

`GET /status`

Returns an object with the status, nodeURL and chainId it serves.
```
{ 
  status: “ok”, 
  nodeUrl: “node-url”, 
  chainId: “chainId”
  } 
````
`POST /faucet`

Accepts a json post something like: 

```
{ 
  address: “chain-specific-address-here”, 
  token: “IOV”, 
  chainId:“friendnet-fast” // (ChainId is not really required, but you can enforce it matches the chain you connected to. You could also provide a default for token, but nice to let them choose (and shows they know what they are asking for).
}
```
Upon a valid request, the faucet should query the given address on the connected chain. If it is a valid address, but has no tokens, then a “send transaction” should be executed signed by one of the many accounts on a round-robin basis

### PR DESCRIPTION
**Pull request by webmaster128**
Closes #451

### DIFF
--- a/packages/iov-faucets/src/iovfaucet.ts
+++ b/packages/iov-faucets/src/iovfaucet.ts
@@ -1,0 +1,25 @@
+import axios from "axios";
+
+import { Address, TokenTicker } from "@iov/bcp-types";
+
+export class IovFaucet {
+  private readonly baseUrl: string;
+
+  constructor(baseUrl: string) {
+    if (!baseUrl.match(/^https?:\/\//)) {
+      throw new Error("Expected base url to start with http:// or https://");
+    }
+
+    // Strip trailing /
+    const strippedBaseUrl = baseUrl.replace(/(\/)+$/, "");
+    this.baseUrl = strippedBaseUrl;
+  }
+
+  public async credit(address: Address, ticker: TokenTicker): Promise<void> {
+    const body = {
+      address: address,
+      ticker: ticker,
+    };
+    await axios.post(this.baseUrl + "/credit", body);
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.8.0
 
+* @iov/faucets: Add `IovFaucet` to connect to the new faucet application
+
 Deprecations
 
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by ethanfrey** [comment]
I didn't look at the code yet, but this makes me happy! :)

CI Testing on the whole ui flow in the react app cannot proceed beyond signup as it depends on a faucet (and we don't want to run tests on a real testnet)

File: packages/iov-faucets/src/iovfaucet.ts [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+    const strippedBaseUrl = baseUrl.replace(/(\/)+$/, "");
+    this.baseUrl = strippedBaseUrl;
+  }
+
+  public async credit(address: Address, ticker: TokenTicker): Promise<void> {
+    const body = {
+      address: address,
+      ticker: ticker,
+    };
+    await axios.post(this.baseUrl + "/credit", body);
```
LucasIsasmendi: In bovfaucet we have an url (with /faucet) and in iovfaucet we hace url + api (/credit), can we standardize this posts?

LucasIsasmendi: No response/return from post?

webmaster128: > can we standardize this posts?

`baseUrl` is the new default, allowing the class to add other functions later (like available tokens from /status).

But since `BovFaucet` is already deprecated, I don't want to do any incompatible change there anymore.

> No response/return from post?

Right, just `void` or an error. No message is good message ;)

PR #486, opened 2018-11-02
Extract class Parse into @iov/dpos

webmaster128:


packages/iov-dpos/src/parse.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+describe("Parse", () => {
+  it("can parse zero amount", () => {
+    const expected: Amount = { whole: 0, fractional: 0 };
+    expect(Parse.parseAmount("0")).toEqual(expected);
+    expect(Parse.parseAmount("00")).toEqual(expected);
+    expect(Parse.parseAmount("000000000")).toEqual(expected);
+  });
+
+  it("can parse 1 LSK", () => {
```
LucasIsasmendi: Maybe instead of state amounts as LSK token, use LSK/RISE description, since is a shared function

webmaster128: oh, indeed. I moved the token name out of the package. Will update

webmaster128: https://github.com/iov-one/iov-core/commit/42a699a6820bea2756bb94baec328283980d8730

PR #487, opened 2018-11-02
Move all blockchain scripts into repo root

webmaster128:
They don't belong to the npm packages but to our CI setup


LucasIsasmendi:
Readme should be updated to run scripts from script folder

webmaster128:
thx, updated

Issue #490, opened 2018-11-06
Implement binding between Ledger wallet and Ledger device


webmaster128:
A given wallet should be connected to one device only (otherwise labeling etc. may be off). It should be possible to have ultiple wallets that connect to the same device.

PR #492, opened 2018-11-06
Use random, fixed IDs for LedgerSimpleAddressWallet 

### LINKED ISSUE #334: @iov/ledger-bns should get id from first connection with app
#332 added `id` to all keyring entries, which is tied to the cryptographic secret stored inside. (no private info is revealed, but we can do an equality check as to whether they have the same secret information inside).

With the ledger, we should grab for example, a hash for the publickey at index 0. Unfortunately we can only do this asynchrnously once it is connected with the app open.

We should check every time we connect to the app. If the id was unset before, set to the id derived from the current ledger. If we connect to a ledger and it has a different derived id that what is stored in our currenly open keyring, maybe throw an error?We need a separatekeyring, as the locally stored public keys don't reflect the ledger device.

webmaster128:
Closes #334

See https://github.com/iov-one/iov-core/issues/490 for alternative aproach

### DIFF
--- a/packages/iov-ledger-bns/.prettierrc.json
+++ b/packages/iov-ledger-bns/.prettierrc.json
@@ -1,12 +1,4 @@
 {
   "printWidth": 110,
-  "trailingComma": "all",
-  "overrides": [
-    {
-      "files": "*.spec.ts",
-      "options": {
-        "printWidth": 5000
-      }
-    }
-  ]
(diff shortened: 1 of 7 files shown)


webmaster128:
Tested with `LEDGER_ALL=1`, all green

webmaster128:
CI fail unrelated to this change

Issue #404, opened 2018-10-05
Convert secp public key to ethereum address


ethanfrey:
This is the beginning of TxCodec for ethereum.

This should produce the compatible bytes to query ethereum rpc, and validate in integration tests as in #403

PR #505, opened 2018-11-07
Add MetaMask "HD Key Tree" derivation paths

webmaster128:
This together with https://github.com/iov-one/iov-core/issues/404 allows us to generate MetaMask addresses from a mnemonic

### DIFF
--- a/packages/iov-keycontrol/src/hdpaths.ts
+++ b/packages/iov-keycontrol/src/hdpaths.ts
@@ -10,4 +10,27 @@ export class HdPaths {
     const iovPurpose = 4804438;
     return [Slip10RawIndex.hardened(iovPurpose), Slip10RawIndex.hardened(index)];
   }
+
+  /**
+   * Only use this for the real BIP-0044 with
+   * - 5 component path (m / purpose' / coin_type' / account' / change / address_index)
+   * - no ed25519 support (due to the use of unhardened path components)
+   *
+   * Don't be misled by people calling their path BIP44 in cases where it is not.
+   */
+  public static bip44(
+    coinType: number,
+    account: number,
+    change: number,
+    address: number,
+  ): ReadonlyArray<Slip10RawIndex> {
+    const bip44Purpose = 44;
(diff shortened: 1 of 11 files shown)


webmaster128:
@LucasIsasmendi this one is easier to review commit by commit

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -10,4 +10,37 @@ export class HdPaths {
...
+      Slip10RawIndex.hardened(bip44Purpose),
+      Slip10RawIndex.hardened(coinType),
+      Slip10RawIndex.hardened(account),
+      Slip10RawIndex.normal(change),
+      Slip10RawIndex.normal(address),
+    ];
+  }
+
+  /**
+   * The default MetaMask derivation path
```
ethanfrey: :+1:

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -10,4 +10,37 @@ export class HdPaths {
     const iovPurpose = 4804438;
     return [Slip10RawIndex.hardened(iovPurpose), Slip10RawIndex.hardened(index)];
   }
+
+  /**
+   * Only use this for the real BIP-0044 with
```
ethanfrey: Like this.

Can we add a slip10hardened() function or something similar to generate ed25519 addresses?
I agree with a different name to differentiate them, but a similar helper for bns/lisk would be great.

webmaster128: slip10 does not specify any path sheme. What we want is

```
trezorStyleBip44LikeAccountBasedEd25519(coinType: number, a: number)
```

which covers Stellar, Lisk, NEM, Monero as documented here https://github.com/trezor/trezor-core/tree/master/docs/coins as well as RISE and maybe others.

Anyway, we need to be much more specific in what kind of paths we need. For BNS no final path scheme other than simple address was specified.

Issue #514, opened 2018-11-08
Split @iov/tendermint-types


webmaster128:
@iov/tendermint-types somehow became home of very generic types we use everywhere. But it should be possible to use IOV-Core without Tendermint-chains and without the term "tendermint" in packages.

In there we currently have

```
// keys
enum Algorithm
type PrivateKeyBytes
interface PrivateKeyBundle
type PublicKeyBytes
interface PublicKeyBundle

// network
type ChainId

// transactions
type SignatureBytes
interface SignatureBundle
type PostableBytes
type TxId
interface TxQuery
interface Tag
```

Tendermint types used by tendermint:

| Type name  | used in @iov/tendermint-rpc | used in @iov/bcp-types | used in @iov/bns |
| ------------- | ------------- | ------------- | ------------- |
| Algorithm  | ✅ | (implicitly in PublicKeyBundle) | ✅ |
| PrivateKeyBytes  | - | - |  ✅ |
| PrivateKeyBundle  | - | -  | ✅ |
| PublicKeyBytes | ✅ | (implicitly in PublicKeyBundle) | ✅ |
| PublicKeyBundle | ✅ | ✅ | ✅ |
| | | | |
| ChainId | ✅ | ✅ | ✅ |
| | | | |
| SignatureBytes | ✅ | ✅ | ✅ |
| SignatureBundle | ✅ | - | - |
| PostableBytes | ✅ | ✅ | ✅ |
| TxId | ✅ | ✅ | ✅ |
| TxQuery  | ✅ | ✅ | ✅ |
| Tag | ✅ | ✅ | ✅ |

Ideas for improvements:

1. `Tag` and `TxQuery` should be split to allow having different types in BcpConnection and in @iov/tendermint-rpc: `TendermintQueryTag` / `BcpQueryTag` and `TendermintTxQuery` / `BcpTxQuery`. 
2. `PrivateKeyBytes` and `PrivateKeyBundle` should move to @iov/bns and should not be exported as they are only used in bns codec tests
3. Make `SignatureBundle` available in @iov/tendermint-rcp only and rename to `VoteSignatureBundle`

The rest can go to a package called `@iov/base-types` for now.

ethanfrey:
I think these were all bcp-types and you pulled them out, thus the name tendermint-types.

I am fine with 2 and 3.
I also like renaming this package or just rejoining what is left with iov-bcp-types for simplicity.

[...]

PR #516, opened 2018-11-08
Split @iov/tendermint-types

webmaster128:
Everything from #514 except the package renaming.

The only 7 base types shared between tendermint and BCP now are

```ts
export declare enum Algorithm {
    Ed25519 = "ed25519",
    Secp256k1 = "secp256k1"
}
export declare type PublicKeyBytes = Uint8Array & As<"public-key">;
export interface PublicKeyBundle {
    readonly algo: Algorithm;
    readonly data: PublicKeyBytes;
}
/** Used to differentiate a blockchain. Should be alphanumeric or -_/ and unique */
export declare type ChainId = string & As<"chain-id">;
export declare type SignatureBytes = Uint8Array & As<"signature">;
export declare type PostableBytes = Uint8Array & As<"postable">;
export declare type TxId = Uint8Array & As<"txid">;
```


packages/iov-bcp-types/src/bcp.ts
```diff
@@ -1,6 +1,6 @@
 import { Stream } from "xstream";
 
-import { ChainId, PostableBytes, PublicKeyBundle, Tag, TxId, TxQuery } from "@iov/tendermint-types";
+import { ChainId, PostableBytes, PublicKeyBundle, TxId } from "@iov/tendermint-types";
```
ethanfrey: Weren't you going to rename `@iov/tendermint-types` to `@iov/base-types`?

packages/iov-bns/src/util.ts
```diff
@@ -103,3 +105,15 @@ export function isSwapRelease(
     tx.transaction.kind === TransactionKind.SwapClaim || tx.transaction.kind === TransactionKind.SwapTimeout
   );
 }
+
+export function buildTxQuery(query: BcpTxQuery): QueryString {
```
ethanfrey: Yes, this is a better place for the logic

ethanfrey:
Also, please set the project on the right hand sidebar, so it shows up in the issue tracker

webmaster128:
> Did you change your mind, or is that a different pr?

The renaming would increase the diff dramatically so I decided to do it separately. This keeps the important stuff and the dumb stuff separated for easier review.

PR #519, opened 2018-11-08
Secp pubkey to eth address

### LINKED ISSUE #404: Convert secp public key to ethereum address

LucasIsasmendi:
Closes #404

### DIFF
--- a/packages/iov-ethereum/src/ethereumcodec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.ts
@@ -11,7 +11,7 @@ import {
 } from "@iov/bcp-types";
 import { Keccak256 } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
-import { ChainId, PostableBytes, PublicKeyBundle } from "@iov/tendermint-types";
+import { Algorithm, ChainId, PostableBytes, PublicKeyBundle } from "@iov/tendermint-types";
 
 const { toAscii, toHex } = Encoding;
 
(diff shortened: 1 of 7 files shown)


ethanfrey:
Nice progress.

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -0,0 +1,51 @@
...
+  const addressLower = address.toLowerCase().replace("0x", "") as Address;
+  const addressHash = toHex(new Keccak256(addressLower).digest());
+  let checksumAddress = "0x";
+  for (let i = 0; i < 40; i++) {
+    checksumAddress += parseInt(addressHash[i], 16) > 7 ? addressLower[i].toUpperCase() : addressLower[i];
+  }
+  return checksumAddress as Address;
+}
+
+export const ethereumCodec: TxCodec = {
```
webmaster128: can you export `ethereumCodec` in index.ts please? Then it can be tested from the cli

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error(`Not implemented tx: ${signed}`);
+  },
+  identifier: (signed: SignedTransaction): TransactionIdBytes => {
+    throw new Error(`Not implemented tx: ${signed}`);
+  },
+  parseBytes: (bytes: PostableBytes, chainId: ChainId): SignedTransaction => {
+    throw new Error(`Not implemented bytes: ${bytes}, chainId: ${chainId}`);
+  },
+  keyToAddress: (pubkey: PublicKeyBundle): Address => {
+    const hash = toHex(new Keccak256(pubkey.data).digest());
```
webmaster128: Could you add two checks here:

* `pubkey.algo === Algorithm.Secp256k1`
* `pubkey.data.length === 64`

In both cases just trow an exception.

webmaster128: The function should expect uncompressed pubkeys with prefix byte `04`, so you need to hash `toHex(new Keccak256(pubkey.data.slice(1)).digest());` and add "04" to the test

webmaster128:
Unrelated CI fail

PR #524, opened 2018-11-09
Update BNS codec to 0.9

### LINKED ISSUE #493: Update BNS codec for 0.9.0 support
`@iov/bns` is still using the codec from weave and bcp-demo v0.6.0. These were merged in 0.7.0 and new extensions were added in 0.8.0 and 0.9.0. Luckily they have all been backwards compatible, so the older codec still produces binary transactions compatible with the newer blockchain, but we need to expose the newer features.

- [x] Update the protobuf generation scripts for single repo (moved to #504)
- [ ] Re-generate protobuf definitions
- [ ] Update wrappers in `@iov/bns` to refer to the new naming

The end result should expose the same features as now, but have access to the newer, fuller codec

webmaster128:
Closes #493

### DIFF
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -69,6 +69,7 @@ export interface SwapCounterTx extends BaseTx {
   readonly recipient: RecipientId;
   readonly timeout: number; // number of blocks in the future
   readonly hashCode: Uint8Array; // pulled from the offer transaction
+  readonly memo?: string;
 }
 
 export interface SwapClaimTx extends BaseTx {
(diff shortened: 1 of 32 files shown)


packages/iov-bcp-types/src/transactions.ts
```diff
@@ -69,6 +69,7 @@ export interface SwapCounterTx extends BaseTx {
   readonly recipient: RecipientId;
   readonly timeout: number; // number of blocks in the future
   readonly hashCode: Uint8Array; // pulled from the offer transaction
+  readonly memo?: string;
```
ethanfrey: good call

webmaster128:
👍 thanks for feedback. Changelog was added on master: https://github.com/iov-one/iov-core/commit/29246dafb0a8403c0f77b97a87df2703b560aa8f

Issue #502, opened 2018-11-06
Convert whole and fraction to strings


webmaster128:

ethanfrey:
Fixed point format:
```
{
  "amount": "1234",
  "sigFigs": 2
}
```

is equivalent to "12.34" (amount * 10^-sigFigs)

We must add conversions between FixedPoint and the native serialized format for the blocchains. Also add clone

Helper functions to convert to/from UI "12.34" <-> FixedPoint and functions to add/subtract/compare two FixedPoints can be implemented in client library (wallet-demo) and use native BigInt.

webmaster128:
This can be decomposed into the old form using string operations:

```js
const coin = { amount: "1234", sigFigs: 2, }
const fractional = BigInt(coin.amount.slice(-coin.sigFigs)); // 34n
const whole = BigInt(coin.amount.slice(0, -coin.sigFigs)); // 12n

const coin2 = { amount: "1234", sigFigs: 8, }
const fractional2 = BigInt(coin2.amount.slice(-coin2.sigFigs)); // 1234n
const whole2 = BigInt(coin2.amount.slice(0, -coin2.sigFigs)); // 0n
```

BigInt could be replaced by Long or BN. However, we should give the TypeScript implementation a try that was merged yesterday: https://github.com/Microsoft/TypeScript/pull/25886

There is no native BigDecimal interface yet

> BigInt could form the basis of an eventual BigDecimal implementation.

https://developers.google.com/web/updates/2018/05/bigint

webmaster128:
`amount` -> `quantity` as amount described the whole object.

```ts
interface BcpCoin {
    readonly quantity: string;
    readonly fractionalDigits: number;
}
```

As discussed, BcpCoin could be renamed to BcpAmount.

ethanfrey:
I agree with the format. Let's just look at the three affected types.

Currently `iov-bcp-types/src/bcp.ts`:

```ts
export interface BcpTicker {
  readonly tokenTicker: TokenTicker;
  readonly sigFigs: number;
  readonly tokenName: string;
}


export interface BcpCoin extends BcpTicker {
  readonly whole: number;
  readonly fractional: number;
}
```

And in `transactions.ts`:

```
export interface FungibleToken {
  readonly whole: number;
  readonly fractional: number;
  readonly tokenTicker: TokenTicker;
}
```

------

Changes:

`FungibleToken` -> `BcpAmount` as mentioned above:

```
export interface BcpAmount {
[...]

PR #526, opened 2018-11-13
Rename Amount to Quantity in @iov/dpos

webmaster128:
"Amount" should be "quantity" + "ticker"


webmaster128:
The [fresh BcpAmount](https://github.com/iov-one/iov-core/issues/502#issuecomment-438208755) will guide future code

PR #527, opened 2018-11-13
Improve tendermint starter

webmaster128:
This makes the start of the raw Tendermint on port 12345 independent of the IOV blockchain.

Simply use `./scripts/tendermint/start.sh` and `./scripts/tendermint/stop.sh` to start and stop as well as  `export TENDERMINT_ENABLED=1` in the console that runs the tests. No sourcing required anymore!

### DIFF
--- a/scripts/tendermint/start.sh
+++ b/scripts/tendermint/start.sh
@@ -2,7 +2,9 @@
 set -o errexit -o nounset -o pipefail
 command -v shellcheck > /dev/null && shellcheck "$0"
 
-TM_VERSION=0.21.0
+# Choose from https://hub.docker.com/r/tendermint/tendermint/tags/
+# TENDERMINT_VERSION is available for manualy testing different versions and usually unset
+VERSION=${TENDERMINT_VERSION:-0.21.0}
 PORT=12345
 
 TMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/tendermint.XXXXXXXXX")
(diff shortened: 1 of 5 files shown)


scripts/tendermint/start.sh
```diff
@@ -3,25 +3,28 @@ set -o errexit -o nounset -o pipefail
 command -v shellcheck > /dev/null && shellcheck "$0"
 
 TM_VERSION=0.21.0
-TM_DIR=$(mktemp -d "${TMPDIR:-/tmp}/tendermint.XXXXXXXXX")
-echo "TM_DIR = $TM_DIR"
+PORT=12345
```
ethanfrey: Why not allow override?
I was thinking we could run multiple copies of this script to test eg.
0.21 on 12345 and 0.25 on 11223

scripts/tendermint/stop.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+command -v shellcheck > /dev/null && shellcheck "$0"
+
+LABEL_PART="tendermint/tendermint"
+
+CONTAINER_ID=$(docker container ls | grep -F "$LABEL_PART:" | awk '{print $1}')
```
ethanfrey: sure. this is a reliable way to find the PID

webmaster128: The nice thing is, you can stop from a different console

scripts/tendermint/start.sh
webmaster128: As far as I can see the feature was not needed for now so I simplified the code.

The problem with multiple Tendermint containers is they are all called "tendermint/tendermint:<version>" and stopping the right one gets comlicated. Do we really need that feature? or can we just stop and start a different version?

ethanfrey: I want to start everything before running jasmine, as I think controlling this from js code is gonna get messy.

If we want to test both 0.21 and 0.25 adaptors work on the same commit, how do you propose we do so?

scripts/tendermint/stop.sh
ethanfrey: :+1:

scripts/tendermint/start.sh
webmaster128: 1. set TENDERMINT_ENABLED
2. start tendermint in default version (0.21)
3. run tests
4. stop tendermint
5. start tendermint in new version (0.25)
6. run tests (no JS change or rebuild required since Tendermint URL does not change)
7. stop tendermint

5. requires a code change but I can expose a config variable for the version instead

webmaster128: Point 5 works with `TENDERMINT_VERSION=0.25.0 ./scripts/tendermint/start.sh` now

PR #529, opened 2018-11-13
Register username tx

webmaster128:
Expected to fail with

```
1. BnsConnection : can register a username
   Error: {
     "checkTx": {
       "log": "Id is invalid",
       "code": 501
     },
     "deliverTx": {
       "code": 0
     }
   }
```

### DIFF
--- a/packages/iov-bcp-types/types/transactions.d.ts
+++ b/packages/iov-bcp-types/types/transactions.d.ts
@@ -20,7 +20,8 @@ export declare enum TransactionKind {
     SwapOffer = 2,
     SwapCounter = 3,
     SwapClaim = 4,
-    SwapTimeout = 5
+    SwapTimeout = 5,
+    RegisterUsername = 6
 }
 export interface BaseTx {
     readonly chainId: ChainId;
@@ -67,4 +68,9 @@ export interface SwapTimeoutTx extends BaseTx {
     readonly kind: TransactionKind.SwapTimeout;
     readonly swapId: SwapIdBytes;
 }
-export declare type UnsignedTransaction = SendTx | SetNameTx | SwapOfferTx | SwapCounterTx | SwapClaimTx | SwapTimeoutTx;
+export interface RegisterUsernameTx extends BaseTx {
+    readonly kind: TransactionKind.RegisterUsername;
+    readonly username: string;
+    readonly addresses: Map<ChainId, Address>;
+}
+export declare type UnsignedTransaction = SendTx | SetNameTx | SwapOfferTx | SwapCounterTx | SwapClaimTx | SwapTimeoutTx | RegisterUsernameTx;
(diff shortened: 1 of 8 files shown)


packages/iov-bcp-types/types/transactions.d.ts
```diff
@@ -61,4 +68,9 @@ export interface SwapTimeoutTx extends BaseTx {
     readonly kind: TransactionKind.SwapTimeout;
     readonly swapId: SwapIdBytes;
 }
-export declare type UnsignedTransaction = SendTx | SetNameTx | SwapOfferTx | SwapCounterTx | SwapClaimTx | SwapTimeoutTx;
+export interface RegisterUsernameTx extends BaseTx {
+    readonly kind: TransactionKind.RegisterUsername;
+    readonly username: string;
+    readonly addresses: Map<ChainId, Address>;
```
ethanfrey: I kept reading that tagged types didn't work as keys in a map.
So I was always using sting.
Nice if you can keep that safety here

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -321,6 +322,43 @@ describe("BnsConnection", () => {
...
+    const identity = await profile.createIdentity(wallet.id, HdPaths.simpleAddress(0));
+
+    // Create and send registration
+    const address = keyToAddress(identity.pubkey);
+    const username = `testuser`; // _${Math.random()}`;
+    const registration: RegisterUsernameTx = {
+      kind: TransactionKind.RegisterUsername,
+      chainId: chainId,
+      signer: identity.pubkey,
+      addresses: new Map([[chainId, address]]),
```
ethanfrey: This is a good question for me, what the address should be here, why bytes is human readable encoding, etc...

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -321,6 +322,43 @@ describe("BnsConnection", () => {
...
+      username: username,
+    };
+    const nonce = await getNonce(connection, address);
+    const signed = await profile.signTransaction(wallet.id, identity, registration, bnsCodec, nonce);
+    const txBytes = bnsCodec.bytesToPost(signed);
+    await connection.postTx(txBytes);
+
+    // Find registration transaction
+    const searchResult = await connection.searchTx({ tags: [bnsFromOrToTag(address)] });
```
ethanfrey: Do not expect this to be found, this is only if the balance changes at the given address.

`bnsNonceTag` should find it, as it detects anything that was signed by that address

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Good question indeed. What happens when one chain supports different address types? I tend to think we should store the user's choice.

For some broken chains out there, there is no proper binary address.

webmaster128:
okay, this is green locally. @ethanfrey may I get your approval?

PR #532, opened 2018-11-14
Move, cleanup and test encodeFullSignature()

webmaster128:


webmaster128:
> the motivation

The modules `types.ts` and `encode.ts` should not depend on each other in a circular fashion. Thus all encoding goes to `encode.ts`.

PR #533, opened 2018-11-14
Add RegisterBlockchainTx

webmaster128:
fails with

```
   Error: {
     "checkTx": {
       "log": "Invalid codec",
       "code": 510
     },
     "deliverTx": {
       "code": 0
     }
   }
```

PR #535, opened 2018-11-15
Add transactions to add and remove addresses to a username

### LINKED ISSUE #494: Expose BNS transactions
Depends on #493 

- [ ] Add transaction types for the bns registration for human value names and blockchain names, as well as updating the addresses associated with a value name.
- [ ] Implement these transactions in `@iov/bns`
- [ ] Consider defining subtypes for Signers based on which group of transactions they support (basic, atomic swap, name service, etc.)

webmaster128:
Based on #533

Closes #494

### DIFF
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -26,11 +26,8 @@ export interface Amount {
   readonly tokenTicker: TokenTicker;
 }
 
-/** the primary identifier for a registered blockchain in the BNS */
-export type BnsBlockchainId = Uint8Array & As<"blockchain-nft-id">;
-
 export interface ChainAddressPair {
-  readonly blockchainId: BnsBlockchainId;
+  readonly chainId: ChainId;
   readonly address: Address;
 }
 
@@ -111,9 +108,11 @@ export interface SwapTimeoutTx extends BaseTx {
 
 export interface RegisterBlockchainTx extends BaseTx {
   readonly kind: TransactionKind.RegisterBlockchain;
-  /** the ID of the blockchain to be registered */
-  readonly blockchainId: BnsBlockchainId;
-  /** as defined in https://github.com/iov-one/bns-spec/blob/master/docs/data/ObjectDefinitions.rst#chain */
+  /**
+   * The chain to be registered
+   *
+   * Fields as defined in https://github.com/iov-one/bns-spec/blob/master/docs/data/ObjectDefinitions.rst#chain
+   */
   readonly chain: {
     readonly chainId: ChainId;
     readonly name: string;
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -34,10 +34,12 @@ export enum TransactionKind {
   SwapCounter,
   SwapClaim,
   SwapTimeout,
+  RegisterBlockchain,
   RegisterUsername,
 }
 
 export interface BaseTx {
+  /** the chain on which the transaction should be valid */
   readonly chainId: ChainId;
   readonly fee?: Amount;
   // signer needs to be a PublicKey as we use that to as an identifier to the Keyring for lookup
@@ -90,6 +92,12 @@ export interface SwapTimeoutTx extends BaseTx {
   readonly swapId: SwapIdBytes; // pulled from the offer transaction
 }
 
+export interface RegisterBlockchainTx extends BaseTx {
+  readonly kind: TransactionKind.RegisterBlockchain;
+  /** the ID of the blockchain to be registered */
+  readonly blockchainId: ChainId;
+}
+
 export interface RegisterUsernameTx extends BaseTx {
   readonly kind: TransactionKind.RegisterUsername;
   readonly username: string;
@@ -103,4 +111,5 @@ export type UnsignedTransaction =
   | SwapCounterTx
   | SwapClaimTx
   | SwapTimeoutTx
+  | RegisterBlockchainTx
   | RegisterUsernameTx;
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -116,11 +116,12 @@ export interface RegisterBlockchainTx extends BaseTx {
   /** as defined in https://github.com/iov-one/bns-spec/blob/master/docs/data/ObjectDefinitions.rst#chain */
   readonly chain: {
     readonly chainId: ChainId;
-    readonly networkId: string | undefined;
     readonly name: string;
     readonly enabled: boolean;
     readonly production: boolean;
-    readonly mainTickerId: TokenTicker | undefined;
+
+    readonly networkId?: string;
+    readonly mainTickerId?: TokenTicker;
   };
   readonly codecName: string;
   readonly codecConfig: string;
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -96,6 +96,8 @@ export interface RegisterBlockchainTx extends BaseTx {
   readonly kind: TransactionKind.RegisterBlockchain;
   /** the ID of the blockchain to be registered */
   readonly blockchainId: ChainId;
+  readonly codecName?: string;
+  readonly codecConfig?: string;
 }
 
 export interface RegisterUsernameTx extends BaseTx {
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -26,7 +26,13 @@ export interface Amount {
   readonly tokenTicker: TokenTicker;
 }
 
+export interface ChainAddressPair {
+  readonly chainId: ChainId;
+  readonly address: Address;
+}
+
 export enum TransactionKind {
+  AddAddressToUsername,
   Send,
   /** @deprecated see SetNameTx */
   SetName,
@@ -36,6 +42,7 @@ export enum TransactionKind {
   SwapTimeout,
   RegisterBlockchain,
   RegisterUsername,
+  RemoveAddressFromUsername,
 }
 
 export interface BaseTx {
@@ -45,6 +52,13 @@ export interface BaseTx {
   // signer needs to be a PublicKey as we use that to as an identifier to the Keyring for lookup
   readonly signer: PublicKeyBundle;
   readonly ttl?: TtlBytes;
+}
+
+export interface AddAddressToUsernameTx extends BaseTx {
+  readonly kind: TransactionKind.AddAddressToUsername;
+  /** the username to be updated, must exist on chain */
+  readonly username: string;
+  readonly payload: ChainAddressPair;
 }
 
 export interface SendTx extends BaseTx {
@@ -106,7 +120,15 @@ export interface RegisterUsernameTx extends BaseTx {
   readonly addresses: Map<ChainId, Address>;
 }
 
+export interface RemoveAddressFromUsernameTx extends BaseTx {
+  readonly kind: TransactionKind.RemoveAddressFromUsername;
+  /** the username to be updated, must exist on chain */
+  readonly username: string;
+  readonly payload: ChainAddressPair;
+}
+
 export type UnsignedTransaction =
+  | AddAddressToUsernameTx
   | SendTx
   | SetNameTx
   | SwapOfferTx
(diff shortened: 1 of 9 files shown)


packages/iov-bcp-types/src/transactions.ts
```diff
@@ -90,17 +106,35 @@ export interface SwapTimeoutTx extends BaseTx {
...
+  /** the ID of the blockchain to be registered */
+  readonly blockchainId: ChainId;
+  readonly codecName: string;
+  readonly codecConfig: string;
+}
+
 export interface RegisterUsernameTx extends BaseTx {
   readonly kind: TransactionKind.RegisterUsername;
   readonly username: string;
   readonly addresses: Map<ChainId, Address>;
```
ethanfrey: Now that you have introduced `ChainAddressPair `, do we still want a `Map` here or `ReadonlyArray<ChainAddressPair>`?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -363,6 +404,109 @@ describe("BnsConnection", () => {
     connection.disconnect();
   });
 
+  it("can add address to username and remove again", async () => {
```
ethanfrey: Can you test removing before adding, to show the failure when it is not registered (or just removing some random non-matching value)?

This test can be enhanced much more when we support queries to view current state, until then, mainly the check of what does and doesn't work.

packages/iov-bns/src/decode.spec.ts
```diff
@@ -66,6 +68,46 @@ describe("Decode", () => {
...
+
+    it("works for RegisterBlockchain", () => {
+      const transactionMessage: codecImpl.app.ITx = {
+        issueBlockchainNftMsg: {
+          id: Encoding.toAscii("wonderland"),
+          owner: Encoding.fromHex("0011223344556677889900112233445566778899"),
+          approvals: undefined,
+          details: {
+            iov: {
+              codec: "wonderland_rules",
```
ethanfrey: these are both arbitrary strings for the backend, just used by client, right?

packages/iov-bns/src/encode.ts
```diff
@@ -158,6 +177,23 @@ function buildSwapTimeoutTx(tx: SwapTimeoutTx): codecImpl.app.ITx {
...
 }
 
+function buildRegisterBlockchainTx(tx: RegisterBlockchainTx): codecImpl.app.ITx {
+  return {
+    issueBlockchainNftMsg: codecImpl.blockchain.IssueTokenMsg.create({
+      id: Encoding.toUtf8(tx.blockchainId),
+      owner: decodeBnsAddress(keyToAddress(tx.signer)).data,
+      approvals: undefined,
+      details: codecImpl.blockchain.TokenDetails.create({
+        chain: codecImpl.blockchain.Chain.create({}),
```
ethanfrey: Why is this empty?

packages/iov-bcp-types/src/transactions.ts
webmaster128: I prefer the later one as it better reflects what we get from the server. Working with Maps is not always as convenient as it could be in JS land and the server will take care we have no duplicate keys hopefully. Map only exist for history reasons.

packages/iov-bns/src/decode.spec.ts
webmaster128: right. To be honest I don't understand the semantics of it

packages/iov-bns/src/encode.ts
webmaster128: Because we can ;) This is the smallest possible implementation accepted by the server

packages/iov-bns/src/decode.spec.ts
ethanfrey: This should be documented, but I did request some info, so we can read this info, and then initialize a BnsConnection or LiskConnection or whatever.

The codec and options are only meant to be understood by iov-core... And if there is garbage there, we don't connect.

packages/iov-bns/src/encode.ts
webmaster128: After [splitting BnsBlockchainId and ChainId](https://github.com/iov-one/iov-core/pull/535/commits/f1aa56291987bd3ce62c09df146ba56f99a60a42) I now have a better idea what goes into `chain`. Will add some initial `blockchain.Chain` data now.

webmaster128: Added in https://github.com/iov-one/iov-core/pull/535/commits/f265f1d2f363dab10c43f01dee377429d75cff5f

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: ok: https://github.com/iov-one/iov-core/pull/535/commits/f18008d6521d1c052f91659dcb76991a637df5e6

webmaster128:
@ethanfrey Could you have another look at the new commits starting from "Convert Map<ChainId, Address> to ReadonlyArray<ChainAddressPair> 1579974 "

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -26,8 +26,11 @@ export interface Amount {
   readonly tokenTicker: TokenTicker;
 }
 
+/** the primary identifier for a registered blockchain in the BNS */
+export type BnsBlockchainId = Uint8Array & As<"blockchain-nft-id">;
```
ethanfrey: So, you want to separate this from ChainId....
I think we will want them to be the same

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -109,15 +112,24 @@ export interface SwapTimeoutTx extends BaseTx {
 export interface RegisterBlockchainTx extends BaseTx {
   readonly kind: TransactionKind.RegisterBlockchain;
   /** the ID of the blockchain to be registered */
-  readonly blockchainId: ChainId;
+  readonly blockchainId: BnsBlockchainId;
+  /** as defined in https://github.com/iov-one/bns-spec/blob/master/docs/data/ObjectDefinitions.rst#chain */
+  readonly chain: {
```
ethanfrey: We don't have to follow the encoded format here, more what makes most sense on the public side. `codecName` and `codecConfig` are stored under `iov` in the protobuf object.

Maybe `chainId` could go to the top level.
I'm just wondering how much we need to separate them how much we can hide complexity client side.

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -109,15 +112,24 @@ export interface SwapTimeoutTx extends BaseTx {
...
+  };
   readonly codecName: string;
   readonly codecConfig: string;
 }
 
 export interface RegisterUsernameTx extends BaseTx {
   readonly kind: TransactionKind.RegisterUsername;
   readonly username: string;
-  readonly addresses: Map<ChainId, Address>;
+  readonly addresses: ReadonlyArray<ChainAddressPair>;
```
ethanfrey: Yes, I think this is an easier structure also

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -328,20 +331,29 @@ describe("BnsConnection", () => {
...
     const identityAddress = keyToAddress(identity.pubkey);
 
     // Create and send registration
-    const blockchainId = `wonderland_${Math.random()}` as ChainId;
+    const chainId = `wonderland_${Math.random()}` as ChainId;
+    const blockchainId = toUtf8(chainId) as BnsBlockchainId;
     const registration: RegisterBlockchainTx = {
       kind: TransactionKind.RegisterBlockchain,
-      chainId: chainId,
+      chainId: registraChainId,
```
ethanfrey: Correct, we register on one chain, but the chain being registered is a different id

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -328,20 +331,29 @@ describe("BnsConnection", () => {
   it("can register a blockchain", async () => {
     pendingWithoutBnsd();
     const connection = await BnsConnection.establish(bnsdTendermintUrl);
-    const chainId = await connection.chainId();
+    const registraChainId = await connection.chainId();
```
ethanfrey: s/registra/registrar/

packages/iov-bns/src/decode.spec.ts
```diff
@@ -133,11 +149,15 @@ describe("Decode", () => {
...
       }
       expect(parsed.username).toEqual("bobby");
-      expect(parsed.addresses.size).toEqual(2);
-      expect(parsed.addresses.get("chain1" as ChainId)).toEqual("23456782367823X" as Address);
-      expect(parsed.addresses.get("chain2" as ChainId)).toEqual(
-        "0x001100aabbccddffeeddaa8899776655" as Address,
-      );
+      expect(parsed.addresses.length).toEqual(2);
+      expect(parsed.addresses[0]).toEqual({
+        blockchainId: toUtf8("chain1") as BnsBlockchainId,
```
ethanfrey: I would almost suggest a function for this...

`blockchainId: chainToBlockchainId("chain1")`

packages/iov-bns/src/decode.ts
```diff
@@ -156,38 +158,51 @@ function parseRegisterBlockchainTx(
 ): RegisterBlockchainTx {
   const id = ensure(msg.id, "id");
   const details = ensure(msg.details, "details");
+
+  const chain = ensure(details.chain, "details.chain");
```
ethanfrey: Are all of these required and guaranteed to be present by the backend?

Good use of ensure.

packages/iov-bns/src/encode.spec.ts
```diff
@@ -105,12 +107,26 @@ describe("Encode", () => {
...
+          mainTickerId: "WONDER" as TokenTicker,
+          name: "Wonderland",
+        },
         codecName: "rules_of_wonderland",
         codecConfig: `{ rules: ["make peace not war"] }`,
       };
       const msg = buildMsg(registerBlockchain).issueBlockchainNftMsg!;
       expect(msg.id).toEqual(toAscii("wonderland"));
+      expect(msg.details!.chain!.chainID).toEqual("wonderland");
+      expect(msg.details!.chain!.networkID).toEqual("7rg047g4h");
```
ethanfrey: In the decode you ensure `chain` is present.
Here it looks like we are unsure of it's existence in the types.

I don't know what is correct, but should be consistent.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Good. Now I only need to fix the typo in `registraChainId`

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: I think "register" is the work I am looking for

packages/iov-bns/src/encode.spec.ts
webmaster128: >  Here it looks like we are unsure of it's existence in the types.

`msg` has the protobuf interface type. `details` and `chain` is coming as optional from the protobuf types in TypeScript. Since they are marked as non-null in
* https://github.com/iov-one/weave/blob/master/x/nft/blockchain/codec.proto#L41
* https://github.com/iov-one/weave/blob/master/x/nft/blockchain/codec.proto#L15
I assume both fields need to exist. Thus I force-cast them to non-null, which means I do the same as in decode.

packages/iov-bns/src/decode.ts
webmaster128: Can we trust the `gogoproto.nullable` annotations in https://github.com/iov-one/weave/blob/master/x/nft/blockchain/codec.proto#L41 / https://github.com/iov-one/weave/blob/master/x/nft/blockchain/codec.proto#L15?

packages/iov-bcp-types/src/transactions.ts
webmaster128: We use `ChainId` to implement `BnsBlockchainId`. But having the types as separate concepts, the implementation of `BnsBlockchainId` can change if we want

packages/iov-bcp-types/src/transactions.ts
webmaster128: Right. The problem with moving `chainId` up to top level is the conflict with BaseTx.chainId. The only potential I see here is removing `readonly blockchainId: BnsBlockchainId;` as we can set this using `chain.chainId`

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: No, registry was the fancy english word for it

PR #536, opened 2018-11-16
Sign eth tx

### LINKED ISSUE #405: Sign ethereum transaction and move ETH
This builds on #403 and #404 and should demonstrate signing and submitting a valid ethereum transaction.

Tests should demonstrate moving tokens from one of the auto-generated genesis accounts to a new account controlled by a `Secp256HdWallet`, as well as from one account on that `HdWallet` to another.

LucasIsasmendi:
Closes #405 
**WIP**
Pending:
- [x] Fix signature, `r` and `s` are not right
- [x] Improve logic to get `v` value
- [x] Add memo logic and test

We need to wait #546 to run postTx test in CI. Tests were done against ropsten


packages/iov-ethereum/src/serialization.ts
```diff
@@ -0,0 +1,73 @@
...
+            Buffer.from([]),
+            Buffer.from([]),
+          ]),
+        );
+      default:
+        throw new Error("Unsupported kind of transaction");
+    }
+  }
+  public static amountFromComponents(whole: number, fractional: number): string {
+    const base10BN = new BN(10);
```
webmaster128: Please use strict CamelCase notation, even if external libraries don't: `base10Bn` etc. Nobody likes code SCREAMING_AT the developer

packages/iov-ethereum/src/utils.spec.ts
```diff
@@ -77,6 +86,68 @@ describe("Ethereum utils", () => {
...
+      expect(encQtyHex).toEqual("0x41");
+      encQtyHex = encodeQuantity(2);
+      expect(encQtyHex).toEqual("0x2");
+      encQtyHex = encodeQuantity(0);
+      expect(encQtyHex).toEqual("0x0");
+    });
+
+    it("throws error for invalid inputs", () => {
+      expect(() => encodeQuantity(NaN)).toThrowError(/Input is not a number/);
+      expect(() => encodeQuantity(12345678901234567890)).toThrowError(/Input is not a safe integer/);
```
webmaster128: * What about negative numbers?
* What about fractional numbers?

packages/iov-ethereum/src/utils.ts
```diff
@@ -4,7 +4,8 @@ import { Int53 } from "@iov/encoding";
 import BN = require("bn.js");
 
 export function decodeHexQuantity(hexString: string): number {
-  if (hexString.match(/^0x[a-f0-9]+$/)) {
+  // "0x00" in ganache response is an error: zero should be represented as "0x0"
+  if (hexString === "0x0" || hexString === "0x00" || hexString.match(/^0x{1}[a-f1-9][a-f0-9]*$/)) {
```
webmaster128: The `{1}` in the regexp is unnecessary.

packages/iov-ethereum/src/utils.ts
```diff
@@ -22,3 +23,37 @@ export function decodeHexQuantityNonce(hexString: string): Nonce {
...
+  return "0x" + new BN(value).toString(16);
+}
+
+export function stringDataToHex(stringData: string): string {
+  let hex = "";
+  for (let i = 0; i < stringData.length; i++) {
+    const n = stringData.charCodeAt(i).toString(16);
+    hex += n.length < 2 ? "0" + n : n;
+  }
+  return "0x" + hex;
```
webmaster128: This is a reimplementation of `Encoding.toHex(Encoding.toUtf8(stringData))`. Please use those well-tested methods

packages/iov-ethereum/src/utils.spec.ts
```diff
@@ -77,6 +86,68 @@ describe("Ethereum utils", () => {
...
+
+  describe("stringDataToHex", () => {
+    it("verify valid inputs", () => {
+      let dataToHex;
+      dataToHex = stringDataToHex("A");
+      expect(dataToHex).toEqual("0x41");
+      dataToHex = stringDataToHex("\0B\0");
+      expect(dataToHex).toEqual("0x004200");
+      dataToHex = stringDataToHex("");
+      expect(dataToHex).toEqual("0x");
```
webmaster128: Since this is used for memo, please also add characters that are more than one byte, e.g.

> ⇉ (Rightwards Paired Arrows, U+21c9) takes 2 bytes in UTF-8

packages/iov-keycontrol/src/wallets/slip10wallet.ts
```diff
@@ -283,7 +283,10 @@ export class Slip10Wallet implements Wallet {
         signature = await Ed25519.createSignature(message, await Ed25519.makeKeypair(privkey));
         break;
       case Slip10Curve.Secp256k1:
-        signature = await Secp256k1.createSignature(message, privkey);
+        signature =
+          prehashType === PrehashType.Keccak256
+            ? await Secp256k1.createSignatureEth(message, privkey)
+            : await Secp256k1.createSignature(message, privkey);
```
webmaster128: Hmm, this means we hash twice for `prehashType === Sha256` which is broken in the current code

packages/iov-crypto/src/secp256k1.ts
```diff
@@ -57,6 +57,13 @@ export class Secp256k1 {
     return signature;
   }
 
+  public static async createSignatureEth(message: Uint8Array, privkey: Uint8Array): Promise<Uint8Array> {
```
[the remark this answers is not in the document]
LucasIsasmendi: We still need this function because ethereum uses recovery parameter to encode chain id

packages/iov-ethereum/src/utils.ts
```diff
@@ -49,3 +49,11 @@ export function hexPadToEven(hex: string): string {
   }
   return hex.replace("0x", "");
 }
+
+export function trimLeadingZero(hex: string): string {
```
webmaster128: Why does this function accept odd number of characters (e.g. "002") but does not trim odd number of leading zeros?

* if this is for bytes encoded as hex -> input length must be even
* if this is for integers encoded as hex -> all leading zeros should be trimmed

packages/iov-ethereum/src/utils.ts
```diff
@@ -49,3 +49,11 @@ export function hexPadToEven(hex: string): string {
...
+  let hexInput = hex;
+  while (hexInput && hexInput.startsWith("00")) {
+    hexInput = "" + hexInput.slice(2);
```
webmaster128: why ""? This is the same as

```ts
hexInput = hexInput.slice(2);
```

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -82,7 +82,7 @@ export const ethereumCodec: TxCodec = {
         const r = sig.slice(4, 36);
         const s = sig.slice(38, 70);
         // TODO: find the right way to do this
-        let v = Number(sig.slice(70, 71)) + 27;
+        let v = Number(sig.slice(-1)) + 27;
```
webmaster128: The slice API is to beautiful :) Nearly as nice as Python

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -97,8 +97,8 @@ export const ethereumCodec: TxCodec = {
             Buffer.from(fromHex(hexPadToEven(valueHex))),
             Buffer.from(fromHex(hexPadToEven(dataHex))),
             Buffer.from(fromHex(hexPadToEven(chainIdHex))),
-            Buffer.from(r),
-            Buffer.from(s),
+            Buffer.from(fromHex(trimLeadingZero(toHex(r)))),
+            Buffer.from(fromHex(trimLeadingZero(toHex(s)))),
```
webmaster128: You can implement trimLeadingZero on bytes directly (untested):

```ts
function trimLeadingNullBytes(inData: Uint8Array): Uint8Array {
  let numberOfLeadingNullBytes = 0;
  for (const byte of inData) {
    if (byte === 0x00) {
      numberOfLeadingNullBytes++;
    } else {
      break;
    }
    return inData.slice(numberOfLeadingNullBytes);
  }
}
```

packages/iov-keycontrol/src/wallets/slip10wallet.ts
```diff
@@ -307,7 +307,10 @@ export class Slip10Wallet implements Wallet {
         //   const r = sig.r();
         //   const s = sig.s();
         //   const recoveryParam = sig.recovery;
-        signature = (await Secp256k1.createSignature(message, privkey)).toFixedLength();
+        signature =
+          prehashType === PrehashType.Keccak256
+            ? (await Secp256k1.createSignature(message, privkey)).toFixedLength()
+            : (await Secp256k1.createSignature(message, privkey)).toDer();
```
webmaster128: Please revert that change. Everyone gets fix length encoding. There is no good reason to default to DER. When you need DER, you can convert fixed length -> DER as described in the comment.

packages/iov-keycontrol/src/wallets/secp256k1hdwallet.spec.ts
```diff
@@ -123,7 +123,7 @@ describe("Secp256k1HdWallet", () => {
     );
 
     const valid = await Secp256k1.verifySignature(
-      ExtendedSecp256k1Signature.fromFixedLength(signatureBytes),
+      ExtendedSecp256k1Signature.fromDer(signatureBytes),
```
webmaster128: Also revert this one. Fixed length encoding is the default now, as it contains all the data.

packages/iov-ethereum/src/utils.ts
```diff
@@ -22,3 +24,36 @@ export function decodeHexQuantityNonce(hexString: string): Nonce {
...
+}
+
+export function hexPadToEven(hex: string): string {
+  if (hex.length % 2 !== 0) {
+    return "0" + hex.replace("0x", "");
+  }
+  return hex.replace("0x", "");
+}
+
+export function trimLeadingZero(hex: string): string {
```
webmaster128: This function becomes obsolete

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -62,4 +75,70 @@ describe("EthereumConnection", () => {
...
+
+    const wallet = Secp256k1HdWallet.fromMnemonic(
+      "oxygen fall sure lava energy veteran enroll frown question detail include maximum",
+    );
+    const mainIdentity = await wallet.createIdentity([
+      Slip10RawIndex.hardened(44),
+      Slip10RawIndex.hardened(60),
+      Slip10RawIndex.hardened(0),
+      Slip10RawIndex.normal(0),
+      Slip10RawIndex.normal(1),
```
webmaster128: `await wallet.createIdentity(HdPaths.bip44(60, 0, 0, 1))`

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -26,10 +34,76 @@ export function toChecksumAddress(address: string): Address {
...
+          dataHex += Encoding.toHex(Encoding.toUtf8(signed.transaction.memo));
+        }
+        if (!isValidAddress(signed.transaction.recipient)) {
+          throw new Error("Invalid recipient address");
+        }
+        const sig = signed.primarySignature.signature;
+        const r = sig.slice(0, 32);
+        const s = sig.slice(32, 64);
+        let v = Number(sig.slice(-1)) + 27;
+        const chainId = Number(signed.transaction.chainId);
```
webmaster128: Please use the tested code to parse the fixed length encoding:

```ts
const sig = ExtendedSecp256k1Signature.fromFixedLength(signed.primarySignature.signature);
const r = sig.r();
const s = sig.s();
const recoveryParam = sig.recovery;
```

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -26,10 +34,76 @@ export function toChecksumAddress(address: string): Address {
...
+          toRlp([
+            Buffer.from(fromHex(hexPadToEven(nonceHex))),
+            Buffer.from(fromHex(hexPadToEven(gasPriceHex))),
+            Buffer.from(fromHex(hexPadToEven(gasLimitHex))),
+            Buffer.from(fromHex(hexPadToEven(signed.transaction.recipient))),
+            Buffer.from(fromHex(hexPadToEven(valueHex))),
+            Buffer.from(fromHex(hexPadToEven(dataHex))),
+            Buffer.from(fromHex(hexPadToEven(chainIdHex))),
+            Buffer.from(fromHex(trimLeadingZero(toHex(r)))),
+            Buffer.from(fromHex(trimLeadingZero(toHex(s)))),
```
webmaster128: `r` and `s` from ExtendedSecp256k1Signature have no leading 0s. So this becomes just

```
Buffer.from(r),
Buffer.from(s),
```

and `trimLeadingZero` can be removed

packages/iov-ethereum/src/derivation.ts
```diff
@@ -0,0 +1,23 @@
...
+      if (
+        (parseInt(addressHash[i], 16) > 7 && addressLower[i].toUpperCase() !== address[i + 2]) ||
+        (parseInt(addressHash[i], 16) <= 7 && addressLower[i] !== address[i + 2])
+      ) {
+        return false;
+      }
+    }
+    return true;
+  }
+  throw new Error("Invalid ethereum address");
```
webmaster128: Just return `false` in this case, no exception. `isValidAddress` is not for checksum matching but also returns false when you enter a Lisk address in an Ethereum transaction.

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -26,10 +34,76 @@ export function toChecksumAddress(address: string): Address {
...
+          toRlp([
+            Buffer.from(fromHex(hexPadToEven(nonceHex))),
+            Buffer.from(fromHex(hexPadToEven(gasPriceHex))),
+            Buffer.from(fromHex(hexPadToEven(gasLimitHex))),
+            Buffer.from(fromHex(hexPadToEven(signed.transaction.recipient))),
+            Buffer.from(fromHex(hexPadToEven(valueHex))),
+            Buffer.from(fromHex(hexPadToEven(dataHex))),
+            Buffer.from(fromHex(hexPadToEven(chainIdHex))),
+            Buffer.from(r),
+            Buffer.from(s),
```
webmaster128: ❤️

PR #539, opened 2018-11-16
Add support for tendermint 0.25

### LINKED ISSUE #295: Upgrade tendermint client to support 0.24 (and 0.23?)
We only support 0.20 and 0.21 for now.

0.22 seems like it was a bit of a refactoring series, and no need to add support to it now.

0.24 has quite a few fixes and we should plan to upgrade the backend in the near future, so we will need client support. If we can also get 0.23 without much extra work, why not?

ethanfrey:
Closes #295 

- [x] Update request formatting (integers as strings)
- [ ] Update response formatting (quite a bit)
- [ ] Test compatibility with basic tendermint rpc kvstore
- [ ] Test compatibility in bns

### DIFF
--- a/packages/iov-tendermint-rpc/src/encodings.ts
+++ b/packages/iov-tendermint-rpc/src/encodings.ts
@@ -48,6 +48,10 @@ export function parseInteger(str: IntegerString): number {
   return Int53.fromString(str).toNumber();
 }
 
+export function encodeInteger(num: number): IntegerString {
+  return new Int53(num).toString();
+}
+
 export class Base64 {
   public static encode(data: Uint8Array): Base64String {
     return Encoding.toBase64(data) as Base64String;
--- a/packages/iov-tendermint-rpc/src/0-25/requests.ts
+++ b/packages/iov-tendermint-rpc/src/0-25/requests.ts
@@ -1,0 +1,129 @@
+import { Encoding } from "@iov/encoding";
+
+import { JsonRpcRequest, jsonRpcWith } from "../common";
+import { Base64, Base64String, encodeInteger, HexString, may, notEmpty } from "../encodings";
+import * as requests from "../requests";
+
+/***** queries *****/
+
+export class Params extends requests.DefaultParams {
+  public static encodeAbciQuery(req: requests.AbciQueryRequest): JsonRpcRequest {
+    return jsonRpcWith(req.method, encodeAbciQueryParams(req.params));
+  }
+
+  public static encodeBlock(req: requests.BlockRequest): JsonRpcRequest {
+    return jsonRpcWith(req.method, encodeHeightParam(req.params));
+  }
+
+  public static encodeBlockchain(req: requests.BlockchainRequest): JsonRpcRequest {
+    return jsonRpcWith(req.method, encodeBlockchainRequestParams(req.params));
+  }
+
+  public static encodeBlockResults(req: requests.BlockResultsRequest): JsonRpcRequest {
+    return jsonRpcWith(req.method, encodeHeightParam(req.params));
+  }
+
+  public static encodeBroadcastTx(req: requests.BroadcastTxRequest): JsonRpcRequest {
+    return jsonRpcWith(req.method, encodeBroadcastTxParams(req.params));
+  }
+
+  public static encodeCommit(req: requests.CommitRequest): JsonRpcRequest {
+    return jsonRpcWith(req.method, encodeHeightParam(req.params));
+  }
+
+  public static encodeSubscribe(req: requests.SubscribeRequest): JsonRpcRequest {
+    const allTags: ReadonlyArray<requests.QueryTag> = [
+      { key: "tm.event", value: req.query.type },
+      ...(req.query.tags ? req.query.tags : []),
+    ];
+
+    const queryString = requests.buildTagsQuery(allTags);
+    return jsonRpcWith("subscribe", { query: queryString });
+  }
+
+  public static encodeTx(req: requests.TxRequest): JsonRpcRequest {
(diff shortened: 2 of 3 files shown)


ethanfrey:
This is ready for review.
Tendermint 0.25 tested in the CI.

I will make a separate PR for weave 0.9 support, maybe yet another for testing multiple tendermint versions at once (eg. 0.21 and 0.25)

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -208,197 +399,13 @@ describe("Client", () => {
     expect(info3).toBeTruthy();
   });
 
-  describe("With HttpClient", () => {
-    kvTestSuite(() => new HttpClient(tendermintUrl));
+  describe("With HttpClient: v0-20", () => {
```
webmaster128: Remove (wrong) version from description.

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -208,197 +399,13 @@ describe("Client", () => {
...
-
-        const transaction1 = buildKvTx(randomId(), randomId());
-        const transaction2 = buildKvTx(randomId(), randomId());
-
-        await client.broadcastTxCommit({ tx: transaction1 });
-        await client.broadcastTxCommit({ tx: transaction2 });
-      })().catch(fail);
-    });
+    const factory = () => new WebsocketClient(tendermintUrl, onError);
+    websocketTestSuite(factory, v0_25);
```
webmaster128: This adds some strange nesting that is hard to follow. We can keep it flat as before:

```ts
const factory = () => new WebsocketClient(tendermintUrl, onError);
kvTestSuite(factory, v0_25);
websocketTestSuite(factory, v0_25);
```

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -188,6 +188,197 @@ function kvTestSuite(rpcFactory: () => RpcClient): void {
   });
 }
 
+function websocketTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor): void {
+  // don't print out WebSocket errors if marked pending
```
webmaster128: Comment belongs to `const onError = skipTests() ? () => 0 : console.log;`

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -188,6 +188,197 @@ function kvTestSuite(rpcFactory: () => RpcClient): void {
...
+  kvTestSuite(rpcFactory, adaptor);
+
+  it("can subscribe to block header events", done => {
+    pendingWithoutTendermint();
+
+    // seems that tendermint just guarantees within the last second for timestamp
+    const testStart = ReadonlyDate.now() - 1000;
```
webmaster128: Could you try `expect(event.time.getTime()).toBeGreaterThanOrEqual(testStart);` instead of this unintuitive substraction?

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -188,6 +188,197 @@ function kvTestSuite(rpcFactory: () => RpcClient): void {
...
+        complete: () => fail("Stream must not close just because we don't listen anymore"),
+      });
+    })().catch(fail);
+  });
+
+  it("can subscribe to block events", done => {
+    pendingWithoutTendermint();
+
+    // seems that tendermint just guarantees within the last second for timestamp
+    const testStart = ReadonlyDate.now() - 1000;
```
webmaster128: same here

packages/iov-tendermint-rpc/src/client.spec.ts
webmaster128: I see. I am fine with leaving the logic as it is. But moving the `- 1000` from `testStart` into the matcher makes the situation much clearer

ethanfrey: Nice

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -44,8 +44,21 @@ export function may<T, U>(transform: (val: T) => U, value: T | null | undefined)
...
+export class Integer {
+  public static parse(str: IntegerString): number {
+    return Int53.fromString(str).toNumber();
+  }
+
+  public static encode(num: number): IntegerString {
+    return new Int53(num).toString() as IntegerString;
+  }
+
+  public static ensure(n: number): number {
```
webmaster128: If you don't trust the runtime type to be the type system type, this shout change to `n: unknown`.

Also since this is called Integer.ensure, we need at least a [Number.isInteger](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) test

ethanfrey: Yeah, the docs say it is a string, but in tests it was an integer.
I wanted to make sure it was there

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -53,11 +53,14 @@ export class Integer {
...
 
-  public static ensure(n: number): number {
-    if (typeof n !== "number") {
-      throw new Error(`${n} is not a number`);
+  public static ensure(num: unknown): number {
+    if (typeof num !== "number" || !Number.isInteger(num)) {
+      throw new Error(`${num} is not a number`);
     }
-    return n;
+    if (!Number.isInteger(num)) {
```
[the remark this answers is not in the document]
ethanfrey: Not any more. Amended that commit.

PR #545, opened 2018-11-20
Fix ganache-cli version

webmaster128:
We cannot upgrade to 6.2 because of https://github.com/trufflesuite/ganache-cli/issues/595


davidmurdoch:
Fixed in [ganache-cli@6.2.2](https://github.com/trufflesuite/ganache-cli/releases/tag/v6.2.2). Let me know if you have any more issues.

PR #547, opened 2018-11-20
Upgrade dependencies (next try)

webmaster128:


webmaster128:
Tested this with Ledger as well

webmaster128:
With this upgrade, we get nodejs 11 support

Issue #495, opened 2018-11-06
Add BNS query/watch support


ethanfrey:
Depends on #493 

We should be able to query and watch the various bns data types.

- [ ] value name by name
- [ ] value name by (chain, address) - reverse lookup
- [ ] blockchain by chainId

After this and #494, a client should be able to register and query bns value names against a running bnsd 0.8.0/0.9.0 instance

webmaster128:
> value name by name

Is "name" the deprecated name from `namecoin.SetWalletNameMsg`? Then I think we should drop that and query _value name by address_ instead.

ethanfrey:
> Then I think we should drop that and query value name by BNS address instead.

We should query by username (first point)

Maybe there is a better word for it. In any case the id of the nft token

webmaster128:
Thanks for clarification, makes sense now. We call it `username` in `RegisterUsernameTx` (which is implemented as the ID of the NFT)

PR #549, opened 2018-11-21
[iov-bns] Add query username support

webmaster128:
This is supposed to cover points 1 and 2 from https://github.com/iov-one/iov-core/issues/495

### DIFF
--- a/packages/iov-bns/src/bnsconnection.ts
+++ b/packages/iov-bns/src/bnsconnection.ts
@@ -55,6 +55,8 @@ import {
   keyToAddress,
 } from "./util";
 
+const { toAscii, toHex } = Encoding;
+
 /**
  * Returns a filter that only passes when the
  * value is different than the last one
@@ -89,7 +91,7 @@ export class BnsConnection implements BcpAtomicSwapConnection {
 
     // inlining getAllTickers
     const res = await performQuery(tmClient, "/tokens?prefix", Uint8Array.from([]));
-    const parser = createParser(codecImpl.namecoin.Token, 4);
+    const parser = createParser(codecImpl.namecoin.Token, "tkn:");
     const data = res.results.map(parser).map(Normalize.token);
 
     const toKeyValue = (t: BcpTicker): [string, BcpTicker] => [t.tokenTicker, t];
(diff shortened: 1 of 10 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -445,15 +477,23 @@ export interface QueryResponse {
   readonly results: ReadonlyArray<Result>;
 }
 
-function parseMap<T extends {}>(decoder: Decoder<T>, sliceKey: number): (res: Result) => T & Keyed {
-  const mapper = (res: Result): T & Keyed => {
+function createParser<T extends {}>(decoder: Decoder<T>, keyPrefix: string): (res: Result) => T & Keyed {
```
ethanfrey: Yes, this is a better, cleaner implementation than the other

packages/iov-bns/src/decode.spec.ts
```diff
@@ -18,9 +18,34 @@ import {
 } from "./testdata";
 import { decodePrivkey, decodePubkey } from "./types";
 
-const { toUtf8 } = Encoding;
+const { fromHex, toUtf8 } = Encoding;
 
 describe("Decode", () => {
+  it("decodes username NFT", () => {
```
ethanfrey: :+1:

packages/iov-bns/src/decode.ts
```diff
@@ -23,11 +23,22 @@ import {
...
 import * as codecImpl from "./generated/codecimpl";
-import { asNumber, decodeFullSig, ensure } from "./types";
+import { asNumber, BnsAddressBytes, BnsUsernameNft, decodeFullSig, ensure } from "./types";
 import { encodeBnsAddress, isHashIdentifier } from "./util";
 
 const { fromUtf8 } = Encoding;
 
+export function decodeUsernameNft(nft: codecImpl.username.IUsernameToken): BnsUsernameNft {
+  return {
+    id: fromUtf8(nft.base!.id!),
```
ethanfrey: What I often do in these cases, where the data comes from outside, is to use `ensure`, which will throw a more useful error than just `.id!`. (or at least should throw a better error)

ethanfrey: ex: `const sig = ensure(sigs[0], "first signature");`

So error is "first signature is missing", not "cannot access id of undefined"

packages/iov-bns/src/types.ts
```diff
@@ -1,12 +1,59 @@
...
 
-import { Algorithm, PublicKeyBundle, PublicKeyBytes, SignatureBytes } from "@iov/base-types";
-import { FullSignature, Nonce } from "@iov/bcp-types";
+import { Algorithm, ChainId, PublicKeyBundle, PublicKeyBytes, SignatureBytes } from "@iov/base-types";
+import { Address, ChainAddressPair, FullSignature, Nonce } from "@iov/bcp-types";
 import { Int53 } from "@iov/encoding";
 
 import * as codecImpl from "./generated/codecimpl";
 
+export interface BnsUsernamesByUsernameQuery {
```
ethanfrey: nice set of queries

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -416,6 +428,26 @@ export class BnsConnection implements BcpAtomicSwapConnection {
     );
   }
 
+  public async getUsernames(query: BnsUsernamesQuery): Promise<ReadonlyArray<BnsUsernameNft>> {
```
ethanfrey: very nice function, i like this.
good use of the paths and the arguments.

you are entering weave mastery :)

webmaster128: :)

packages/iov-bns/src/decode.ts
webmaster128: Thanks, right. Let's do what we always do. Will update for both decodeBlockchainNft/decodeUsernameNft in the other PR

webmaster128:
Closing in favour of #550

PR #550, opened 2018-11-21
[iov-bns] Add query blockchain support

### LINKED ISSUE #495: Add BNS query/watch support

webmaster128:
This is supposed to cover points 1-3 from #495 but without the watching part

Based on #549 

Closes #495

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -608,6 +608,80 @@ describe("BnsConnection", () => {
 
       connection.disconnect();
     });
+
+    it("can query username by (chain, address)", async () => {
+      pendingWithoutBnsd();
+      const connection = await BnsConnection.establish(bnsdTendermintUrl);
+      const registryChainId = await connection.chainId();
+
+      const profile = new UserProfile();
+      const wallet = profile.addWallet(Ed25519HdWallet.fromEntropy(await Random.getBytes(32)));
+      const identity = await profile.createIdentity(wallet.id, HdPaths.simpleAddress(0));
+      const identityAddress = keyToAddress(identity.pubkey);
+
+      // Register blockchain
+      const chainId = `wonderland_${Math.random()}` as ChainId;
+      const blockchainRegistration: RegisterBlockchainTx = {
+        kind: TransactionKind.RegisterBlockchain,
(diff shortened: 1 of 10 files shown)


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -570,6 +570,65 @@ describe("BnsConnection", () => {
...
+            identity,
+            blockchainRegistration,
+            bnsCodec,
+            await getNonce(connection, identityAddress),
+          ),
+        ),
+      );
+
+      // Query by chain ID
+      const results = await connection.getBlockchains({ chainId: chainId });
```
ethanfrey: Perfect. Love the registration and query interfaces.

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -428,6 +431,20 @@ export class BnsConnection implements BcpAtomicSwapConnection {
     );
   }
 
+  public async getBlockchains(query: BnsBlockchainsQuery): Promise<ReadonlyArray<BnsBlockchainNft>> {
```
ethanfrey: Nice simple query. Looks good.

packages/iov-bns/src/decode.ts
```diff
@@ -23,11 +23,43 @@ import {
 import { Encoding } from "@iov/encoding";
 
 import * as codecImpl from "./generated/codecimpl";
-import { asNumber, BnsAddressBytes, BnsUsernameNft, decodeFullSig, ensure } from "./types";
+import { asNumber, BnsAddressBytes, BnsBlockchainNft, BnsUsernameNft, decodeFullSig, ensure } from "./types";
 import { encodeBnsAddress, isHashIdentifier } from "./util";
 
 const { fromUtf8 } = Encoding;
 
+export function decodeBlockchainNft(nft: codecImpl.blockchain.IBlockchainToken): BnsBlockchainNft {
```
ethanfrey: Thanks for using `ensure` here.

PR #558, opened 2018-11-22
Create structured type Secp256k1Signature

webmaster128:
The problem with secp256k1 signatures is that there are multiple ways to encode them. DER is one common way but Ethereum needs a different way.

- [x] use in createSignature
- [x] use in verifySignature


webmaster128:
@LucasIsasmendi this gives you signature bytes that can be used as simple as

```ts
const sig = ExtendedSecp256k1Signature.fromFixedLength(signatureBytes);
const r = sig.r();
const s = sig.s();
const recoveryParam = sig.recovery;
```

for Ethereum

webmaster128:
The original commits are merged with #563 and the rest of this work lives in #536

PR #559, opened 2018-11-22
Update ganache-cli to ^6.2.2

### LINKED ISSUE #546: Update ganache-cli to 6.2+
when https://github.com/trufflesuite/ganache-cli/issues/595 is fixed

webmaster128:
Closes #546

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@types/jasmine": "^3.3",
     "@types/memdown": "^3.0.0",
     "cross-env": "^5.2.0",
-    "ganache-cli": "~6.1",
+    "ganache-cli": "^6.2.2",
     "jasmine": "^3.3",
     "jasmine-console-reporter": "^3.0.2",
     "jasmine2-custom-message": "^0.9.1",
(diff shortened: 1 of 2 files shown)


webmaster128:
Oh wow, that pulls in tons of dependencies. We should consider installing ganache via docker …

davidmurdoch:
@webmaster128 I know `iov-core` is yarn only, so this probably doesn't affect you: the `npm-shrinkwrap.json` file in v6.2.2 had an issue with some npm versions, so I've released v6.2.3.

As for as the dependency issues, we're planning on moving back to a webpacked build (6.1.* was all webpacked) in the very near future which will get rid of nearly all of these dependencies.

Just curious... is the concern with # of dependencies due to the increased installation time?

webmaster128:
@davidmurdoch thanks, then we go to 6.2.3 straight, even if it should not change anything.

> Just curious... is the concern with # of dependencies due to the increased installation time?

I was surprised but thinking about it again, I don't see any real issue with the number of dependencies.

webmaster128:
All green, thanks @davidmurdoch!

PR #561, opened 2018-11-23
Implement block info as return type of postTx

### LINKED ISSUE #413: Properly handle pending/unconfirmed/confirmed state on ethereum transactions
Builds on #410 / #411 

Since block times are long and there is no instant finality, we need to mark transactions as `Pending | Unconfirmed | Confirmed`

* `Pending`: successfully reach a blockchain node and is "in mempool"
* `Unconfirmed`: successfully written in a block, but cannot yet guarantee it won't be reverted
* `Confirmed`: confident the transaction will not be reverted (block depth depends per chain)

This may require updates to the APIs (if not addressed earlier for lisk), and will require additions to the ethereum implementation.

webmaster128:
Now we get

* immediate resolving promise when transaction was posted sucessfully
* a block info value (height, confirmations) for tendermint
* an `.updates` stream we can listen to if we care (no overhead if nobody is listening)

Closes #413

### DIFF
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -76,6 +76,8 @@ export interface BcpTransactionResponse {
 export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTransaction>
   extends SignedTransaction<T> {
   readonly height: number; // the block it was written to
+  /** number of blocks on top of the transaction's block */
+  readonly confirmations: number;
   readonly txid: TxId; // a unique identifier (hash of the data)
   /** Data from executing tx (result, code, tags...) */
   readonly result?: Uint8Array;
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -76,7 +76,7 @@ export interface BcpTransactionResponse {
 export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTransaction>
   extends SignedTransaction<T> {
   readonly height: number; // the block it was written to
-  /** number of blocks on top of the transaction's block */
+  /** depth of the transaction's block, starting at 1 as soon as transaction is in a block */
   readonly confirmations: number;
   readonly txid: TxId; // a unique identifier (hash of the data)
   /** Data from executing tx (result, code, tags...) */
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -77,9 +77,9 @@ export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTr
   extends SignedTransaction<T> {
   readonly height: number; // the block it was written to
   readonly txid: TxId; // a unique identifier (hash of the data)
-  // Data from executing tx (result, code, tags...)
-  readonly result: Uint8Array;
-  readonly log: string;
+  /** Data from executing tx (result, code, tags...) */
+  readonly result?: Uint8Array;
+  readonly log?: string;
   // readonly tags: ReadonlyArray<Tag>;
 }
 
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -1,6 +1,7 @@
 import { Stream } from "xstream";
 
 import { ChainId, PostableBytes, PublicKeyBundle, TxId } from "@iov/base-types";
+import { ValueAndUpdates } from "@iov/stream";
 
 import { Address, SignedTransaction, TxCodec } from "./signables";
 import { Nonce, TokenTicker, UnsignedTransaction } from "./transactions";
@@ -62,10 +63,31 @@ export interface BcpTicker {
   readonly tokenName: string;
 }
 
+export enum BcpTransactionState {
+  /** accepted by a blockchain node and in mempool */
+  Pending,
+  /** successfully written in a block, but cannot yet guarantee it won't be reverted */
+  InBlock,
+}
+
+/** Information attached to a signature about its state in a block */
+export type BcpBlockInfo =
+  | { readonly state: BcpTransactionState.Pending }
+  | {
+      readonly state: BcpTransactionState.InBlock;
+      /** block height, if the transaction is included in a block */
+      readonly height: number;
+      /** depth of the transaction's block, starting at 1 as soon as transaction is in a block */
+      readonly confirmations: number;
+    };
+
 export interface BcpTransactionResponse {
+  /** @deprecated use blockInfo instead */
   readonly metadata: {
     readonly height?: number;
   };
+  /** TODO: Make non-optional as soon as implemented everywhere */
+  readonly blockInfo?: ValueAndUpdates<BcpBlockInfo>;
   readonly data: {
     readonly message: string;
     readonly txid: TxId; // a unique identifier (hash of the data)
--- a/packages/iov-bcp-types/src/bcp.ts
+++ b/packages/iov-bcp-types/src/bcp.ts
@@ -86,8 +86,8 @@ export interface BcpTransactionResponse {
   readonly metadata: {
     readonly height?: number;
   };
-  /** TODO: Make non-optional as soon as implemented everywhere */
-  readonly blockInfo?: ValueAndUpdates<BcpBlockInfo>;
+  /** Information abot the block the transaction is in */
+  readonly blockInfo: ValueAndUpdates<BcpBlockInfo>;
   readonly data: {
     readonly message: string;
     readonly txid: TxId; // a unique identifier (hash of the data)
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -327,70 +327,73 @@ describe("BnsConnection", () => {
       connection.disconnect();
     });
 
-    it("can post transaction and watch confirmations", async done => {
+    it("can post transaction and watch confirmations", done => {
       pendingWithoutBnsd();
-      const connection = await BnsConnection.establish(bnsdTendermintUrl);
-      const chainId = await connection.chainId();
 
-      const { profile, mainWalletId, faucet } = await userProfileWithFaucet();
-      const faucetAddr = keyToAddress(faucet.pubkey);
-      const rcpt = await profile.createIdentity(mainWalletId, HdPaths.simpleAddress(2));
-      const rcptAddr = keyToAddress(rcpt.pubkey);
+      (async () => {
+        const connection = await BnsConnection.establish(bnsdTendermintUrl);
+        const chainId = await connection.chainId();
 
-      // construct a sendtx, this is normally used in the MultiChainSigner api
-      const sendTx: SendTx = {
-        kind: TransactionKind.Send,
-        chainId,
-        signer: faucet.pubkey,
-        recipient: rcptAddr,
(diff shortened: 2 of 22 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -149,13 +153,42 @@ export class BnsConnection implements BcpAtomicSwapConnection {
       const { checkTx, deliverTx } = txresp;
       throw new Error(JSON.stringify({ checkTx, deliverTx }, null, 2));
     }
+    const height = txresp.height!;
+
+    const firstEvent: BcpBlockInfo = {
```
ethanfrey: This is a good first start (want to comment on line 151 above), but we can actually use the broadcastTxSync interface to wait for `CheckTx` on the server (made it into mempool), but not block on the commit. And then subscribe the the tx to watch when it was committed to a block.

I would also have to check if the proper message was sent on error, but I think I did this in golang a while ago, so it does actually send the info over rpc.

ethanfrey: I would do that in a separate PR to refine the implementation.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -261,434 +263,353 @@ describe("BnsConnection", () => {
...
-        fractional: 1,
-        tokenTicker: cash,
-      },
-    };
+        const events = new Array<BcpBlockInfo>();
+        const subscription = result.blockInfo!.updates.subscribe({
+          next: info => {
+            events.push(info);
+
+            if (events.length === 3) {
```
ethanfrey: Good test.

But everytime I see such code, I wonder if we could write a nice utility to help with streams in Promise land (when we want to await on some condition). Maybe some pieces from https://github.com/iov-one/iov-core/blob/master/packages/iov-stream/src/reducer.ts

Like:

```ts
const collector= asArray(result.blockInfo!.updates.take(3))
await collector.finished();
const events = collector.value();
expect(events.length).toEqual(3);
expect(events[0]).toEqual(...)
```

I think the take(3) will also handle unsubscribing

packages/iov-lisk/src/liskconnection.ts
```diff
@@ -100,10 +106,40 @@ export class LiskConnection implements BcpConnection {
...
+                height: confirmedTransaction.height,
+                confirmations: confirmedTransaction.confirmations,
+              };
+
+              if (!equal(event, lastEventSent)) {
+                blockInfoProducer.update(event);
+                lastEventSent = event;
+              }
+            }
+          }, 3000);
```
ethanfrey: Can you pull this constant (poll frequency) into a named variable at the top of the file (or possibly as an optional argument to the LiskConnection constructor)?

packages/iov-lisk/src/liskconnection.ts
```diff
@@ -198,8 +234,40 @@ export class LiskConnection implements BcpConnection {
...
   }
 
-  public searchTx(_: BcpTxQuery): Promise<ReadonlyArray<ConfirmedTransaction>> {
-    throw new Error("Not implemented");
+  public async searchTx(query: BcpTxQuery): Promise<ReadonlyArray<ConfirmedTransaction>> {
+    if (query.height || query.minHeight || query.maxHeight || query.tags.length) {
+      throw new Error("Query by height, minHeight, maxHeight, tags not supported");
+    }
+
+    if (query.hash) {
```
ethanfrey: Nice to add this support to Lisk.
There is no indexing service to allow any other queries, is there?

packages/iov-stream/src/defaultvalueproducer.ts
```diff
@@ -0,0 +1,63 @@
+import { Listener, Producer } from "xstream";
+
+export interface DefaultValueProducerCallsbacks {
+  readonly onStarted: () => void;
+  readonly onStop: () => void;
+}
+
+// allows pre-producing values before anyone is listening
+export class DefaultValueProducer<T> implements Producer<T> {
```
ethanfrey: This is just moving it into it's own file and a bit of doc cleanup, right?

packages/iov-stream/src/valueandupdates.ts
```diff
@@ -19,11 +23,19 @@ export class ValueAndUpdates<T> {
...
+   */
+  public waitFor(search: SearchFunction<T> | T): Promise<void> {
+    const searchImplementation: SearchFunction<T> =
+      typeof search === "function" ? (search as SearchFunction<T>) : (value: T): boolean => value === search;
+
     return new Promise((resolve, reject) => {
       const subscription = this.updates.subscribe({
         next: newValue => {
-          if (newValue === value) {
+          if (searchImplementation(newValue)) {
```
ethanfrey: Since you seem to be updating this code a bit, how about handling error case here as well with a reject?

packages/iov-rise/src/riseconnection.ts
```diff
@@ -111,10 +117,38 @@ export class RiseConnection implements BcpConnection {
...
+              height: confirmedTransaction.height,
+              confirmations: confirmedTransaction.confirmations,
+            };
+
+            if (!equal(event, lastEventSent)) {
+              blockInfoProducer.update(event);
+              lastEventSent = event;
+            }
+          }
+        }, 10000);
```
ethanfrey: Again, please use a named constant for this.
Also a comment why it is 10s for rise and 3s for lisk (you told me different block times, but nice to add a mention in the code)

packages/iov-lisk/src/liskconnection.ts
webmaster128: > There is no indexing service to allow any other queries, is there?

We can query all kinds of things and probably most of the interface can be supported. But I only implemented what I need now, which already blows up this PR enough. But we can make it happen as soon as we need it :)

packages/iov-stream/src/defaultvalueproducer.ts
webmaster128: The most impotant change is the addition of the `onStarted` and `onStop` callbacks, that did not exist before. Before we used DefaultValueProducer in cases where producing updates was cheap and we produced no matter if anyone is listening or not. Now, producing updates requires polling, so we can enable/disable polling only when someone cares for the updates.

packages/iov-stream/src/valueandupdates.ts
webmaster128: Which error do you mean?

packages/iov-stream/src/defaultvalueproducer.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+    const stream = Stream.createWithMemory(producer);
+
+    expect(producerActive).toEqual([]);
+
+    const subscription1 = stream.subscribe({});
+    expect(producerActive).toEqual([true]);
+
+    const subscription2 = stream.subscribe({});
+    expect(producerActive).toEqual([true]);
```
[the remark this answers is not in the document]
webmaster128: This is because the producer is started as soon as 1 or more subscriptions exist. At that point you can add as many subscriptions to the producer as you want. When the amount of subscriptions drops to 0, the producer stops.

The `producerActive` array is a list of _one or more are listening_ and _no one is listening_ values.

Any idea how to improve readability? I could create an enum:

```
enum SomeoneListening {
  Yes,
  No
}
```

instead of the boolean. Do you think that helps?

packages/iov-bns/src/bnsconnection.ts
webmaster128: Having BNS resolve after block and not after successful adding to the mempool is indeed inconsistent. It is not a huge issue because Tendermint resolves within a second or two.

> And then subscribe the the tx to watch when it was committed to a block.

Do we have code to do that (watch a transaction)? I had to use the block headers to generate my data because I did not find something.

packages/iov-bcp-types/src/bcp.ts
```diff
@@ -62,10 +63,31 @@ export interface BcpTicker {
...
+      readonly confirmations: number;
+    };
+
 export interface BcpTransactionResponse {
+  /** @deprecated use blockInfo instead */
   readonly metadata: {
     readonly height?: number;
   };
+  /** Information abot the block the transaction is in */
+  readonly blockInfo: ValueAndUpdates<BcpBlockInfo>;
```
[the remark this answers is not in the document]
webmaster128: I created this test to cover that question: https://github.com/iov-one/iov-core/pull/561/files#diff-9ec4e6b9247ddaaed82d11713281b0b4R339

Resolving means the transaction accepted by a node. Everything after that happens in the stream.

packages/iov-rise/src/riseconnection.ts
webmaster128: Good point, done

packages/iov-bns/src/bnsconnection.ts
ethanfrey: watchTx based on hash will get an event when it is writen to a block.

If you subscribe before the commit, you will get notified when it is inside. Then you can listen to blockheaders to increment the confirmation. 

This is just a change to the blocking call up to the first confirmation

packages/iov-stream/src/valueandupdates.ts
ethanfrey: The stream subscription can call
Next (n times) 
Completed (0 or 1)
Error (0 or 1)

If error is called, we don't reject the promise

packages/iov-stream/src/defaultvalueproducer.spec.ts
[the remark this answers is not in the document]
ethanfrey: It looks good to me as is.

I just had to scroll up to see the push onStart and onStop and it makes sense. Since that is obvious effect, I'd leave it as is.

packages/iov-bcp-types/src/bcp.ts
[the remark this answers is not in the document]
ethanfrey: I understand why. It just seems there are two flows.

1. Initial push to server fails -> promise error
2. Initial push to server succeeds - > promise resolves and pending emitted on stream
3. TX rejected from block - > error on stream
4. TX entered block - > committed on stream

Couldn't we just emit everything on a stream?

Then only issue is when the client wants to wait for mempool before proceeding. In such case, we could provide a helper, like 

```ts
const result = conn.postTx(tx)
await untilMempool(result.updates)
```

In the cases you don't need this extra check, you can just subscribe to the stream, and handle immediate or delayed rejection with the same code.

It feels cleaner to me, but is more complex if most clients will want to await on entering mempool and have different error handler in that case.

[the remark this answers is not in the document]
webmaster128: There are good reasons for the current design, even if it is a bit complex for the wallet:

- In many cases you just want to know then the transaction is successfully sent (tests, faucet, etc.) and do not care what happens then. For those the `Promise<Stream>` is much simpler to use. In those cases we never start the producer, avoiding a lot of unnecessary overhead (like polling data nobody cares about)
- Errors in 1. and 3. are entirely different class of problems that a wallet needs to handle differently. 1. is _you messed up creating a valid transaction_ and 3. is _some consensus problem_.
- Currently we have no way to detect 3.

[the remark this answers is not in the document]
webmaster128: So I'd consider it a two step process and not _two flows_.

packages/iov-bns/src/bnsconnection.ts
webmaster128: Thanks. Extracted to reduce the diff https://github.com/iov-one/iov-core/issues/564

PR #565, opened 2018-11-27
New coin format

### LINKED ISSUE #502: Convert whole and fraction to strings

webmaster128:
Closes #502


packages/iov-dpos/src/parse.spec.ts
```diff
@@ -3,41 +3,35 @@ import { ReadonlyDate } from "readonly-date";
...
-    const expected: Quantity = { whole: 1, fractional: 23000000 };
-    expect(Parse.parseQuantity("123000000")).toEqual(expected);
+    expect(Parse.parseQuantity("123000000")).toEqual("123000000");
   });
 
   it("can parse quantity 1.23456789", () => {
-    const expected: Quantity = { whole: 1, fractional: 23456789 };
-    expect(Parse.parseQuantity("123456789")).toEqual(expected);
+    expect(Parse.parseQuantity("123456789")).toEqual("123456789");
   });
```
LucasIsasmendi: Add parse example for whole: 0, fractional: 23

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -178,7 +177,7 @@ describe("BnsConnection", () => {
     expect(addrAcct.name).toEqual("admin");
     expect(addrAcct.balance.length).toEqual(1);
     expect(addrAcct.balance[0].tokenTicker).toEqual(cash);
-    expect(addrAcct.balance[0].whole).toBeGreaterThan(1000000);
+    expect(Number.parseInt(addrAcct.balance[0].quantity, 10)).toBeGreaterThan(1000000_000000000);
```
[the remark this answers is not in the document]
webmaster128: In JavaScript `_` is an optional separator that has no meaning other then helping readability. I think this can be context specific. In Lisk e.g. you might want to express 3, 5, 14 LSK as 3_00000000, 5_00000000, 14_00000000 but 60 seconds can be expressed as 60_000 ms.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1190,7 +1184,11 @@ describe("BnsConnection", () => {
     expect(faucetAcct.value()!.balance.length).toEqual(1);
     const end = faucetAcct.value()!.balance[0];
     expect(end).not.toEqual(start);
-    expect(end.whole + 680).toEqual(start.whole);
+    expect(end.quantity).toEqual(
+      Long.fromString(start.quantity)
```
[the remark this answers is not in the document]
webmaster128: @iov/bns already has the package "long" as a dependency (because of protobuf) and bn.js is not available. The numbers is this test are known to fit in the signed 64 bit range of `Long`.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1342,8 +1340,8 @@ describe("BnsConnection", () => {
       recipient: rcptAddr,
       amount: [
         {
-          whole: 21,
-          fractional: 0,
+          quantity: "21000000000",
+          fractionalDigits: 9,
           tokenTicker: cash,
```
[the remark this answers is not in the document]
webmaster128: `quantity: "1234", fractionalDigits: 9,`

packages/iov-bns/src/context.ts
```diff
@@ -0,0 +1,123 @@
...
+      balance: ensure(acct.coins).map(c => this.coin(c)),
+    };
+  }
+
+  public coin(coin: codecImpl.x.ICoin): BcpCoin {
+    const amount = decodeAmount(coin);
+    return this.amountToCoin(amount);
+  }
+
+  public swapOffer(swap: codecImpl.escrow.Escrow & Keyed): BcpAtomicSwap {
```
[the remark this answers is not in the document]
webmaster128: I had to refactor the whole Normalize class into the `Context`, which merges `Amount` and `BcpTicker` into `BcpCoin`.

packages/iov-bns/src/decode.spec.ts
```diff
@@ -94,6 +102,38 @@ describe("Decode", () => {
...
+      },
+      nonce: new Int53(7) as Nonce,
+    });
+  });
+
+  it("has working decodeToken", () => {
+    const token: codecImpl.namecoin.IToken & Keyed = {
+      _id: toUtf8("TRASH"),
+      name: "Trash",
+      sigFigs: 22, // Will be ignored. It is always 9.
```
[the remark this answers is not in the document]
webmaster128: `sigFigs` is a name that comes from the `codecImpl.namecoin.IToken` interface of the backend. This interface is auto-generated by protobuf and we cannot change it.

packages/iov-bns/src/decode.ts
```diff
@@ -82,10 +97,35 @@ export function decodeUsernameNft(nft: codecImpl.username.IUsernameToken): BnsUs
...
+  return {
+    tokenTicker: Encoding.fromAscii(data._id) as TokenTicker,
+    tokenName: ensure(data.name),
+  };
+}
+
 export function decodeAmount(coin: codecImpl.x.ICoin): Amount {
+  const fractionalDigits = 9; // fixed for all tickers in BNS
+
+  const wholeNumber = asNumber(coin.whole);
```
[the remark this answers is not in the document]
webmaster128: Because `number` is sufficiently large when dealing with (whole, fractional) as separate components.

packages/iov-bns/src/encode.ts
```diff
@@ -47,11 +57,11 @@ export function encodePrivkey(privateKey: PrivateKeyBundle): codecImpl.crypto.IP
 }
 
 export function encodeAmount(amount: Amount): codecImpl.x.Coin {
+  const whole = Int53.fromString(amount.quantity.slice(0, -amount.fractionalDigits) || "0");
```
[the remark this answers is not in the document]
webmaster128: Because Int53 is large enough to hold all amounts

packages/iov-bns/src/bnsconnection.spec.ts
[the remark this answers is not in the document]
LucasIsasmendi: In one hand we have the exercise to analyze the max amount of a blockchain, then decide if it fits in specific range (Long, Int53, Uint64), in the other hand, use big number for all the blockchains. What is the advantage of have specific types for different ranges against use bignumber for all, now that we have quantity as string and not as number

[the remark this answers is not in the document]
webmaster128: Yeah, well, we could use a lowest common denominator, which is probably bn.js. But math on amounts is blockchain specific anyway. As long as all the chain implementations are conform to the common interface (integer strings now), the implementation can do what it wants and what is most appropriate given specific requirements and the available types.

In this case both `Long` and `BN` work fine but `Long` is already available.

packages/iov-bns/src/bnsconnection.spec.ts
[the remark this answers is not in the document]
webmaster128: I added a test for decoding (whole, fractional) -> string with `whole: 0` in this new commit https://github.com/iov-one/iov-core/pull/565/commits/c9a8e72bd9a14299bbf00fd415f0a73d087d0f0e

packages/iov-dpos/src/parse.spec.ts
webmaster128: Done: https://github.com/iov-one/iov-core/pull/565/commits/3179e31b326408404d937279b9a0723558ce58b6. However, Parse.parseQuantity got a little bit pointless now that we convert Lisk strings to BCP strings. Anyway, we can remove the function later

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -293,15 +291,12 @@ describe("BnsConnection", () => {
...
+      const gotMoney = await connection.getAccount({ address: recipient });
       expect(gotMoney).toBeTruthy();
       expect(gotMoney.data.length).toEqual(1);
       const paid = gotMoney.data[0];
       expect(paid.balance.length).toEqual(1);
-      // we may post multiple times if we have multiple tests,
-      // so just ensure at least one got in
-      expect(paid.balance[0].whole).toBeGreaterThanOrEqual(500);
-      expect(paid.balance[0].fractional).toBeGreaterThanOrEqual(75000);
+      expect(paid.balance[0].quantity).toEqual("5000075000");
```
ethanfrey: Maybe define the quantity as a const (same input as balance)?

packages/iov-bns/src/bnsconnection.spec.ts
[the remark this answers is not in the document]
ethanfrey: This test is explicitly for BNS, where it is guaranteed 9 fractional digits.
For ethereum, where the 18 digits overflow number, but work in long, yes, it is better to be more careful, as 100 ETH will overflow long.

I think fine as is, but good eye to look at it.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1342,8 +1340,8 @@ describe("BnsConnection", () => {
       recipient: rcptAddr,
       amount: [
         {
-          whole: 21,
-          fractional: 0,
+          quantity: "21000000000",
```
ethanfrey: you know what would make these tests easier to read... a little helper function.

```ts
amount: [
  {
    ...toQuantity("21.0", 9),
    tokenTicker: cash
  }
]

function toQuantity(decimal: string, fractionalDigits: number): {quantity: string, fractionalDigits: number} {
   ....
}
```

ethanfrey: And yes, this helper is just meant for tests, not for exporting to library clients, but it would make the tests easier to read and write

packages/iov-bns/src/context.ts
```diff
@@ -0,0 +1,123 @@
...
+      };
+    } else {
+      return {
+        kind: SwapState.Expired,
+        data: swap.data,
+      };
+    }
+  }
+
+  private amountToCoin(amount: Amount): BcpCoin {
```
ethanfrey: I like extracting the "context" for this function into a class with method and add all that use it.
It was a bit confusing before, and I think all my functions returning functions was not that clear, but  I didn't see a better way. This is nice.

packages/iov-bns/src/decode.ts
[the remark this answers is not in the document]
ethanfrey: Design of weave with whole and fractional as int64 was so they could fit into number.

That is also why we allow only 9 sigfigs, less than the 15 max that fit into number.
As long as whole is always less than 10^15, then number is fine, if not Long is guaranteed to fit.
It should be simple to use. I'd rather not use bn.js unless really needed. This is something I don't like about ethereum, adding all this complexity, when no one ever really needs those 18 places....

packages/iov-bns/src/decode.ts
```diff
@@ -82,10 +97,35 @@ export function decodeUsernameNft(nft: codecImpl.username.IUsernameToken): BnsUs
...
+  const fractionalNumber = asNumber(coin.fractional);
+
+  const quantity = Long.fromNumber(wholeNumber)
```
ethanfrey: This one can have errors if whole number is greater than 10^7, which is actually feasible (okay, someone moved 20% of the IOV tokens, at 8*10^7.... this will overflow. Please add a test case with such large number. (In fact starting balance of IOV reserve multisig may be close to that)

I am good with number (or Long) for whole and fractional independently, but combining them please use string operations. I do some similar string manipulations in the wallet, which aren't too hard: https://github.com/iov-one/wallet-demo/blob/master/src/logic/balances.ts (these are a bit different than your format, but it should only be a few lines to do so.

packages/iov-bns/src/encode.spec.ts
```diff
@@ -57,10 +58,49 @@ describe("Encode", () => {
...
   });
 
-  it("encodes amount", () => {
-    const amount = encodeAmount(coinJson);
-    const encoded = codecImpl.x.Coin.encode(amount).finish();
-    expect(Uint8Array.from(encoded)).toEqual(coinBin);
+  describe("encodeAmount", () => {
+    it("can encode amount 3.123456789 ASH", () => {
+      const amount: Amount = {
+        quantity: "3123456789",
```
ethanfrey: This is where the `toQuantity` helper I proposed above would look nice... it would just match the description.

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -21,8 +21,26 @@ export type SwapIdString = string & As<"swap-id">;
...
 export interface Amount {
-  readonly whole: number;
-  readonly fractional: number;
+  /**
+   * The quantity expressed as atimic units.
+   *
+   * Convert to whole and fractional part using
+   *   const whole = amount.quantity.slice(0, -amount.fractionalDigits);
+   *   const fractional = amount.quantity.slice(-amount.fractionalDigits);
+   * or to a floating point approximation (not safe!)
```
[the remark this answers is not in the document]
webmaster128: Because you need to draw the line somewhere. When integer string is the library interface, it should be the thing the library exposes. There are at least 3 different ways in here to parse the result and all require some kind of big int library. Also I expect users to have their own ways. Especially because every node10+ caller can use native BigInteger, which we can't supporting node 8.

The comment aims to explain how to understand the result.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Does the whole and the factional split matter for that test? This is 21000000000 CASH atomics. I don't like indirection in general as it intoduces more potental for bugs.

But what we could do is allow "_" in out integer strings just as in pure JavaScript. The API contract would allow them and any function dealing with `Amount` inputs must strip them off.

packages/iov-bns/src/decode.ts
webmaster128: Nice catch. I will use this docs as the max amount for tests am implementation, okay? https://github.com/iov-one/weave/blob/master/x/codec.proto#L15-L19

Do we need support for negative amounts?

packages/iov-bns/src/encode.spec.ts
webmaster128: Counter offer (as I wrote above): `quantity: "3_123456789"`

packages/iov-bcp-types/src/transactions.ts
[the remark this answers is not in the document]
webmaster128: If I was a caller I wanted to have a BigDecimal class that does not only parse and store but also do math

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: This would change parsing to

```ts
const whole = amount.replace("_", "").quantity.slice(0, -amount.fractionalDigits);
const fractional = amount.replace("_", "").quantity.slice(-amount.fractionalDigits);
```

ethanfrey: Sure, allowing _ in strings seems like an easy win.

packages/iov-bns/src/decode.ts
ethanfrey: yes protobuf comment is proper range.

We use negative amounts internally in weave to perform calculations  but don't allow it in any stored data or transactions

ethanfrey:
Yes, I like the proposal to add underscore to the string

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: A little bit if readability for a little bit of added complexity. Like that tradeof? https://github.com/iov-one/iov-core/pull/565/commits/2374e9b1df83dac1ddc86c0f5858c674a0bbb7da

webmaster128: This makes checks like `expect(tx).toEqual(sendTx);` invalid because `quantity` has multiple string represenations now that are all equal amounts (https://github.com/iov-one/iov-core/pull/565/commits/9dcce03a160ba6e6086fc95ff6e696558ad99cca).

webmaster128:
This fixes the issue with large ICoin amounts: https://github.com/iov-one/iov-core/pull/565/commits/64721a733897d3ce25c93a50ebc5af6090fd51bf

packages/iov-bns/src/bnsconnection.spec.ts
ethanfrey: facepalm.

yeah, this can be annoying. maybe revert my suggestion and stick with the original.
whatever you find best.

webmaster128: Yeah, will revert. I am afraid this was not the last surprise here

PR #566, opened 2018-11-28
Add eip155V function

### LINKED ISSUE #560: Add EIP-155 helper
Before we continue with #536, we need a well understood, well tested and well documented

```ts
function eip155V(chainId: number, recoveryParam: number): number
```

that gives the value `v`. Support for pre-eip155 can be ignored for now

LucasIsasmendi:
Closes #560

### DIFF
--- a/packages/iov-ethereum/src/encoding.ts
+++ b/packages/iov-ethereum/src/encoding.ts
@@ -7,3 +7,28 @@ export function toRlp(data: any): Uint8Array {
   const dataBuffer = rlp.encode(data);
   return Uint8Array.from(safeBufferValues(dataBuffer));
 }
+
+/** changes with each chain */
+export enum BlknumForkState {
+  /** before height 2,675,000 for mainnet */
+  Before,
+  /** for mainnet height 2,675,000 and after */
+  Forked,
+}
+
+/** Information attached to a signature about its state in a block */
+export type Eip155ChainId =
+  | { readonly forkState: BlknumForkState.Before }
+  | {
+      readonly forkState: BlknumForkState.Forked;
+      /** numeric chain ID as defined in EIP155 */
+      readonly chainId: number;
+    };
+
+export function eip155V(chain: Eip155ChainId, recoveryParam: number): number {
+  if (chain.forkState === BlknumForkState.Forked && chain.chainId > 0) {
+    // chain ID available
+    return chain.chainId * 2 + recoveryParam + 35;
+  }
+  throw new Error("transaction not supported before eip155 implementation");
+}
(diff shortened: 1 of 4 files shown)


packages/iov-ethereum/src/encoding.ts
```diff
@@ -7,3 +7,11 @@ export function toRlp(data: any): Uint8Array {
   const dataBuffer = rlp.encode(data);
   return Uint8Array.from(safeBufferValues(dataBuffer));
 }
+
+export function eip155V(chainId: number, recoveryParam: number): number {
```
webmaster128: Can `chainId` ever become 0? I think it did not exist before `FORK_BLKNUM`. So we can improve that parameter into:

```ts
export enum BlknumForkState {
  /** before height 2,675,000 */
  Before,
  /** height 2,675,000 and after */
  Forked,
}

/** Information attached to a signature about its state in a block */
export type Eip155ChainId =
  | { readonly forkState: BlknumForkState.Before }
  | {
      readonly forkState: BlknumForkState.Forked;
      /** numeric chain ID as defined in EIP155 */
      readonly chainId: number;
    };

export function eip155V(chainId: Eip155ChainId, recoveryParam: number): number {
  // ...
  if (chainId.forkState === BlknumForkState.Forked) {
    // chain ID available
  }
  // ..
}
```

This interface makes clear, we do not support transactions for block < 2,675,000 (yet).

This is no issue for new transactions but becoms interesting when validating incoming transactions.

LucasIsasmendi: FORK_BLKNUM = 2,675,000 is only valid for mainnet, each chain is different, so we will have to list in testconfig or constants the block number for each chain to use 0 or one of the values from the list. The question is if that logic is part of eip155V or the caller code?: 
- if we use this function to post transaction, and we are targeting ropsten but we use chainId 0, the node will return invalid signature (no need to ask for block number here)
- if we use this function to validate the signature from a previous transaction, we need to validate the block number where this eip was implemented for each chain

So, maybe right now introduce all this new logic is over complicate the eip155V function, which works fine to post a new transaction. in the case of validate previous transaction, we can create another function that handles chainId based on blocknumber, and eip155V will remains the same

webmaster128: > FORK_BLKNUM = 2,675,000 is only valid for mainnet, each chain is different

Good point, I did not consider this.

> the block number for each chain to use 0 or one of the values from the list

At the moment we don't need that list, as we just need to pass the information _forked_ into eip155V and we are _forked_ on all chains we post to, right?

But later we can have a different function in the caller that produces `Eip155ChainId` depending on height and chain, which needs to have the config available.

The comments in `BlknumForkState` need to be adapted to express the height differs from chain to chain and 2,675,000 is an example for mainnet.

PR #569, opened 2018-11-29
Expose bns headers

### LINKED ISSUE #568: Get block header from height
To fetch the time information for transactions, need to fetch the block header which has same height as transaction.
Currently, there's no function available for this.

ethanfrey:
Closes #568 

Quick patch for a 0.9.2 release.

Expose tendermint headers in BnsConnection. Also allow for subscriptions.
This should be exposed in BcpConnection as well, but that requires more design time and working with other blockchains (lisk, rise, ethereum) to find a common abstraction

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -467,19 +467,21 @@ describe("BnsConnection", () => {
     const header = await connection.getHeader(3);
     expect(header.height).toEqual(3);
     expect(header.appHash.length).toEqual(20);
+    connection.disconnect();
   });
 
   it("throws if it cannot get header", async () => {
     pendingWithoutBnsd();
     const connection = await BnsConnection.establish(bnsdTendermintUrl);
-    connection
+    await connection
       .getHeader(123456)
       .then(() => fail("must not resolve"))
-      .catch(error => expect(error).toMatch(/123456 doesn't exist/i));
-    connection
+      .catch(() => 0);
+    await connection
       .getHeader(-3)
       .then(() => fail("must not resolve"))
-      .catch(error => expect(error).toMatch(/123456 doesn't exist/i));
+      .catch(() => 0);
+    connection.disconnect();
   });
 
   it("can register a blockchain", async () => {
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -485,6 +485,7 @@ describe("BnsConnection", () => {
   });
 
   it("watches headers with same data as getHeader", async () => {
+    pendingWithoutBnsd();
     const connection = await BnsConnection.establish(bnsdTendermintUrl);
 
     const headers = lastValue(connection.watchHeaders().take(2));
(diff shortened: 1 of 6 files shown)


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -461,6 +461,27 @@ describe("BnsConnection", () => {
...
+    const header = await connection.getHeader(3);
+    expect(header.height).toEqual(3);
+    expect(header.appHash.length).toEqual(20);
+  });
+
+  it("throws if it cannot get header", async () => {
+    pendingWithoutBnsd();
+    const connection = await BnsConnection.establish(bnsdTendermintUrl);
+    connection
+      .getHeader(123456)
```
webmaster128: This will fail for development chains running 34 hours, which is not uncommon on developer machines. Can you use something significantly larger?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -461,6 +461,29 @@ describe("BnsConnection", () => {
...
+    pendingWithoutBnsd();
+    const connection = await BnsConnection.establish(bnsdTendermintUrl);
+    await connection
+      .getHeader(123456)
+      .then(() => fail("must not resolve"))
+      .catch(() => 0);
+    await connection
+      .getHeader(-3)
+      .then(() => fail("must not resolve"))
+      .catch(() => 0);
```
webmaster128: where did the nice error expect go?

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -435,6 +436,22 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+      throw new Error(`Header ${height} doesn't exist yet`);
+    }
+    const { header } = blockMetas[0];
+    if (header.height !== height) {
+      throw new Error(`Requested header ${height} but got ${header.height}`);
+    }
+    return header;
+  }
+
+  public watchHeaders(): Stream<Header> {
```
webmaster128: This obsoletes `public changeBlock(): Stream<number>` once it goes into BcpConnection, right?

ethanfrey: The other just returns a stream of numbers, this is a bit more complex.
But yes, when there is a clear interface, we can replace that in BcpConnection

packages/iov-stream/src/reducer.ts
```diff
@@ -34,12 +34,20 @@ export class Reducer<T, U> {
...
-        complete: () => resolve(),
-        error: (err: any) => reject(err),
+      const subscription = this.stream.subscribe({
         next: (evt: T) => {
           this.state = this.reducer(this.state, evt);
         },
+        complete: () => {
+          resolve();
+          // this must happen after resolve, to ensure stream.subscribe() has finished
+          subscription.unsubscribe();
```
webmaster128: I don't think the order is correct here. Caller code will continue to run after `resolve();` and it is unclear if and when `subscription.unsubscribe();` is executed.

You can safely unsubscribe first, since `stream.subscribe()` is done at the point where one of the `next`, `complete`, `error` callbacks are running.

ethanfrey: If you switch the error a number of tests fail inside of iov-stream.
I had it the other way at first

PR #580, opened 2018-12-03
Add searchTx

### LINKED ISSUE #406: Query and parse transactions from ethereum
Builds on #405 

We should be able to query for transactions by account. To list all SendTx sent from a given address, or sent to a given address (this may be multiple queries on ethereum node, but should expose over standard BcpConnection api).

The parser (at this point) doesn't have to handle any data fields, just simple ETH movements (no smart contract calls, ERC20, etc)

LucasIsasmendi:
Closes #406 

**WIP**
- [x] Search transaction by hash
- [x] Search transactions by account

### DIFF
--- a/packages/iov-ethereum/src/ethereumcodec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.ts
@@ -1,4 +1,4 @@
-import { ChainId, PostableBytes } from "@iov/base-types";
+import { Algorithm, ChainId, PostableBytes, PublicKeyBytes, SignatureBytes } from "@iov/base-types";
 import {
   Nonce,
   PrehashType,
@@ -13,10 +13,18 @@ import {
 import { ExtendedSecp256k1Signature } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
 
+import { constants } from "./constants";
 import { isValidAddress, keyToAddress } from "./derivation";
 import { BlknumForkState, Eip155ChainId, eip155V, toRlp } from "./encoding";
 import { Serialization } from "./serialization";
-import { encodeQuantity, encodeQuantityString, hexPadToEven } from "./utils";
+import {
+  decodeHexQuantity,
+  decodeHexQuantityNonce,
+  decodeHexQuantityString,
+  encodeQuantity,
+  encodeQuantityString,
+  hexPadToEven,
+} from "./utils";
 
 const { fromHex } = Encoding;
 
@@ -82,7 +90,51 @@ export const ethereumCodec: TxCodec = {
     throw new Error(`Not implemented tx: ${signed}`);
   },
   parseBytes: (bytes: PostableBytes, chainId: ChainId): SignedTransaction => {
-    throw new Error(`Not implemented bytes: ${bytes}, chainId: ${chainId}`);
+    const json = JSON.parse(Encoding.fromUtf8(bytes));
+    // signature
+    const r = Encoding.fromHex(json.r.replace("0x", ""));
+    const s = Encoding.fromHex(json.s.replace("0x", ""));
+    const v = decodeHexQuantity(json.v);
+    const signature = new Uint8Array([...r, ...s, v]) as SignatureBytes;
+
+    switch (json.type) {
+      case 0:
+        return {
+          transaction: {
+            chainId: chainId,
+            fee: {
+              quantity: decodeHexQuantityString(json.gas),
+              fractionalDigits: constants.primaryTokenFractionalDigits,
+              tokenTicker: constants.primaryTokenTicker,
+            },
+            signer: {
+              algo: Algorithm.Secp256k1,
+              data: json.from,
+            },
+            ttl: undefined,
+            kind: TransactionKind.Send,
+            amount: {
+              quantity: decodeHexQuantityString(json.value),
+              fractionalDigits: constants.primaryTokenFractionalDigits,
+              tokenTicker: constants.primaryTokenTicker,
+            },
+            recipient: json.to,
+            memo: json.input,
+          },
+          primarySignature: {
+            nonce: decodeHexQuantityNonce(json.nonce),
+            // fake pubkey, we cannot know this
+            pubkey: {
+              algo: Algorithm.Secp256k1,
+              data: new Uint8Array([]) as PublicKeyBytes,
+            },
+            signature: signature,
+          },
+          otherSignatures: [],
+        };
+      default:
+        throw new Error("Unsupported transaction type");
+    }
   },
   keyToAddress: keyToAddress,
   isValidAddress: isValidAddress,
--- a/packages/iov-ethereum/src/ethereumcodec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.ts
@@ -15,7 +15,7 @@ import { Encoding } from "@iov/encoding";
 
 import { constants } from "./constants";
 import { isValidAddress, keyToAddress } from "./derivation";
-import { BlknumForkState, Eip155ChainId, eip155V, toRlp } from "./encoding";
+import { BlknumForkState, Eip155ChainId, eip155V, getRecoveryParam, toRlp } from "./encoding";
 import { Serialization } from "./serialization";
 import {
   decodeHexQuantity,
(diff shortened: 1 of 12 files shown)


packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -84,63 +99,130 @@ describe("EthereumConnection", () => {
     connection.disconnect();
   });
 
-  it("can post transaction", async () => {
-    pendingWithoutEthereum();
+  describe("postTx", () => {
+    let postedTxId: TxId;
```
webmaster128: Tests should never depend on each other. This allows you to [randomize](https://jasmine.github.io/api/3.3/Configuration.html) [the oder](https://github.com/w3c/respec/issues/1552) and [focus](https://jasmine.github.io/2.1/focused_specs.html) or [exclude](https://jasmine.github.io/2.0/introduction.html#section-Pending_Specs) tests one by one.

It is better to just duplicate the posting

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -84,63 +99,130 @@ describe("EthereumConnection", () => {
...
+      };
+      const bytesToPost = ethereumCodec.bytesToPost(signedTransaction);
+
+      const result = await connection.postTx(bytesToPost);
+      postedTxId = result.data.txid;
+      expect(result).toBeTruthy();
+      expect(result.data.message).toBeNull();
+    });
+
+    it("can search previous transaction by hash", async () => {
```
webmaster128: I'd move this one to the searchTx because you assume posting works and test the search.

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -84,63 +99,130 @@ describe("EthereumConnection", () => {
...
-
-    const result = await connection.postTx(bytesToPost);
-    expect(result).toBeTruthy();
-    expect(result.data.message).toBeNull();
+  describe("searchTx", () => {
+    it("can search non-existing transaction by hash", async () => {
+      pendingWithoutEthereum();
+      const connection = await EthereumConnection.establish(base);
+      const nonExistingId = new Uint8Array([]) as TxId;
+      const results = await connection.searchTx({ hash: nonExistingId, tags: [] });
```
webmaster128: We should make `tags` optional (independent of this PR)

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -84,63 +99,130 @@ describe("EthereumConnection", () => {
...
-    const bytesToPost = ethereumCodec.bytesToPost(signedTransaction);
-
-    const result = await connection.postTx(bytesToPost);
-    expect(result).toBeTruthy();
-    expect(result.data.message).toBeNull();
+  describe("searchTx", () => {
+    it("can search non-existing transaction by hash", async () => {
+      pendingWithoutEthereum();
+      const connection = await EthereumConnection.establish(base);
+      const nonExistingId = new Uint8Array([]) as TxId;
```
webmaster128: For this input I'd rather see an exception being thrown because this indicates a development bug.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -173,9 +174,78 @@ export class EthereumConnection implements BcpConnection {
   public watchNonce(_: BcpAddressQuery | BcpPubkeyQuery): Stream<Nonce | undefined> {
     throw new Error("Not implemented");
   }
-
-  public searchTx(_: BcpTxQuery): Promise<ReadonlyArray<ConfirmedTransaction>> {
-    throw new Error("Not implemented");
+  public async searchTx(query: BcpTxQuery): Promise<ReadonlyArray<ConfirmedTransaction>> {
+    if (query.height || query.minHeight || query.maxHeight) {
+      throw new Error("Query by height, minHeight, maxHeight, tags not supported");
```
webmaster128: the "tags" in the message is wrong as we use them below

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -173,9 +174,78 @@ export class EthereumConnection implements BcpConnection {
...
+      const transactionId = Encoding.fromHex(hexPadToEven(txUncodified.data.result.hash)) as TxId;
+      return [
+        {
+          ...transaction,
+          height: height,
+          confirmations: confirmations,
+          txid: transactionId,
+        },
+      ];
+    } else if (query.tags[1].key === "account") {
```
webmaster128: I know this API is not great (we'll fix that) , but here we need to test all 3 tag `key`s.

packages/iov-ethereum/src/parse.ts
```diff
@@ -19,3 +21,43 @@ export class Parse {
     };
   }
 }
+
+export class Scraper {
+  public static parseBytesTx(bytes: PostableBytes, chainId: ChainId): SignedTransaction {
+    const json = JSON.parse(Encoding.fromUtf8(bytes));
+    const signature = new Uint8Array([]) as SignatureBytes;
```
webmaster128: Don't we get the signature?

packages/iov-ethereum/src/testconfig.ts
```diff
@@ -35,6 +39,10 @@ const local: EthereumNetworkConfig = {
   nonce: new Int53(0) as Nonce,
   gasPrice: "20000000000",
   gasLimit: "2100000",
+  waitForTx: 100,
+  scraperApi: "https://api-ropsten.etherscan.io/api",
```
webmaster128: Does api-ropsten.etherscan.io index your local chain?

You can make this optional:

```ts
  readonly scraper?: {
    readonly api: string;
    readonly chainId: ChainId;
    readonly address: Address;
  }
```

and use `scraper!.api` for chains you know `scraper` is set

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -82,7 +90,51 @@ export const ethereumCodec: TxCodec = {
     throw new Error(`Not implemented tx: ${signed}`);
   },
   parseBytes: (bytes: PostableBytes, chainId: ChainId): SignedTransaction => {
-    throw new Error(`Not implemented bytes: ${bytes}, chainId: ${chainId}`);
+    const json = JSON.parse(Encoding.fromUtf8(bytes));
+    // signature
+    const r = Encoding.fromHex(json.r.replace("0x", ""));
+    const s = Encoding.fromHex(json.s.replace("0x", ""));
+    const v = decodeHexQuantity(json.v);
+    const signature = new Uint8Array([...r, ...s, v]) as SignatureBytes;
```
webmaster128: The binary signature should be fixed length encoding of extended secp256k1 signature again.

```ts
 const signature = new ExtendedSecp256k1Signature(r, s, recoveryParam).toFixedLength() as SignatureBytes;
```

so we need a reverse operation to `eip155V`

webmaster128:
Cool stuff overall, looks promising

packages/iov-ethereum/src/ethereumconnection.spec.ts
LucasIsasmendi: Yes, you are right. Anyway..this is a work around, we should get a ganache database loaded with multiple transactions and blocks, such database can be installed from iov github

packages/iov-ethereum/src/ethereumconnection.spec.ts
LucasIsasmendi: done

packages/iov-ethereum/src/ethereumconnection.spec.ts
LucasIsasmendi: Yes

packages/iov-ethereum/src/ethereumconnection.spec.ts
LucasIsasmendi: good catch!

packages/iov-ethereum/src/parse.ts
LucasIsasmendi: Sadly not. A way to do this is get each full transaction from the node by calling searchTx by hash per each transaction, but also another way is to provide/get a scraper that returns signature values (r,s,v). We can define this step when we start to validate transactions into iov-core.

packages/iov-ethereum/src/ethereumcodec.ts
LucasIsasmendi: Done

packages/iov-ethereum/src/parse.ts
webmaster128: Alright, thanks for feedback

packages/iov-ethereum/src/testconfig.ts
```diff
@@ -35,6 +41,12 @@ const local: EthereumNetworkConfig = {
   nonce: new Int53(0) as Nonce,
   gasPrice: "20000000000",
   gasLimit: "2100000",
+  waitForTx: 100,
+  scraper: {
+    api: "https://api-ropsten.etherscan.io/api",
+    chainId: "3" as ChainId,
+    address: "0x0A65766695A712Af41B5cfECAaD217B1a11CB22A" as Address,
+  },
```
webmaster128: Is this supposed to be here? How can https://api-ropsten.etherscan.io/api index the local chain?

LucasIsasmendi: I made this way (ugly) to allow test in CI, but forgot to `export ETHEREUM_SCRAPER=1` to travis.sh, should I add the export value or skip this test in travis?

webmaster128: Since there is no scaper for the clocal chain, I would just skip it and set `scraper: undefined` in the config. scraper+blockchain should belong together. Otherwise it gets very confusing.

We can test other networks manually on dev machines and later add a scraper for the local chain.

PR #587, opened 2018-12-05
Fix encoding of transaction hash (lower to upper hex)

webmaster128:
Based on #586

This bug existed all the time but was never revealed because before #586 we could not subscribe to transaction hash updates.

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -664,7 +664,7 @@ describe("BnsConnection", () => {
   });
 
   describe("searchTx", () => {
-    it("can search for transactions", async () => {
+    it("can search for transactions by tags", async () => {
       pendingWithoutBnsd();
       const connection = await BnsConnection.establish(bnsdTendermintUrl);
       const chainId = await connection.chainId();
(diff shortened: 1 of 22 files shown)


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -716,8 +716,48 @@ describe("BnsConnection", () => {
...
+        },
+      };
+
+      const nonce = await getNonce(connection, faucetAddress);
+      const signed = await profile.signTransaction(mainWalletId, faucet, sendTx, bnsCodec, nonce);
+      const txBytes = bnsCodec.bytesToPost(signed);
+      const response = await connection.postTx(txBytes);
+      const transactionIdToSearch = response.data.txid;
+
+      await sleep(50); // Tendermint needs some time to update search index
```
LucasIsasmendi: Later we can move this to a test config parameter so we can switch between environments.

webmaster128: Jap! It took us a while to understand the issue here. Now we can deduplicate code and comment.

It turns out that some ms sleep are even needed on a fast local development machine. Those 50 ms work for local and Travis. For production networks, a longer sleep may be necessary. However, the tests only run against local chains. Let's see

Issue #570, opened 2018-11-29
Expose getHeader in BcpConnection


ethanfrey:
Builds on #568 (implementation for bns connection)

Add support for:
- [ ] Lisk
- [ ] Rise
- [ ] Ethereum

decide on a common, basic format for headers to expose over bcp, minimally height, timestamp, and blockhash, but possibly other information is common.

webmaster128:
Also:

- [ ] Do not use tendermint-rpc type `Header` in the interface
- [ ] Remove obsolete `changeBlock`, which can be used as `watchHeaders().map(h => h.height)`

webmaster128:
Extracted watch implementation for Lisk and RISE into https://github.com/iov-one/iov-core/issues/599 to ensure the general interface gets in easily.

PR #590, opened 2018-12-05
Add getHeader function

### LINKED ISSUE #577: Add support for getHeader in ethereum connection
This is needed to fully implement #570

Provide a function that returns a full ethereum header with all information normalized (eg. base64 encoding strings as Unit8Arrays). It should accept a height and return the corresponding header.

This header should have all data and can be normalized into a standard "bcp header" as another issue.

LucasIsasmendi:
Closes #577

### DIFF
--- a/packages/iov-ethereum/src/responses.ts
+++ b/packages/iov-ethereum/src/responses.ts
@@ -1,0 +1,11 @@
+import { ReadonlyDate } from "readonly-date";
+
+import { ChainId } from "@iov/base-types";
+
+export interface Header {
+  readonly chainId: ChainId;
+  readonly height: number;
+  readonly time: ReadonlyDate;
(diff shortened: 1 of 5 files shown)


packages/iov-ethereum/src/responses.ts
```diff
@@ -0,0 +1,11 @@
+import { ReadonlyDate } from "readonly-date";
+
+import { ChainId } from "@iov/base-types";
+
+export interface Header {
```
webmaster128: Can we call this `BlockHeader`? I know "header" was the name used so far but I think this is not descriptive enough for a newcomers to find their way. In https://github.com/iov-one/iov-core/issues/570 I'll use "block header" for all interfaces.

PR #594, opened 2018-12-06
Enable BNSD_ENABLED browser testing in @iov/bns and @iov/core 

webmaster128:


scripts/release.sh
```diff
@@ -1,101 +0,0 @@
-#!/bin/bash
```
LucasIsasmendi: why this script was deleted?

webmaster128: It is unused for a long time now. In the very early days (before IOV-Core was on npm), we used this script together with webpack to bundle the library for internal testing. I saw it again when I deleted the also unused "pack-node" script.

PR #604, opened 2018-12-11
Gracefully handle duplicate Tendermint subscriptions; deprecate Tendermint 0.20/0.21

### LINKED ISSUE #581: Gracefully handle duplicate subscriptions in bns
Currently, if I run `watchAccount` and `liveTx` on the same account, I will get an error from tendermint, as those both use the same subscription to monitor the transactions on tendermint.

Ideally, we could keep track of that client side, and the second call that wanted to use the subscription, could reuse the existing stream. So those two calls can work side by side.

Right now, this is forcing jumping a few hoops client side to watch them both

webmaster128:
A lot of cleanup and movement, so I suggest commit-by-commit review.

The two interesting commits are

* Reuse subscription streams when listening to the same query twice af86e2a
* Deprecate Tendermint 0.20./0.21.0 support 2723e4b

Closes #581


packages/iov-tendermint-rpc/types/types.d.ts
```diff
@@ -0,0 +1,6 @@
+import { As } from "type-tagger";
```
ethanfrey: I don't really see why you pulled this out from encoding, as they are specific encoding types, but not a blocker for me.

packages/iov-tendermint-rpc/src/requests.ts
```diff
@@ -161,25 +159,6 @@ export interface ValidatorsRequest {
   };
 }
 
-// DefaultParams provides pass-through for calls with no params
```
ethanfrey: Okay, these are boilerplate stuff... figured to be DRY.
They have no arguments every, but sure, can pull them out.

packages/iov-tendermint-rpc/src/client.ts
```diff
@@ -14,8 +14,20 @@ export class Client {
   }
 
   public static async detectVersion(client: RpcClient): Promise<Client> {
-    const adaptor = await findAdaptor(client);
-    return new Client(client, adaptor);
+    const req = jsonRpcWith(requests.Method.Status);
```
ethanfrey: Yes, this is a better place

packages/iov-tendermint-rpc/src/rpcclients/rpcclient.spec.ts
```diff
@@ -18,23 +18,31 @@ function pendingWithoutTendermint(): void {
...
-    expect(instanceOfRpcStreamingClient(new WebsocketClient(tendermintUrl))).toEqual(true);
+    const httpClient = new HttpClient(tendermintUrl);
+    const wsClient = new WebsocketClient(tendermintUrl);
+
+    expect(instanceOfRpcStreamingClient(httpClient)).toEqual(false);
+    expect(instanceOfRpcStreamingClient(wsClient)).toEqual(true);
+
+    httpClient.disconnect();
+    await wsClient.connected();
+    wsClient.disconnect();
```
ethanfrey: :+1: for all the disconnects

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.spec.ts
```diff
@@ -76,6 +78,42 @@ describe("WebsocketClient", () => {
...
+    const stream2 = client.listen(request2);
+
+    // tslint:disable-next-line:readonly-array
+    const eventHeights: any[] = [];
+
+    const subscription = Stream.merge(stream1, stream2).subscribe({
+      error: done.fail,
+      complete: () => done.fail("subscription should not complete"),
+      next: event => {
+        // height is string or number, depending on Tendermint version. But we don't care in this case
```
ethanfrey: I think this is correct, but reads a bit funny...
Maybe could be a bit cleaner

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -3,6 +3,7 @@ import { ReadonlyDate } from "readonly-date";
 import { Stream } from "xstream";
 
 import { Encoding } from "@iov/encoding";
+import { toListPromise } from "@iov/stream";
```
ethanfrey: Nice addition

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -426,6 +427,39 @@ function websocketTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor): void
     })().catch(done.fail);
   });
 
+  it("can unsubscribe and re-subscribe to the same stream", async () => {
+    pendingWithoutTendermint();
+
```
ethanfrey: Yes, this is cleaner, but I don't see the unsubscribe there

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.ts
```diff
@@ -29,6 +29,12 @@ export class WebsocketClient implements RpcStreamingClient {
   private readonly url: string;
   private readonly socket: SocketWrapper;
 
+  // Lazily create streams and use the same stream when listening to the same query twice.
```
ethanfrey: Clever... they can be reused, but subscribe/unsubscribe automatically

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.ts
```diff
@@ -77,8 +83,18 @@ export class WebsocketClient implements RpcStreamingClient {
...
     }
-    const producer = new RpcEventProducer(request, this.socket, this.bridge);
-    return Stream.create(producer);
+
+    const query = (request.params as any).query;
+    if (typeof query !== "string") {
+      throw new Error("request.params.query must be a string");
+    }
+
+    if (!this.subscriptionStreams.has(query)) {
```
ethanfrey: And this is the core of the whole PR :)

Nicely done

packages/iov-tendermint-rpc/types/types.d.ts
webmaster128: I think this belongs to responses, but I did not want to write `response.TxHash` all the time, especially because there are different response implementations.

packages/iov-tendermint-rpc/src/requests.ts
webmaster128: Yeah, I see. With that DRY I had to jump back and forth all the time when trying to follow the code. I guess there is a cost of DRY: indirection.

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.spec.ts
webmaster128: The core issue here is that we are testing an RPC client without knowing which backend this client is talking to. So we either need to make assumptions in the backend or write tests that are backend-independent.

packages/iov-tendermint-rpc/src/client.spec.ts
webmaster128: `toListPromise` unsubscribes after it got the number of elements in the second argument

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.ts
webmaster128: jap, thanks :) Great to have it deep deep down in the RPC client to make it work everywhere

packages/iov-tendermint-rpc/src/requests.ts
ethanfrey: Fair enough.

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.spec.ts
ethanfrey: Sorry, it wasn't the string/number issue, but rather how it was merge and set and check....

I mean you could take both streams and do something like 

```ts
const resA = await toListPromise(streamA, 2);
const resB = await toListPromise(streamB, 2);
expect(resA.length).toEqual(2);
expect(resA).toEqual(resB);
```

Or something along those lines... gets the same checks but reads cleaner to me.

packages/iov-tendermint-rpc/src/client.spec.ts
ethanfrey: Ohhh... very nice

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.spec.ts
webmaster128: Yeah, two aspects here:

* `toListPromise` is a very new helper that allows to write stream tests as ayncs/await instead of using `done` callbacks. This recent addition can probably cleanup dozens of tests, including this one.
* `toListPromise` on `streamA` causes a subscribe, capture, unsubscribe cycle, i.e. `resA` contains block headers number 1, 2 and `resB` contains block headers number 3, 4. Thus the Stream.merge is required.

To sum up, we can indeed do

```ts
const eventHeights = await toListPromise(Stream.merge(stream1, stream2).map(event => {
  return event.data.value.header.height;
}), 4);
expect(new Set(eventHeights).size).toEqual(2);

// that's it
```

webmaster128: This is how we can rewrite it now: https://github.com/iov-one/iov-core/pull/609/commits/e1344350bd00c35112838bf37092693bd6a8f2cd?utf8=%E2%9C%93&diff=split

ethanfrey: Looks good

PR #605, opened 2018-12-11
Decentralize transaction types

### LINKED ISSUE #603: Simplify extending Transaction types
Currently, we define `UnsignedTransaction` as a union type in `iov-bcp-types/src/transactions.ts`:

```ts
export type UnsignedTransaction =
  | AddAddressToUsernameTx
  | SendTx
  | SetNameTx
  | SwapOfferTx
  | SwapCounterTx
  | SwapClaimTx
  | SwapTimeoutTx
  | RegisterBlockchainTx
  | RegisterUsernameTx
  | RemoveAddressFromUsernameTx;
```

This means that every time we want to add a custom transaction for one blockchain (eg. bns), we must modify this core package. This is an annoyance for internal development, but a blocker for external development. I was contacted by a dev trying to use iov-core for a cosmos-sdk based blockchain and their main issue was adding custom transaction types.

The main usage of UnsignedTransaction is in signing:

```ts
/** A signable transaction knows how to serialize itself and how to store signatures */
export interface SignedTransaction<T extends UnsignedTransaction = UnsignedTransaction> {
  /** transaction is the user request */
  readonly transaction: T;

  readonly primarySignature: FullSignature;

  /** signatures can be appended as this is signed */
  readonly otherSignatures: ReadonlyArray<FullSignature>;
}

/** TxCodec knows how to convert Transactions to bytes for a given blockchain */
export interface TxCodec extends TxReadCodec {
  /** these are the bytes we create to add a signature */
  /** they often include nonce and chainID, but not other signatures */
  readonly bytesToSign: (tx: UnsignedTransaction, nonce: Nonce) => SigningJob;
  /** bytesToPost includes the raw transaction appended with the various signatures */
  readonly bytesToPost: (tx: SignedTransaction) => PostableBytes;
  /** identifier is usually some sort of hash of bytesToPost, chain-dependent */
  readonly identifier: (tx: SignedTransaction) => TransactionIdBytes;
}
```

Since many blockchains (eg. Lisk), only support a subset of the UnsignedTransaction types, it is not even providing us compile-time security.

I would propose the following adjustment to allow anything that is properly shaped to be used as a transaction:

```ts
export interface UnsignedTransaction extends BaseTx {
  readonly kind: string;
}
```

The generic transaction types should be in bns, also with corresponding guards:

```ts
function isSendTx(tx: UnsignedTransaction): tx is SendTx {
  return tx.kind === TransactionKind.Send
}

export type AtomicSwapTransaction =  |
  SwapOfferTx |
  SwapCounterTx |
  SwapClaimTx |
  SwapTimeoutTx;

function isAtomicSwapTx(tx: UnsignedTransaction): tx is AtomicSwapTransaction {
  return (tx.kind === TransactionKind.SwapOffer) ||
    (tx.kind === TransactionKind.SwapCounter) ||
    (tx.kind === TransactionKind.SwapClaim) ||
    (tx.kind === TransactionKind.SwapTimeout);
```

Extended blockchain-specific transactions, such as bns-specific ones, should be defined in that one package. bcp-types should only define shapes we expect to function over many blockchains.

In each package, when using the codec, the set of valid transactions (subset of `UnsignedTransaction`) should be enforced at runtime. eg.

```ts
  bytesToSign: (unsigned: UnsignedTransaction, nonce: Nonce): SigningJob => {
    if (!isSendTx(unsigned) && !isAtomicSwapTx(unsigned)) {
      throw new Error(`unsupported transction type: ${unsigned.kind}`);
     }
     // now, we know it is SendTx | AtomicSwapTx
     switch (unsigned.kind) {
       // ....
     }
  }
```

webmaster128:
Closes #603

- [x] Code
- [x] Add `isSwap{Offer,Claim,Counter,Timeout}Transaction`
- [x] CHANGELOG
- [x] READMEs
- [x] Rename helpers `isBnsSwapCounter`, `isBnsSwapRelease` to something that reflacts tey test confirmed transactins

-----------

Some notes on this:
* BaseTx was renamed to UnsignedTransaction, which is now the common interface.
* The term "tx" for transaction is pushed from BCP into Tendermint-world
* The BCP types `SendTransaction`, `SwapOfferTransaction`, `SwapCounterTransaction`, `SwapClaimTransaction`, `SwapTimeoutTransaction` are preserved as interfaces to be implemented by blockchains
* Lisk, RISE and Ethereum do not have explicit sub-types of `SendTransaction` but use it directly


packages/iov-bcp-types/src/transactions.ts
```diff
@@ -7,19 +7,12 @@ import { Address } from "./signables";
...
-// TODO: can't we just make this a number (block height?)
-export type TtlBytes = Uint8Array & As<"ttl">;
-
 // TokenTicker should be 3-4 letters, uppercase
 export type TokenTicker = string & As<"token-ticker">;
 
 export type SwapIdBytes = Uint8Array & As<"swap-id">;
 export type SwapIdString = string & As<"swap-id">;
 
-// TODO: we may want to make this a union type BNSName | PublicKey | Address
```
ethanfrey: Yeah, this is good... getting rid of some placeholders that were not needed

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -44,131 +37,62 @@ export interface Amount {
...
-  readonly kind: TransactionKind.AddAddressToUsername;
-  /** the username to be updated, must exist on chain */
-  readonly username: string;
-  readonly payload: ChainAddressPair;
 }
 
-export interface SendTx extends BaseTx {
-  readonly kind: TransactionKind.Send;
+export interface SendTransaction extends UnsignedTransaction {
+  readonly kind: "send";
```
ethanfrey: Yeah, I learned with redux actions that these enum descriptions are just verbose and don't add type safety. Cleaner this way

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -44,131 +37,62 @@ export interface Amount {
...
+/** The basic transaction type all transactions should extend */
+export interface UnsignedTransaction {
+  /**
+   * The domain in which the concrete transaction is valid
+   *
+   * This should be used for type detection only and not be encoded somewhere.
+   * Right now we use "bns", "ethereum", "lisk" and "rise" but this could also
+   * be migrated to a Java-style package names like "io.lisk.mainnet" later on.
+   */
+  readonly domain: string;
```
ethanfrey: Okay.... So I need to change this to use SendTransaction on lisk vs ethereum.

If we could add a `bcp` domain for all transactions defined in `iov-bcp-types` it seems it would be cleaner to allow the exact same transaction be cross-chain, while allowing each chain to carve out it's own extensions.

I need to see examples

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -44,131 +37,62 @@ export interface Amount {
...
 
-/**
- * Associates a simple name to an account on a weave-based blockchain.
- *
- * @deprecated will be dropped in favour of RegisterUsernameTx
- */
-export interface SetNameTx extends BaseTx {
-  readonly kind: TransactionKind.SetName;
-  readonly name: string;
+export function isSendTransaction(transaction: UnsignedTransaction): transaction is SendTransaction {
```
ethanfrey: This is good. If we have this, why not also `isSwapTransaction` or such?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -82,6 +88,7 @@ async function ensureNonceNonZero(
 ): Promise<void> {
   const nonce = await getNonce(connection, keyToAddress(identity.pubkey));
   const sendTx: SendTx = {
+    domain: "bns",
```
ethanfrey: Yeah, I was hoping this would be a domain for the `kind` (like isolating the class definitions), rather than a domain for the specific `object`.

So instead of `kind: "bcp/send"` and `kind: "bns/addUsername"`, we have

```ts
{
  domain: "bcp",
  kind: "send",
},
{
  domain: "bns",
  kind: "addUsername",
}
```

This allows us to avoid conflict in the use of kind in different packages, but doesn't add extra differentiation on the instances.

What do you think of this?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -318,7 +326,7 @@ describe("BnsConnection", () => {
       expect(mine.primarySignature.nonce).toEqual(nonce);
       expect(mine.primarySignature.signature.length).toBeTruthy();
       expect(mine.otherSignatures.length).toEqual(0);
-      const tx = mine.transaction;
+      const tx = mine.transaction as BnsTx;
```
ethanfrey: I like this. Better yet, we have an `isBnsTx()` guard to check it matches on runtime, and maybe `requireBnsTx()` helper that matches the guard or throws an error.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -470,19 +480,20 @@ describe("BnsConnection", () => {
       // Find registration transaction
       const searchResult = await connection.searchTx({ tags: [bnsNonceTag(identityAddress)] });
       expect(searchResult.length).toEqual(1);
-      if (searchResult[0].transaction.kind !== TransactionKind.RegisterBlockchain) {
+      const firstSearchResult = searchResult[0].transaction;
+      if (!isRegisterBlockchainTx(firstSearchResult)) {
```
ethanfrey: Yes, I like this

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -566,6 +580,7 @@ describe("BnsConnection", () => {
       // Register a blockchain
       const chainId = `wonderland_${Math.random()}` as ChainId;
       const blockchainRegistration: RegisterBlockchainTx = {
+        domain: "bns",
```
ethanfrey: And this is where I think the bns domain is quite useful

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -738,7 +757,7 @@ describe("BnsConnection", () => {
       const results = await connection.searchTx({ tags: [bnsFromOrToTag(rcptAddress)] });
       expect(results.length).toBeGreaterThanOrEqual(1);
       const mostRecentResult = results[results.length - 1];
-      expect(mostRecentResult.transaction.kind).toEqual(TransactionKind.Send);
+      expect((mostRecentResult.transaction as BnsTx).kind).toEqual(TransactionKind.Send);
```
ethanfrey: Again, guard or runtime check instead of just a cast, please.

packages/iov-bns/src/encode.ts
```diff
@@ -102,29 +110,32 @@ export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.app.ITx {
...
-      return buildRegisterUsernameTx(tx);
-    case TransactionKind.RemoveAddressFromUsername:
-      return buildRemoveAddressFromUsernameTx(tx);
-    default:
-      throw new Error("Received transacion of unsupported kind.");
+  if (!isBnsTx(tx)) {
+    throw new Error("Transaction is not a BNS transaction");
+  }
+
+  if (isAddAddressToUsernameTx(tx)) {
```
ethanfrey: Switch does look cleaner, but I see why you use this.
Same type safety in any case.

packages/iov-bns/src/types.ts
```diff
@@ -176,3 +194,153 @@ export const ensure = <T>(maybe: T | null | undefined, msg?: string): T => {
...
 };
+
+// transactions
+
+/**
+ * BNS transaction kinds.
+ *
+ * Raw values are used for type detection only and can change at any time.
+ */
+export enum TransactionKind {
```
ethanfrey: Huh? Why no TransactionKind in bcp, but there are here?

I could see getting rid of it everywhere, or using it everywhere, but why is it different?

packages/iov-bns/src/types.ts
```diff
@@ -176,3 +194,153 @@ export const ensure = <T>(maybe: T | null | undefined, msg?: string): T => {
...
+  SwapClaim = "swap_claim",
+  SwapTimeout = "swap_timeout",
+  RegisterBlockchain = "register_blockchain",
+  RegisterUsername = "register_username",
+  RemoveAddressFromUsername = "remove_address_from_username",
+}
+
+const bnsDomain = "bns";
+
+export interface BnsTx extends UnsignedTransaction {
```
ethanfrey: No union type?
I found those union types (along with ifs/switches) to be very nice to check all cases.
Here it seems a bit more manual boilerplate with the type detection functions everywhere, which may have bugs (rather than `tx.kind == ???`). Wonder why you made this choice.

packages/iov-bns/src/util.ts
```diff
@@ -94,16 +85,19 @@ export const hashFromIdentifier = (ident: HashId): Uint8Array => ident.slice(has
 export const bucketKey = (bucket: string) => Encoding.toAscii(`${bucket}:`);
 export const indexKey = (bucket: string, index: string) => Encoding.toAscii(`_i.${bucket}_${index}:`);
 
-export function isSwapOffer(tx: ConfirmedTransaction): tx is ConfirmedTransaction<SwapCounterTx> {
-  return tx.transaction.kind === TransactionKind.SwapCounter;
+export function isBnsSwapCounter(tx: ConfirmedTransaction): tx is ConfirmedTransaction<SwapCounterTx> {
```
ethanfrey: Yes, good improvements

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -246,10 +248,9 @@ describe("EthereumConnection", () => {
       const result = resultSearch[0];
       expect(result.transactionId).toEqual(resultPost.transactionId);
       const transaction = result.transaction;
-      if (transaction.kind !== TransactionKind.Send) {
+      if (!isSendTransaction(transaction)) {
```
ethanfrey: This custom `isSendTransaction` for every blockchain could be avoided if we just use the same domain for all `SendTransactions`. Or even just remove the whole domain idea, and hope we choose globally unique `kind`s

packages/iov-bcp-types/src/transactions.ts
webmaster128: Well spotted. No other reason than me being lazy plus the compiler did not force me to create them. Adding this to the list and resolving this comment.

packages/iov-ethereum/src/ethereumconnection.spec.ts
webmaster128: `isSendTransaction` is from BCP, not implemented per blockchain. It gives you a `SendTransaction`, which is an interface that describes a shape of a transaction with no domain included. `SendTransaction` just required the `kind` to equal "send"

packages/iov-bns/src/types.ts
webmaster128: In this case I do not really use an enum but abuse `enum` when I need a namespace for string constants. I tried to convert to an object with string properties but those I can only use as values, not as types. Maybe there is a better way to implement it. What I really want is "mini namespace of strings"

packages/iov-bns/src/encode.ts
webmaster128: Unfortunatly, TypeScript does not ensure switches are exhaustive on a language level. If it did, the switch gave us more safety. [There are hacks around that](https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript) but nothing really elegant.

We need the type checkers like `isAddAddressToUsernameTx` anyway because they check `UnsignedTransaction`s not only `BnsTx`s but this is a feature we do not use in this code block.

So this is mostly a readability thing and the switch indeed looked better.

packages/iov-bns/src/types.ts
webmaster128: It was not explicit choice other than: let's get rid of centralized lists. But inside a blockchain package we can of cource have those lists.

We need the manual boilerplate type detection functions because they check two things:

1. is BnsTx
2. is the specific kind of BnsTx we are looking for

and we don't want to have those two steps in the caller code.

So both ways work. I brought back the union type in https://github.com/iov-one/iov-core/pull/605/commits/fd0dfd28787014712cdb102f1441759715ecbfab

packages/iov-bns/src/encode.ts
```diff
@@ -114,28 +103,29 @@ export function buildMsg(tx: UnsignedTransaction): codecImpl.app.ITx {
...
-    return buildSwapClaimTx(tx);
-  } else if (isSwapCounterTx(tx)) {
-    return buildSwapCounterTx(tx);
-  } else if (isSwapOfferTx(tx)) {
-    return buildSwapOfferTx(tx);
-  } else if (isSwapTimeoutTx(tx)) {
-    return buildSwapTimeoutTx(tx);
-  } else {
-    throw new Error("Received transacion of unsupported kind.");
+  switch (tx.kind) {
```
ethanfrey: :)

packages/iov-bns/src/testdata.ts
```diff
@@ -146,7 +145,7 @@ const setNameMsg: SetNameTx = {
   chainId: "bns-mainnet" as ChainId,
   signer: pubJson,
   domain: "bns",
-  kind: TransactionKind.SetName,
+  kind: "set_name",
```
ethanfrey: yeah this looks good

packages/iov-bns/src/types.ts
```diff
@@ -282,65 +267,79 @@ export interface RegisterBlockchainTx extends BnsTx {
...
+  | SwapCounterTx
+  | SwapClaimTx
+  | SwapTimeoutTx
+  | RegisterBlockchainTx
+  | RegisterUsernameTx
+  | RemoveAddressFromUsernameTx;
+
 export function isBnsTx(transaction: UnsignedTransaction): transaction is BnsTx {
-  return transaction.domain === bnsDomain;
+  return (transaction as BnsTx).domain === "bns";
```
ethanfrey: Do you need this `as BnsTx` here?
Unless you just changed it, `UnsignedTransaction` should have a `domain: string` member, which is good enough for a check

webmaster128: Should indeed not be necessary

webmaster128:
The "domain" is the only way to reliably test if a `UnsignedTransaction` is a `BnsTx`. This feature is required e.g. in

```
export function buildMsg(tx: UnsignedTransaction): codecImpl.app.ITx {
  if (!isBnsTx(tx)) {
    throw new Error("Transaction is not a BNS transaction");
  }
  switch (tx.kind) {
```

where the whole switch only works because `tx` was reduced to `BnsTx`.

The domain is also the only way to distingush a transation of kind "send" in Lisk from a transaction of kind "send" in BNS. While in most cases this distinction is not necessary, in other cases it may be, e.g. when a codec extends the SendTransaction by custom properties.

packages/iov-bcp-types/src/transactions.ts
```diff
@@ -91,40 +87,40 @@ export interface SwapCounterTransaction extends UnsignedTransaction {
...
 
 export interface SwapTimeoutTransaction extends UnsignedTransaction {
-  readonly kind: "swap_timeout";
+  readonly kind: "bcp/swap_timeout";
   readonly swapId: SwapIdBytes; // pulled from the offer transaction
 }
 
 export function isSendTransaction(transaction: UnsignedTransaction): transaction is SendTransaction {
-  return transaction.domain === "bcp" && transaction.kind === "send";
+  return (transaction as SendTransaction).kind === "bcp/send";
```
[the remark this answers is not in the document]
webmaster128: This casts adds type safety for the value to check against. While

```ts
return transaction.kind === "send";
```

compiles, 

```ts
return (transaction as SendTransaction).kind === "send";
```

does not because it is always false. Thus I left the casts there to catch those kind of bugs.

It also follows the general rule (which I still don't like, but it is documented by TS): cast the incoming variable to the type to check for, do some checks that verify the casting was okay.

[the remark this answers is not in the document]
ethanfrey: Good explanation. Thanks!

webmaster128:
🎉

PR #608, opened 2018-12-12
Add bcp domain to SendTransaction

ethanfrey:
Demonstrate using some transactions that use a global domain (`bcp`) alongside blockchain-specific transaction extensions (`bns`).

Update SendTransaction usage:

- [x] Bns
- [x] Lisk
- [x] Rise
- [x] Ethereum

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ script:
 branches:
   only:
   - master
-  - 0.9
+  - decentralize-transaction-types
 
 notifications:
   email: false
--- a/packages/iov-bcp-types/src/transactions.ts
+++ b/packages/iov-bcp-types/src/transactions.ts
@@ -45,8 +45,15 @@ export interface UnsignedTransaction {
    * This should be used for type detection only and not be encoded somewhere.
    * Right now we use "bns", "ethereum", "lisk" and "rise" but this could also
    * be migrated to a Java-style package names like "io.lisk.mainnet" later on.
+   *
+   * We also use the special domain "bcp" for any kind that can be supported in multiple chains
    */
   readonly domain: string;
+  /**
+   * kind is the type of transaction (send, setName, etc)
+   * A (domain, kind) pair will uniquely define the expected shape of any transaction type
+   */
+  readonly kind: string;
   /** the chain on which the transaction should be valid */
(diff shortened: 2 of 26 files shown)


webmaster128:
I like this. Now that there are no different `("bns", "send")`, `("lisk", "send")` etc types anymore, `domain` could also be encoded in the kind:

`"bcp/send"`, `"bcp/swap_offer"`, `"bns/add_address_to_username"`, `"dpos/vote_delegate"`, `"rise/new_transaction_type_that_doe_not_exist_in_lisk"`

ethanfrey:
> "dpos/vote_delegate"

I like that... this is a transaction that works over 2-3 chains (lisk, rise, shift), but not outside of them.
`bcp` is like global, `bns` specific, and we have `dpos` etc as groupings....

ethanfrey:
> "bcp/send", "bcp/swap_offer", "bns/add_address_to_username", "dpos/vote_delegate", "rise/new_transaction_type_that_doe_not_exist_in_lisk"

I am fine with this format or (domain, kind). I think (domain, kind) is more explicit, but combining them less verbose. Take your pick.

ethanfrey:
Note that once we typeguarded to BnsTx union type, I can just switch on kind, as there is only one Transaction in that union with kind "send".

So, for switch and functionality it is the same. Just what seems a cleaner interface.

ethanfrey:
It all builds, basic tests pass.
I will run one suite against a local blockchain (what Travis does) before removing WIP.

webmaster128:
> I am fine with this format or (domain, kind). I think (domain, kind) is more explicit, but combining them less verbose. Take your pick.

This PR changes 2-dimensional data (`domain` and `kind` independent) to 1-dimensional data (`domain` fixed for each `kind`). Thus I think this is better represented as a single property.

ethanfrey:
Okay, this PR should pass and does the refactor for `(domain, kind)` as a dual element selector of the shape rather than a two-dimensional range. `domain` acts mainly as a scope for the kind, specifying what set of blockchains may support his (`bcp` being all blockchains in this library).

Feel free to merge this and then collapse to one dimensional. Or just take a look and then implement the two->one-dimensional collapse separately.

PR #613, opened 2018-12-13
Eth newheaders subscription

### LINKED ISSUE #407: Subscribe to new ethereum block headers
Builds on #402 but exposes `blocksChanges` Stream. Without polling.

LucasIsasmendi:
Closes #407

### DIFF
--- a/packages/iov-ethereum/src/ethereumconnection.spec.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.spec.ts
@@ -2,13 +2,16 @@ import { Algorithm } from "@iov/base-types";
 import {
   Address,
   BcpAccountQuery,
+  BlockHeader,
   isSendTransaction,
+  PostTxResponse,
   SendTransaction,
   SignedTransaction,
   TokenTicker,
   TransactionId,
 } from "@iov/bcp-types";
 import { HdPaths, Secp256k1HdWallet } from "@iov/keycontrol";
+// import { lastValue } from "@iov/stream";
 
 import { ethereumCodec } from "./ethereumcodec";
 import { EthereumConnection } from "./ethereumconnection";
@@ -40,6 +43,7 @@ function sleep(ms: number): Promise<void> {
 
 describe("EthereumConnection", () => {
   const base = TestConfig.base;
+  const ws = TestConfig.webSocket;
   const nodeChainId = TestConfig.chainId;
   const minHeight = TestConfig.minHeight;
   const address = TestConfig.address;
@@ -50,27 +54,27 @@ describe("EthereumConnection", () => {
 
   it(`can be constructed for ${base}`, () => {
     pendingWithoutEthereum();
-    const connection = new EthereumConnection(base, nodeChainId);
+    const connection = new EthereumConnection(base, nodeChainId, undefined);
     expect(connection).toBeTruthy();
   });
 
   it("can get chain ID", async () => {
     pendingWithoutEthereum();
-    const connection = await EthereumConnection.establish(base);
+    const connection = await EthereumConnection.establish(base, undefined);
     const chainId = connection.chainId();
     expect(chainId).toEqual(nodeChainId);
   });
 
   it("can get height", async () => {
     pendingWithoutEthereum();
-    const connection = await EthereumConnection.establish(base);
+    const connection = await EthereumConnection.establish(base, undefined);
     const height = await connection.height();
     expect(height).toBeGreaterThan(minHeight);
   });
 
   it("can get account from address", async () => {
     pendingWithoutEthereum();
-    const connection = await EthereumConnection.establish(base);
+    const connection = await EthereumConnection.establish(base, undefined);
     const query: BcpAccountQuery = { address: address as Address };
     const account = await connection.getAccount(query);
     expect(account.data[0].address).toEqual(address);
@@ -81,7 +85,7 @@ describe("EthereumConnection", () => {
 
   it("can get nonce", async () => {
     pendingWithoutEthereum();
-    const connection = await EthereumConnection.establish(base);
+    const connection = await EthereumConnection.establish(base, undefined);
 
     // by address
     {
@@ -96,8 +100,6 @@ describe("EthereumConnection", () => {
       const nonce = (await connection.getNonce(query)).data[0];
       expect(nonce).toEqual(TestConfig.nonce);
     }
-
-    connection.disconnect();
   });
 
   describe("postTx", () => {
@@ -133,7 +135,7 @@ describe("EthereumConnection", () => {
         },
         memo: "We \u2665 developers – iov.one",
       };
-      const connection = await EthereumConnection.establish(base);
+      const connection = await EthereumConnection.establish(base, undefined);
       const senderAddress = ethereumCodec.keyToAddress(mainIdentity.pubkey);
       const query: BcpAccountQuery = { address: senderAddress as Address };
       const nonceResp = await connection.getNonce(query);
@@ -165,23 +167,21 @@ describe("EthereumConnection", () => {
   describe("searchTx", () => {
     it("throws error for invalid transaction hash", async () => {
       pendingWithoutEthereum();
-      const connection = await EthereumConnection.establish(base);
+      const connection = await EthereumConnection.establish(base, undefined);
       // invalid lenght
       const invalidHashLenght = "0x1234567890abcdef" as TransactionId;
       await connection
         .searchTx({ id: invalidHashLenght })
         .then(() => fail("must not resolve"))
         .catch(error => expect(error).toMatch(/Invalid transaction ID format/i));
-      connection.disconnect();
     });
 
     it("can search non-existing transaction by hash", async () => {
       pendingWithoutEthereum();
-      const connection = await EthereumConnection.establish(base);
+      const connection = await EthereumConnection.establish(base, undefined);
       const nonExistingHash = "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef" as TransactionId;
       const results = await connection.searchTx({ id: nonExistingHash });
       expect(results.length).toEqual(0);
-      connection.disconnect();
     });
 
     it("can search previous posted transaction by hash", async () => {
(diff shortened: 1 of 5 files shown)


packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -269,43 +268,127 @@ describe("EthereumConnection", () => {
...
+      };
+      const bytesToPost = ethereumCodec.bytesToPost(signedTransaction);
+
+      const resultPost = await connection.postTx(bytesToPost);
+      return resultPost;
+    }
+
+    describe("watchBlockHeaders", () => {
+      it("watches headers with same data as getBlockHeader", async done => {
+        pendingWithoutEthereum();
```
webmaster128: Using `async done` together causes all kind of strangeness in Jasmine, e.g. pending() does not work anymore. Thus we should make this either `async` or use th `done` callback.

In this case I recommend a `async`+sleep implementation like I did in https://github.com/iov-one/iov-core/commit/1e9f80d845c24239f4de031cd2d1ac0dd52745ad

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -269,43 +268,127 @@ describe("EthereumConnection", () => {
...
       await connection
         .getBlockHeader(99999999999999)
         .then(() => fail("promise must be rejected"))
         .catch(err => expect(err).toMatch(/Header 99999999999999 doesn't exist yet/));
-      connection.disconnect();
+    });
+  });
+
+  describe("feed", () => {
+    async function postTransaction(amount: string, memo: string): Promise<PostTxResponse> {
```
webmaster128: `amount` -> `quantity`

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -40,6 +41,34 @@ import {
...
 } from "./utils";
 
+interface WsListener {
+  readonly id?: string;
+  readonly subscription?: string;
+}
+
+const wsListeners = new Array<WsListener>();
+
+function getListenerBySubscription(subscription: string): any {
```
webmaster128: Return type `WsListener | undefined`

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -53,16 +82,24 @@ async function loadChainId(baseUrl: string): Promise<ChainId> {
...
+  public static async establish(baseUrl: string, wsUrl: string | undefined): Promise<EthereumConnection> {
     const chainId = await loadChainId(baseUrl);
-    return new EthereumConnection(baseUrl, chainId);
+    let socketStream;
+    if (wsUrl) {
+      socketStream = new StreamingSocket(wsUrl);
+      socketStream.connect();
+      await socketStream.connected;
+    }
+    return new EthereumConnection(baseUrl, chainId, socketStream);
```
webmaster128: All this connection code cam move into constructor. We only need to load the ChainID here because this needs to be available in as a readonly member.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -53,16 +82,24 @@ async function loadChainId(baseUrl: string): Promise<ChainId> {
...
   }
 
   private readonly baseUrl: string;
   private readonly myChainId: ChainId;
+  private readonly socket: StreamingSocket | undefined;
 
-  constructor(baseUrl: string, chainId: ChainId) {
+  constructor(baseUrl: string, chainId: ChainId, socketStream: StreamingSocket | undefined) {
```
webmaster128: `constructor(baseUrl: string, chainId: ChainId, wsUrl: string | undefined)` and connect in here

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -72,7 +109,7 @@ export class EthereumConnection implements BcpConnection {
   }
 
   public disconnect(): void {
-    // no-op
+    this.socket!.disconnect();
```
webmaster128: Allow `disconnect()` to be called when `this.socket` is undefined

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -196,7 +233,62 @@ export class EthereumConnection implements BcpConnection {
...
-    throw new Error("Not implemented");
+    // see https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB#newheads
+    this.socket!.send(
+      JSON.stringify({
+        id: "watchBlockHeaders",
+        jsonrpc: "2.0",
+        method: "eth_subscribe",
+        params: ["newHeads", { includeTransactions: true }],
+      }),
+    );
```
webmaster128: Only send this when producer is started (`start: listener => {` block some lines below). Calling `watchBlockHeaders()` should not do heavy work before someone starts listening to the stream

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -196,7 +233,62 @@ export class EthereumConnection implements BcpConnection {
...
+        });
+      },
+      stop: () => {
+        headersSubscription.unsubscribe();
+      },
+    };
+    return Stream.create(producer);
+  }
+
+  public unwatchBlockHeaders(): void {
```
webmaster128: All the logic from here belongs to the `stop` block of the producer. That way you don't need to call `unwatchBlockHeaders()` explicitly but just unsubscribe from the stream.

packages/iov-ethereum/src/testconfig.ts
```diff
@@ -7,6 +7,7 @@ const { fromHex } = Encoding;
 export interface EthereumNetworkConfig {
   readonly env: string;
   readonly base: string;
+  readonly webSocket: string;
```
webmaster128: `webSocket` -> `webSocketUrl` to make clear this is not a connection

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: Sorry, but I need to see the full updated code and error to debug. This streaming logic is not easy to follow.

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -300,12 +306,102 @@ describe("EthereumConnection", () => {
...
+        const subscription = connection.watchBlockHeaders().subscribe({
+          next: info => {
+            events.push(info);
+            if (events.length === 2) {
+              expect(events[0].height).toEqual(events[1].height - 1);
+              connection.getBlockHeader(info.height).then(header => {
+                expect(header).toEqual(events[1]);
+                subscription.unsubscribe();
+                connection.unwatchBlockHeaders();
+                connection.disconnect();
```
webmaster128: Good. Now move `connection.disconnect();` after the sleep. Otherwise the socket is disconnected directly after the second event and before the second `await postTransaction("5445500", "watch header test 2-" + new Date(), connection);` is done

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -300,12 +306,102 @@ describe("EthereumConnection", () => {
...
+              });
+            }
+          },
+          complete: fail,
+          error: fail,
+        });
+        // post transactions
+        await postTransaction("5445500", "watch header test 1-" + new Date(), connection);
+        await postTransaction("5445500", "watch header test 2-" + new Date(), connection);
+        await sleep(100);
```
webmaster128: is 100 ms enough to wait for 2 blocks?

packages/iov-ethereum/src/ethereumconnection.ts
LucasIsasmendi: if you console.log `public disconnect()` and `public watchBlockHeaders() stop`, you will see in the output that `disconnect()` executes before `watchBlockHeaders() stop`, even if they are called in the right order: first stop (unsubscribe) and then disconnect.
So any logic inside `stop`, related with sockets, wont work.
Do you want me to push the error into CI? (it will be "Failed: Websocket is not open")

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -53,16 +82,25 @@ async function loadChainId(baseUrl: string): Promise<ChainId> {
 }
 
 export class EthereumConnection implements BcpConnection {
-  public static async establish(baseUrl: string): Promise<EthereumConnection> {
+  public static async establish(baseUrl: string, wsUrl: string | undefined): Promise<EthereumConnection> {
     const chainId = await loadChainId(baseUrl);
-    return new EthereumConnection(baseUrl, chainId);
+    const connection = new EthereumConnection(baseUrl, chainId, wsUrl);
+    if (connection.socket) {
+      await connection.socket.connected;
```
webmaster128: Can you move that waiting into a send helper that only waits before sending?

```ts
private async socketSend(data: string): Promise<void> {
  if (!this.socket) {
    throw new Error("No socket available");
  }
  await this.socket.connected;
  await this.socket.send(data);
}
```

This ensures the socket is connected also when we use the constructor directly.

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: Yeah I guess the issue is in the test logic (see different comment)

LucasIsasmendi: got it!

packages/iov-ethereum/src/ethereumconnection.spec.ts
LucasIsasmendi: with ganache blocks are automatic, changed to "waitForTx * 2" (200 ms)

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -294,4 +382,12 @@ export class EthereumConnection implements BcpConnection {
   public liveTx(_: BcpTxQuery): Stream<ConfirmedTransaction> {
     throw new Error("Not implemented");
   }
+
+  private async socketSend(data: string): Promise<void> {
+    if (!this.socket) {
+      throw new Error("No socket available");
+    }
+    await this.socket.connected;
+    this.socket.send(data);
```
webmaster128: `this.socket.send` returns a promise that we ignore here. When this promise is rejected, we get crashes impossible to debug. Either do

```
await this.socket.send(data);
```

or 

```
return this.socket.send(data);
```

Both ways do more or less the same thing in this case.

LucasIsasmendi: fixed!

PR #629, opened 2018-12-19
Multichainsigner bns eth

### LINKED ISSUE #572: MultiChainSigner: rewrite "can add two chains" test to use BNS + Ethereum
The test "can add two chains" uses BNS and tendermint chains, but raw tendermint does not support any BNS queries. So this case is relatively confusing when it should serve as part of the understanding process.

Since we have multiple chains available locally now, we can do BNS + Lisk or BNS + Ethereum.

LucasIsasmendi:
Closes #572

### DIFF
--- a/packages/iov-core/src/multichainsigner.spec.ts
+++ b/packages/iov-core/src/multichainsigner.spec.ts
@@ -8,6 +8,7 @@ import {
 } from "@iov/bcp-types";
 import { bnsCodec, bnsConnector, bnsFromOrToTag } from "@iov/bns";
 import { Random } from "@iov/crypto";
+import { ethereumConnector } from "@iov/ethereum";
 import { Ed25519HdWallet, HdPaths, LocalIdentity, UserProfile, WalletId } from "@iov/keycontrol";
 
 import { MultiChainSigner } from "./multichainsigner";
@@ -21,9 +22,9 @@ function pendingWithoutBnsd(): void {
   }
 }
 
-const pendingWithoutTendermint = () => {
-  if (!process.env.TENDERMINT_ENABLED) {
-    pending("Set TENDERMINT_ENABLED to enable tendermint-based tests");
+const pendingWithoutEthereum = () => {
+  if (!process.env.ETHEREUM_ENABLED) {
+    pending("Set ETHEREUM_ENABLED to enable ethereum-based tests");
   }
 };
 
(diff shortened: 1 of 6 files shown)


webmaster128:
CI fail due to compile error

packages/iov-core/src/multichainsigner.spec.ts
```diff
@@ -126,8 +127,10 @@ describe("MultiChainSigner", () => {
       expect(signer.chainIds().length).toEqual(1);
       const bovId = signer.chainIds()[0];
 
-      // add a raw tendermint chain (don't query, it will fail)
-      await signer.addChain(bnsConnector(rawTendermintUrl));
+      // add a ethereum chain
+      await signer.addChain(ethereumConnector(httpEthereumUrl));
+      expect(signer.chainIds().length).toEqual(1);
```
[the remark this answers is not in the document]
LucasIsasmendi: I did the change locally, forget the add before the amend

webmaster128:
Ah, one more thing: we need @iov/ethereum as a `devDependency` in this package

packages/iov-ethereum/package.json
```diff
@@ -43,5 +43,8 @@
     "readonly-date": "^1.0.0",
     "rlp": "^2.2.0",
     "xstream": "^11.7.0"
+  },
+  "devDependencies": {
+    "@iov/ethereum": "^0.10.2"
```
webmaster128: Now @iov/ethereum depends on @iov/ethereum ;) This needs to go into packages/iov-core

LucasIsasmendi: Ha!, wrong package.json file

packages/iov-core/src/multichainsigner.spec.ts
```diff
@@ -142,6 +144,13 @@ describe("MultiChainSigner", () => {
       expect(acct).toBeTruthy();
       expect(acct.data.length).toBe(1);
       expect(acct.data[0].balance.length).toBe(1);
+
+      const faucetAddr2 = signer.keyToAddress(ethId, faucet.pubkey);
```
webmaster128: This cannot work because the Ethereum codec cannot calculate an address from an Ed25519 pubkey (thus CI fail). We could use

```ts
const ganacheMainIdentity: PublicIdentity = {
  pubkey: {
    algo. Algorithm.Secp256k1,
    data: Encoding.fromHex("04965fb72aad79318cd8c8c975cf18fa8bcac0c091605d10e89cd5a9f7cff564b0cb0459a7c22903119f7a42947c32c1cc6a434a86f0e26aad00ca2b2aff6ba381") as PublicKeyBytes,
  }
}
```

instead of `faucet`

Issue #621, opened 2018-12-18
Add chainId to PublicIdentity


webmaster128:
A cryptographic keypair should be bound to one specific chain to avoid replay attacks. Also it should be easy to convert a PublicIdentity to an address, which requires the knowledge of the chainId.

- Add `chainId` parameter to `Wallet.createIdentity`
- Add `chainId` to `PublicIdentity`
- Remove redundant `chainId` parameter from `Wallet.createTransactionSignature`
- Convert `MultiChainSigner.keyToAddress` into `identityToAddress(identity: PublicIdentity): Address`
- Convert `TxReadCodec.keyToAddress` into `identityToAddress(identity: PublicIdentity): Address`

ethanfrey:
In general, I like the idea, but I have some issues with it as well, and think it needs a deeper design review.

1. The Wallet interface allows a number of derivation schemes, including importing raw keys. This is also exposed (indirectly) in the UserProfile. The approach posted by Isabella had a derivation path as a mandatory field, which would make issues here.
2. The selection of path for hd wallets happens outside of UserProfile in the client library. If we want to do this properly, we will need to use chain-specific derivations for each chain, which must be passed into the Wallet/UserProfile. At least a way to iterate over all possible derivations until we find one not yet used.
3. This then constrains the usage of the wallet beyond the generic functionality we currently provide. I think this makes the API quite opinionated, which is both good and bad.

I would propose a more comprehensive design and review of any new proposal.

And I would propose to use a new class (not UserProfile). Either wrap UserProfile, or have a ChainAwareProfile class that can be quite opinionated and provide a specific API. Note the faucet uses a different derivation path than the wallet for the same chain... this is a nice design, but would be impossible if we have a very opinionate profile. That said, I would support such an opinionated profile to be the standard export for anyone wanting to do user-facing key management, and have the more flexible/complex api for scripts and daemons and such

ethanfrey:
In such a case, we could gladly use `PublicKeyBundle`, `LocalKeyBundle` (with optional name attached) in the existing `UserProfile` and use the `*Identity` names for the new key manager.

webmaster128:
Regarding

1. Right, I would not restrict any of the current options and also not include the derivation path as part of the PublicIdentity. PublicIdentity just becomes

```ts
/** a public key we can identify with on a blockchain */
export interface PublicIdentity {
  readonly chainId: ChainId;
  readonly pubkey: PublicKeyBundle;
}
```

[...]

PR #635, opened 2018-12-20
Generic signer profile interface

ethanfrey:
Allow us to plug in other KeyManagement solutions that extend the UserProfile with app-specific logic.

As discussed in https://github.com/iov-one/iov-core/issues/621#issuecomment-449076733

### DIFF
--- a/packages/iov-core/src/multichainsigner.ts
+++ b/packages/iov-core/src/multichainsigner.ts
@@ -60,8 +60,8 @@ and calculate chain-specific addresses from public keys,
 even if bcp-proxy will handle translating all reads.
 */
 export class MultiChainSigner {
-  public readonly profile: Profile;
   private readonly knownChains: Map<string, Chain>;
+  private readonly profile: Profile;
 
   // initialize a write with a userProfile with secret info,
   // chains we want to connect to added with addChain (to keep async out of constructor)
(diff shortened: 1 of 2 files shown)


packages/iov-core/src/multichainsigner.ts
```diff
@@ -28,19 +30,42 @@ async function connectChain(x: ChainConnector): Promise<Chain> {
...
 
+/**
+ * TransactionSigner is just the methods on `UserProfile` that we need in `MultiChainSigner`.
+ * By only requiring this interface, we allow the use of other implementations with custom
+ * logic for key derivation, etc.
+ */
+export interface Profile {
+  readonly signTransaction: (
+    id: WalletId,
+    identity: PublicIdentity,
```
webmaster128: The existence of multiple wallets and the type `WalletId` is specific to the implementation. If we can guarantee that a `PublicIdentity` (soon a chainId-pubkey-pair) is unique within a Profile, the `id` parameter can be dropped. Such a check is probably not a bad idea.

However, out of scope for this one.

ethanfrey: I agree. Good idea but out of scope.

I would be happy to have `PublicIdentity` be globally unique over all `Wallets` and the `WalletId` disappear as unnecessary details.

But I think that is a different PR

PR #637, opened 2018-12-21
Add chain ID to public identity

### LINKED ISSUE #621: Add chainId to PublicIdentity

webmaster128:
Closes #621

- [x] Add `chainId` parameter to `Wallet.createIdentity`
- [x] Add `chainId` to `PublicIdentity`
- [x] Remove redundant `chainId` parameter from `Wallet.createTransactionSignature`
- [ ] Convert `MultiChainSigner.keyToAddress` into `identityToAddress(identity: PublicIdentity): Address`
- [ ] Convert `TxReadCodec.keyToAddress` into `identityToAddress(identity: PublicIdentity): Address`
- [ ] CHANGELOG


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -303,7 +306,9 @@ export class BnsConnection implements BcpAtomicSwapConnection {
   }
 
   public async getNonce(query: BcpAddressQuery | BcpPubkeyQuery): Promise<Nonce> {
-    const address = isPubkeyQuery(query) ? keyToAddress(query.pubkey) : query.address;
+    const address = isPubkeyQuery(query)
+      ? identityToAddress({ chainId: this.chainId(), pubkey: query.pubkey })
```
ethanfrey: this looks good

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -262,15 +262,18 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+  public async getAccount(query: BcpAccountQuery): Promise<BcpQueryEnvelope<BcpAccount>> {
     let response: QueryResponse;
-    if (isAddressQuery(account)) {
-      response = await this.query("/wallets", decodeBnsAddress(account.address).data);
-    } else if (isPubkeyQuery(account)) {
-      const address = bnsCodec.keyToAddress(account.pubkey);
+    if (isAddressQuery(query)) {
+      response = await this.query("/wallets", decodeBnsAddress(query.address).data);
+    } else if (isPubkeyQuery(query)) {
+      const address = bnsCodec.identityToAddress({
```
ethanfrey: can you remove `bnsCodec.` as you do in `getNonce` below?

packages/iov-bns/src/context.ts
```diff
@@ -47,7 +54,7 @@ export class Context {
   public wallet(acct: codecImpl.namecoin.IWallet & Keyed): WalletData {
     return {
       name: typeof acct.name === "string" ? acct.name : undefined,
-      address: encodeBnsAddress(acct._id),
+      address: encodeBnsAddress(addressPrefix(this.chainData.chainId), acct._id),
```
ethanfrey: ahhh.... so there is some global mapping of chainId to prefix... interesting

packages/iov-bns/src/context.ts
```diff
@@ -83,21 +90,25 @@ export class Context {
...
 
   public swapOfferFromTx(tx: ConfirmedTransaction<SwapCounterTransaction>): OpenSwap {
-    const counter: SwapCounterTransaction = tx.transaction;
+    const counterTransaction: SwapCounterTransaction = tx.transaction;
     // TODO: do we really want errors here, or just filter them out???
-    if (!isHashIdentifier(counter.hashCode)) {
+    if (!isHashIdentifier(counterTransaction.hashCode)) {
       throw new Error("swap not controlled by hash lock");
     }
+    const transactionCreator: PublicIdentity = {
```
ethanfrey: Nice! Pulling it from the Transaction...

Maybe there could be a helper if this is done many places?
`function transactionCreator(transaction: UnsignedTransaction): PublicIdentity`

packages/iov-bns/src/encode.ts
```diff
@@ -182,9 +191,13 @@ function buildSwapOfferTx(tx: SwapOfferTransaction): codecImpl.app.ITx {
 }
 
 function buildSwapCounterTx(tx: SwapCounterTransaction): codecImpl.app.ITx {
+  const creatorIdentity: PublicIdentity = {
```
ethanfrey: Yeah, this looks like the same destructuring again and again, I like my helper function idea....

packages/iov-bns/src/util.ts
```diff
@@ -21,9 +22,15 @@ import { Sha256 } from "@iov/crypto";
 import { Bech32, Encoding } from "@iov/encoding";
 import { QueryString } from "@iov/tendermint-rpc";
 
+const bnsMainnetChainId = "PLEASE_INSERT_HERE_WHEN_GENESIS_EXISTS" as ChainId;
```
ethanfrey: Ahhhh.... we hardcode one `iov` chainId, rest are `tiov`.
Decent decision.

What do you think of the idea of allowing the client to pass in this prefix in the constructor of the bnsCodec. Eg. make the bnsCodec itself no longer a singleton, but a class with constructor (needing a prefix arg)?

Same way we configure the BcpConnection with some nodes, we can also configure the codec with some parameters... these come from client-side config, or later directly from bns... no need to hardcode something that can be controlled externally, if we can do so safely

(this proposal of using as arg can be a separate PR. the current format is fine and can be merged like this, more me thinking of an additional refinement, if you like it, make a new issue)

packages/iov-bns/src/util.ts
```diff
@@ -46,9 +53,10 @@ function keyToIdentifier(key: PublicKeyBundle): Uint8Array {
   return Uint8Array.from([...algoToPrefix(key.algo), ...key.data]);
 }
 
-export function keyToAddress(key: PublicKeyBundle): Address {
-  const bytes = new Sha256(keyToIdentifier(key)).digest().slice(0, 20);
-  return encodeBnsAddress(bytes);
+export function identityToAddress(identity: PublicIdentity): Address {
```
ethanfrey: :+1:

packages/iov-cli/README.md
```diff
@@ -43,13 +43,19 @@ $ iov-cli
 
 ```
 > const profile = new UserProfile();
-> const wallet = Ed25519HdWallet.fromMnemonic("degree tackle suggest window test behind mesh extra cover prepare oak script")
-> profile.addWallet(wallet)
+> const signer = new MultiChainSigner(profile);
```
ethanfrey: Great job keeping this README up to date with all these changes :)

packages/iov-core/README.md
```diff
@@ -152,7 +152,7 @@ to use the chain-dependent `TxCodec` to generate it. In our case, bnsCodec:
 ```ts
 import { bnsCodec } from '@iov/bns';
 
-const addr = bnsCodec.keyToAddress(id1a.pubkey);
+const addr = bnsCodec.identityToAddress(id1a);
```
ethanfrey: And this one 👍

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -92,6 +94,8 @@ export const ethereumCodec: TxCodec = {
       otherSignatures: [],
     };
   },
-  keyToAddress: keyToAddress,
+  identityToAddress: (identity: PublicIdentity): Address => {
```
ethanfrey: This is same on all networks? interesting

packages/iov-keycontrol/src/keyring.spec.ts
```diff
@@ -121,7 +124,7 @@ describe("Keyring", () => {
         "wallets": [
           {
             "implementationId": "ed25519-hd",
-            "data": "{\\"formatVersion\\":1,\\"id\\":\\"f49h7gh34\\",\\"secret\\":\\"side ripple bachelor banner word swear buzz try situate rent desk carry scorpion uphold undo account pumpkin throw\\",\\"curve\\":\\"ed25519 seed\\",\\"identities\\":[]}"
+            "data": "{\\"formatVersion\\":2,\\"id\\":\\"f49h7gh34\\",\\"secret\\":\\"side ripple bachelor banner word swear buzz try situate rent desk carry scorpion uphold undo account pumpkin throw\\",\\"curve\\":\\"ed25519 seed\\",\\"identities\\":[]}"
```
ethanfrey: Hehehe... time to see our migrations in use.
Let's see if there are tests from migrating version 1 to version 2....

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -184,13 +198,23 @@ export class UserProfile {
     codec: TxCodec,
     nonce: Nonce,
   ): Promise<SignedTransaction> {
+    // identity and the pair (transaction.chainId, transaction.signer) are redundant
+    // but we keep both in the interface to be consistent with appendSignature() where
+    // the original transaction creator is not the signer
+    if (identity.chainId !== transaction.chainId) {
```
ethanfrey: Good check!

Interesting comment above. In a future refactor, removing WalletId as you mentioned, we could make this `signTransaction(transaction, codec, nonce)`. I like that.

Append signature clearly needs to have a specified identity, but if we can remove it here, I prefer it. (Issue for the future)

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -207,18 +231,17 @@ export class UserProfile {
     codec: TxCodec,
     nonce: Nonce,
   ): Promise<SignedTransaction> {
+    if (identity.chainId !== originalTransaction.transaction.chainId) {
```
ethanfrey: :+1:

packages/iov-keycontrol/src/wallets/ed25519wallet.ts
```diff
@@ -57,6 +57,12 @@ function deserialize(data: WalletSerializationString): Ed25519WalletSerializatio
   // Case distinctions / migrations based on formatVersion go here
   switch (formatVersion) {
     case 1:
+      throw new Error(
+        "Wallet format version 1 detected. " +
+          "No automatic migration is possible from that format since it is missing chain IDs in identities. " +
```
ethanfrey: :( At least the wallet is not public yet...

packages/iov-keycontrol/src/wallets/slip10wallet.ts
```diff
@@ -376,16 +383,27 @@ export class Slip10Wallet implements Wallet {
...
+        data: bytes,
+      },
     };
     return {
-      pubkey,
-      label,
-      id: Slip10Wallet.identityId({ pubkey }),
+      ...publicIdentity,
+      label: label,
+      id: Slip10Wallet.identityId(publicIdentity),
```
ethanfrey: I don't see changes to the `identityId` function, meaning it is the same regardless of chainId?

If this becomes an intrinsic part, we should add that to the `id` as well, right? (Especially since we are breaking any db migrations this time, let's break it even more)

packages/iov-ledger-bns/src/ledgersimpleaddresswallet.ts
```diff
@@ -300,15 +307,26 @@ export class LedgerSimpleAddressWallet implements Wallet {
     return out;
   }
 
-  private buildLocalIdentity(bytes: PublicKeyBytes, label: string | undefined): LocalIdentity {
-    const pubkey: PublicKeyBundle = {
-      algo: Algorithm.Ed25519, // hardcoded until we support more curves in the ledger app
-      data: bytes,
+  private buildLocalIdentity(
```
ethanfrey: This all seems fine. I was going to think more about how we use chainId in ledger, bu not really worth it until we revive the project... code is good for now, we need to revisit this all anyway when we reboot the ledger project

packages/iov-bns/src/context.ts
webmaster128: I have something different in mind that would clean up many many places: Convert `UnsignedTransaction.chainId` + `UnsignedTransaction.signer` into

```ts
interface UnsignedTransaction {
  creator: PublicIdentity;
  // ...
}
```

Creator is a better name when you think of multi signatures.

packages/iov-bns/src/util.ts
webmaster128: What I implemented here (1 mainnet vs. other = testnet) is how we defined the prefixed "iov"/"tiov", so this is the simplest implementation that serves our needs. I would not create unnecessary flexibility as bnsCodec's job is to implement BNS rules.

I like codec to be stateless and would not change this unless we have a very good reason to.

packages/iov-ethereum/src/ethereumcodec.ts
webmaster128: Jap. This interface helps us understanding blockchains :)

packages/iov-keycontrol/src/keyring.spec.ts
webmaster128: Unfortunatly no migration possible due to missing information

packages/iov-keycontrol/src/wallets/slip10wallet.ts
webmaster128: Great catch! This is a horrible bug

packages/iov-bns/src/bnsconnection.ts
webmaster128: Yes we can. `identityToAddress` and `bnsCodec.identityToAddress` are the exact same object

packages/iov-bns/src/context.ts
ethanfrey: Sure. More breaking to API, but that is fine now, since it is cleaner.

Not sure if it hides the fact the transaction is tied to one chain, but users should soon learn that PublicIdentities are chain-dependent.

Go for it, make this cleanup!

ethanfrey: I'm the biggest user... and deal with all breaking API changes in the wallet, so don't worry

webmaster128: > Not sure if it hides the fact the transaction is tied to one chain

It is not unless it is signed. One could argue that an UnsignedTransaction should not be tied to a signer or a chain and only get this connection at signing time. But we create those _to be signed by one identity_ and thus fix signer/chain before the signature comes in, which is also good.

webmaster128: https://github.com/iov-one/iov-core/issues/640

webmaster128:
Rebased onto current mater and added "wip" label since I'll not get the chance to work on the ID bug before christmas.

packages/iov-keycontrol/src/userprofile.ts
webmaster128: Jap, I think removing wallet ID and identity are indepeendent and can both be done as separate steps. https://github.com/iov-one/iov-core/issues/641

packages/iov-keycontrol/src/wallets/ed25519wallet.ts
```diff
@@ -93,7 +93,7 @@ export class Ed25519Wallet implements Wallet {
   }
 
   private static identityId(identity: PublicIdentity): LocalIdentityId {
-    const id = identity.pubkey.algo + "|" + Encoding.toHex(identity.pubkey.data);
+    const id = [identity.chainId, identity.pubkey.algo, Encoding.toHex(identity.pubkey.data)].join("|");
```
ethanfrey: :+1:

packages/iov-keycontrol/src/wallets/ed25519wallet.spec.ts
```diff
@@ -92,9 +92,23 @@ describe("Ed25519Wallet", () => {
     expect(newIdentity4.label).toEqual(lastIdentity.label);
   });
 
-  it("throws when adding the same keypair twice", async () => {
-    // Same keypair leads to the same identity identifier, so we don't support it
+  it("can create different identities with the same keypair", async () => {
```
ethanfrey: Good test, thought of the same thing when making that comment.
That every identity is tied to a chain, and we *encourage* using different keys for each chain, but nothing *prohibits* the reuse

PR #639, opened 2018-12-21
Fix levelup import in @iov/cli

webmaster128:


packages/iov-cli/src/cli.ts
```diff
@@ -82,7 +82,7 @@ export function main(originalArgs: ReadonlyArray<string>): void {
 
   let init = `
     import leveldown = require('leveldown');
-    import levelup = require('levelup');
+    import levelup from "levelup";
```
ethanfrey: This should work... we use `import levelup, { LevelUp } from "levelup";` in the wallet and it works great.

webmaster128: Jap, I could reproduce the type bug and this diff fixes it.

This is probably broken since we switched levelup types from the custom types file to @types/levelup

PR #646, opened 2018-12-27
Set singleRun to false by default

webmaster128:


webmaster128:
hmm, yeah for now it is just the quick-and-dirty removal of `--single-run` from package.json. I never had to use browser debugging a lot before working with WebWorkers that seem to be non-existent in nodejs. Not sure what the long term strategy is here

PR #648, opened 2018-12-28
Create RPC interface for out of process signer

### LINKED ISSUE #399: Implement "Signer?" interface with facade connection
This implements the interface in #575 to communicate with MultiChainSigner over a facade. 

Both client and server are in process for this case, but should only communicate over the specified interface to validate it works.

Implement the following use case:
* Initialize client and server to point to test blockchain
* Server initialized with a key and some tokens
* Client requests active identity from server
* Client queries current account balance
* Client sends a transaction request to the server to be posted
* Server validates, signs, and posts tx
* Client gets transaction hash back and queries new balance after sending.

### LINKED ISSUE #401: Verify "Signer" works over chrome extension message connections
This is similar to #400 but between a webapp and a chrome extension.

This may need manual testing, and this may belong in another repo. However, this issue can be closed if any demo exists that demonstrates a webapp submitting a request to be signed by the chrome extension

For testing, we can verify it works between a page and a web worker. (Same messaging as extension)

### LINKED ISSUE #575: Define high-level rpc interface for out-of-process signer
As discussed, we use a stateless protocol, either some rest/rpc type or fully asynchronous messages. (eg. send a push notification to a phone as request, get a message from long-polling a bridge server).

@webmaster128 can you link the presentation you made?

Also, look at https://docs.walletconnect.org/tech-spec for inspiration for connecting phone and desktop apps... It is ethereum-only, so I don't know if we can use any of the code, but at least look at the design for inspiration.

We need two main functionalities on the client side (beyond connecting)

* Request the current active identity (only if the user gives permission in the signer)
* Request the signer to sign a transaction (value name lookup, nonce lookup, and posting tx all occur in the signer. user must verify in the signer ui)

This replaces #397 although you can look there for some ideas as well.

Also, please take a look at #305 for some more half-finished ideas.

webmaster128:
Closes #575
Closes #399
Closes #401


packages/iov-core/src/servercore.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+  const defaultAmount: Amount = {
+    quantity: "1",
+    fractionalDigits: 9,
+    tokenTicker: "CASH" as TokenTicker,
+  };
+
+  it("can be constructed", () => {
+    const profile = new UserProfile();
+    const signer = new MultiChainSigner(profile);
+    const core = new ServerCore(profile, signer);
```
ethanfrey: This seems a bit of a stutter when constructing it.

What about `const core = new ServerCore(profile)` and have it instantiate the MultiChainSigner inside of it? Or are there other use cases for the internal signer?

packages/iov-core/src/servercore.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+      ),
+    );
+    const idB0 = copyToPublic(await profile.createIdentity(walletB.id, xnet, HdPaths.simpleAddress(0)));
+    const idB1 = copyToPublic(await profile.createIdentity(walletB.id, ynet, HdPaths.simpleAddress(1)));
+    const idB2 = copyToPublic(await profile.createIdentity(walletB.id, xnet, HdPaths.simpleAddress(2)));
+
+    const signer = new MultiChainSigner(profile);
+    const core = new ServerCore(profile, signer);
+
+    const ynetIdentities = await core.getIdentities("Login to XY service", [ynet]);
```
ethanfrey: Where is the confirm/deny logic triggered?
Would be good to see an example where the "user" rejects the request to show it throws.

packages/iov-core/src/servercore.ts
```diff
@@ -0,0 +1,80 @@
...
+import { UserProfile, WalletId } from "@iov/keycontrol";
+
+import { MultiChainSigner } from "./multichainsigner";
+
+function publicIdentityEquals(left: PublicIdentity, right: PublicIdentity): boolean {
+  return (
+    left.chainId === right.chainId &&
+    left.pubkey.algo === right.pubkey.algo &&
+    left.pubkey.data.length === right.pubkey.data.length &&
+    left.pubkey.data.every((value, index) => value === right.pubkey.data[index])
```
ethanfrey: Nice compare. I think this should be more general in bcp-types or such. To compare Uint8Arrays and then compare PublicIdentities.

packages/iov-core/src/servercore.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  public async getIdentities(
+    _: string,
+    chainIds: ReadonlyArray<ChainId>,
+  ): Promise<ReadonlyArray<PublicIdentity>> {
+    const matchingIdentities = this.allIdentities().filter(identity => {
+      return chainIds.some(chainId => identity.chainId === chainId);
+    });
+
+    // TODO: ask user for permission and allow selection of identities
```
ethanfrey: Maybe we pass in a "authorize" callback in the constructor, like

```
constructor(
  profile: UserProfile, 
  authorizeView: (reason: string, chains: ReadonlyArray<ChainId>) => Promise<boolean>,
  authorizeSign: (reason: string, transaction: UnsignedTransaction) => Promise<boolean>)
```

Or we could wrap ServerCore with another class that exposes the same public methods, yet performs some custom UI checks before calling this class?

Just ideas....

packages/iov-core/src/servercore.ts
```diff
@@ -0,0 +1,80 @@
...
+        throw new Error("More than one wallets contain the identity to sign this transaction");
+    }
+
+    // TODO: ask user for permission
+
+    const response = await this.signer.signAndPost(transaction, walletId);
+    return response.transactionId;
+  }
+
+  private allIdentities(): ReadonlyArray<PublicIdentity> {
```
ethanfrey: Nice function. Thinking this could belong on Keyring as well, as a generic function

packages/iov-core/src/workers/signingservice.worker.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+function parseRpcCall(data: JsonRpcRequest): RpcCall {
+  const params: ReadonlyArray<unknown> = data.params;
+
+  switch (data.method) {
+    case "getIdentities": {
+      const reason = params[0];
+      const chainIds = params[1];
+      if (typeof reason !== "string") {
```
ethanfrey: Nice runtime checks

packages/iov-core/src/workers/signingservice.worker.ts
```diff
@@ -0,0 +1,198 @@
...
+      throw new MethodNotFoundError("Unknown method name");
+  }
+}
+
+const bnsdUrl = "ws://localhost:22345";
+const bnsdFaucetMnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
+const ethereumUrl = "http://localhost:8545";
+const ganacheMnemonic = "oxygen fall sure lava energy veteran enroll frown question detail include maximum";
+
+async function main(): Promise<void> {
```
ethanfrey: This is a bit odd to me... Above looked like production code. This looks like test code. Why in same file (I assume something about webworker build process)

packages/iov-core/src/workers/signingservice.worker.ts
```diff
@@ -0,0 +1,198 @@
...
+  onmessage = async event => {
+    // console.log("Received message", JSON.stringify(event));
+
+    // filter out empty {"isTrusted":true} events
+    if (!event.data) {
+      return;
+    }
+
+    const response = await handleRequest(event);
+    postMessage(response);
```
ethanfrey: Huh? where does this come from?
Ah, this is webworker built-in, right?
Curious how this is matched with the request

packages/iov-jsonrpc/src/parse.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (id === null) {
+    throw new Error("Invalid id field");
+  }
+
+  const method = data.method;
+  if (typeof method !== "string") {
+    throw new Error("Invalid method field");
+  }
+
+  const params = data.params;
```
ethanfrey: AFAIK (from experience with tendermint),
Params may be an Array or an object with named arguments.

I personally find the object format easier to work with than the array format.

packages/iov-jsonrpc/src/parse.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  return {
+    jsonrpc: "2.0",
+    id: id,
+    method: method,
+    params: params,
+  };
+}
+
+export function parseJsonRpcError(data: any): JsonRpcErrorResponse | undefined {
```
ethanfrey: `data: unknown`?

packages/iov-jsonrpc/src/parse.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  return {
+    jsonrpc: "2.0",
+    id: id,
+    error: error,
+  };
+}
+
+export function parseJsonRpcResponse(data: any): JsonRpcResponse {
```
ethanfrey: `data: unknown`?

packages/iov-core/src/servercore.spec.ts
webmaster128: You need all the signer methods available to be able to connect to chains and from that you need chain IDs to create identities:

```ts
  const ed25519Wallet = profile.addWallet(Ed25519HdWallet.fromMnemonic(bnsdFaucetMnemonic));
  const secp256k1Wallet = profile.addWallet(Secp256k1HdWallet.fromMnemonic(ganacheMnemonic));
  const signer = new MultiChainSigner(profile);

  const bnsdConnection = (await signer.addChain(bnsConnector(bnsdUrl))).connection;
  const bnsdChainId = bnsdConnection.chainId();
  const ethereumConnection = (await signer.addChain(ethereumConnector(ethereumUrl, undefined))).connection;
  const ethereumChainId = ethereumConnection.chainId();

  // faucet identity
  await profile.createIdentity(ed25519Wallet.id, bnsdChainId, HdPaths.simpleAddress(0));
  // ganache second identity
  await profile.createIdentity(secp256k1Wallet.id, ethereumChainId, HdPaths.bip44(60, 0, 0, 1));

  const serverCore = new ServerCore(profile, signer);
```

I don't see a great way to simplify that right now but maybe ideas will come when working with that code a little more

packages/iov-core/src/servercore.spec.ts
webmaster128: Good point. Not implemented yet: https://github.com/iov-one/iov-core/issues/651

packages/iov-core/src/servercore.ts
webmaster128: Good idea, I prefer the callbacks. Added the approach to https://github.com/iov-one/iov-core/issues/651

packages/iov-core/src/servercore.ts
webmaster128: Indeed. Such a function makes sense in UserProfile as soon as signTransaction works without the `WalletId` argument (i.e. uniqueness of PublicIdentity in a UserProfile). https://github.com/iov-one/iov-core/issues/652

packages/iov-core/src/workers/signingservice.worker.ts
webmaster128: The (Web)Worker is it's own JS entry file and this file only exists for tests/demo.

However, we sould extract much of this test file into a reusable JSON-RPC wrapper around ServerCore (https://github.com/iov-one/iov-core/issues/650)

packages/iov-core/src/workers/signingservice.worker.ts
webmaster128: The globals `postMessage` and `onmessage` are available in the WebWorker API, which is added at compile time via `/// <reference lib="webworker" />` and added by the browser at runtime. This messaging system has no request-response semantics. You can push messages in both directions at any time. JSON-RPC will match those messages using the `id` property in JSON-RPC requests and responses.

packages/iov-jsonrpc/src/parse.ts
webmaster128: Oh wow, you are right: https://www.jsonrpc.org/specification#parameter_structures "4.2 Parameter Structures". I did not know about this option. Great! Will migrate to it

packages/iov-jsonrpc/src/parse.ts
webmaster128: In case of objects I have the following ptoblem: using runtime checks I narrow the type from `unknown` to `object`. But than I cannot use `data: object` because of

```
src/parse.ts:48:12 - error TS2339: Property 'jsonrpc' does not exist on type 'object'.

48   if (data.jsonrpc !== "2.0") {
              ~~~~~~~
```

I don't know how to specify the type "any object"

packages/iov-core/src/servercore.ts
webmaster128: Good point, implemented: https://github.com/iov-one/iov-core/pull/648/commits/c4187304ae6a4be01e7b3808ce79f2088e39e0eb

PR #654, opened 2019-01-02
Create JsonRpcSigningServer around SigningServerCore

### LINKED ISSUE #650: Create transport-agnostic JSON-RPC server around ServerCore
Create a

```ts
class JsonRpcSigningServer {
  constructor(core: ServerCore) {
  }

  public handleUnchecked(request: unknown): Promise<JsonRpcResponse | JsonRpcErrorResponse> {
    // Do all the JsonRPC request validatation
    return this.handleChecked(checkedRequest);
  }

  public handleChecked(request: JsonRpcRequest): Promise<JsonRpcResponse | JsonRpcErrorResponse> {
    // 1. convert JsonRpcRequest into calls to server core
    // 2. call server core
    // 3. convert result to JSON-RPC format
  }
}
```

that separates the general code from the application code in signingservice.worker.ts from #648

webmaster128:
Closes #650

### DIFF
--- a/packages/iov-core/src/jsonrpcsigningserver.spec.ts
+++ b/packages/iov-core/src/jsonrpcsigningserver.spec.ts
@@ -64,8 +64,7 @@ async function makeJsonRpcSigningServer(): Promise<JsonRpcSigningServer> {
   // ganache second identity
   await profile.createIdentity(secp256k1Wallet.id, ethereumChainId, HdPaths.bip44(60, 0, 0, 1));
 
-  const server = new JsonRpcSigningServer(new SigningServerCore(profile, signer));
-  return server;
+  return new JsonRpcSigningServer(new SigningServerCore(profile, signer));
 }
 
 describe("JsonRpcSigningServer", () => {
@@ -114,6 +113,7 @@ describe("JsonRpcSigningServer", () => {
       fromHex("533e376559fa551130e721735af5e7c9fcd8869ddd54519ee779fce5984d7898"),
     );
 
+    server.shutdown();
     bnsConnection.disconnect();
   });
 
@@ -140,6 +140,8 @@ describe("JsonRpcSigningServer", () => {
     expect(response.result).toEqual(jasmine.any(Array));
     expect((response.result as ReadonlyArray<any>).length).toEqual(1);
     expect(response.result[0]).toEqual(ganacheSecondIdentity);
+
+    server.shutdown();
   });
 
   it("can get BNS or Ethereum identities", async () => {
@@ -172,6 +174,7 @@ describe("JsonRpcSigningServer", () => {
     );
     expect(response.result[1]).toEqual(ganacheSecondIdentity);
 
+    server.shutdown();
     bnsConnection.disconnect();
   });
 
@@ -224,6 +227,7 @@ describe("JsonRpcSigningServer", () => {
     expect(transactionSearch[0].transactionId).toEqual(transactionId);
     expect(transactionSearch[0].transaction).toEqual(send);
 
+    server.shutdown();
     bnsConnection.disconnect();
   });
 });
--- a/packages/iov-core/src/jsonrpcsigningserver.spec.ts
+++ b/packages/iov-core/src/jsonrpcsigningserver.spec.ts
@@ -1,0 +1,229 @@
+import {
+  Address,
+  Algorithm,
+  Amount,
+  ChainId,
+  PublicIdentity,
+  PublicKeyBytes,
+  SendTransaction,
+  TokenTicker,
+  TransactionId,
+} from "@iov/bcp-types";
+import { bnsCodec, bnsConnector } from "@iov/bns";
+import { Ed25519, Random } from "@iov/crypto";
+import { Encoding } from "@iov/encoding";
+import { isJsonRpcErrorResponse, JsonCompatibleDictionary } from "@iov/jsonrpc";
+import { Ed25519HdWallet, HdPaths, Secp256k1HdWallet, UserProfile } from "@iov/keycontrol";
+import { toListPromise } from "@iov/stream";
+
+import { JsonRpcSigningServer } from "./jsonrpcsigningserver";
+import { MultiChainSigner } from "./multichainsigner";
+import { SigningServerCore } from "./signingservercore";
+
+const { fromHex } = Encoding;
+
+function pendingWithoutBnsd(): void {
+  if (!process.env.BNSD_ENABLED) {
+    pending("Set BNSD_ENABLED to enable bnsd-based tests");
+  }
+}
+
+function pendingWithoutEthereum(): void {
+  if (!process.env.ETHEREUM_ENABLED) {
+    pending("Set ETHEREUM_ENABLED to enable ethereum-based tests");
+  }
+}
+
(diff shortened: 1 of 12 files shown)


packages/iov-core/src/jsonrpcsigningserver.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+      id: 123,
+      method: "getIdentities",
+      params: {
+        reason: "Who are you?",
+        chainIds: [bnsConnection.chainId()],
+      },
+    });
+    expect(response.jsonrpc).toEqual("2.0");
+    expect(response.id).toEqual(123);
+    if (isJsonRpcErrorResponse(response)) {
```
ethanfrey: style q:

why not use `expect(isJsonRpcErrorResponse(response)).toEqual(true)` here?

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+function parseRpcCall(data: JsonRpcRequest): RpcCall {
+  if (!isJsonCompatibleDictionary(data.params)) {
+    throw new Error("Request params are only supported as dictionary");
+  }
+
+  switch (data.method) {
+    case "getIdentities": {
+      const reason = data.params.reason;
```
ethanfrey: style: 
`const { chainIds, reason } = data.params` 
is simpler and easier to read (for me).

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,184 @@
...
+        throw new ParamsError("Found non-string element in chainIds array");
+      }
+      return {
+        name: "getIdentities",
+        reason: reason,
+        chainIds: chainIds,
+      };
+    }
+    case "signAndPost": {
+      const reason = data.params.reason;
```
ethanfrey: Same as above about destructuring params

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,184 @@
...
+      const transaction = data.params.transaction;
+      if (typeof reason !== "string") {
+        throw new ParamsError("1st parameter (reason) must be a string");
+      }
+      if (typeof transaction !== "object") {
```
ethanfrey: Not more tests than Object?
What about asserting the common fields exist at least, like kind, signer, etc....

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,184 @@
...
+  readonly reason: string;
+  readonly transaction: UnsignedTransaction;
+}
+
+type RpcCall = RpcCallGetIdentities | RpcCallSignAndPost;
+
+class ParamsError extends Error {}
+class MethodNotFoundError extends Error {}
+
+function isArrayOfStrings(array: ReadonlyArray<any>): array is ReadonlyArray<string> {
```
ethanfrey: :+1:

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,184 @@
...
+
+interface RpcCallSignAndPost {
+  readonly name: "signAndPost";
+  readonly reason: string;
+  readonly transaction: UnsignedTransaction;
+}
+
+type RpcCall = RpcCallGetIdentities | RpcCallSignAndPost;
+
+class ParamsError extends Error {}
```
ethanfrey: Custom classes to allow checks by client, right?

Fine as is, but wondering what is best practice? I have seen sometimes errors with special members eg `err.notFound` to signify a particular error type. I guess `instanceof` on a class is more modern?

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+    default:
+      throw new MethodNotFoundError("Unknown method name");
+  }
+}
+
+/**
+ * A transport-agnostic JSON-RPC wrapper around SigningServerCore
+ */
+export class JsonRpcSigningServer {
```
ethanfrey: I very much like pulling out this logic into the rpc handling aspect.

packages/iov-core/src/signingservercore.ts
```diff
@@ -60,6 +60,15 @@ export class ServerCore {
     return response.transactionId;
   }
 
+  /**
+   * Call this to free ressources when server is not needed anymore
+   */
+  public shutdown(): void {
```
ethanfrey: :+1:

packages/iov-core/src/workers/signingservice.worker.ts
```diff
@@ -1,91 +1,15 @@
 /// <reference lib="webworker" />
 
-/// A proof-of-work RPC server for the out pf process signer
+// A proof-of-work RPC server implementation for the out of process signer.
```
ethanfrey: Thanks for update

packages/iov-core/src/workers/signingservice.worker.ts
```diff
@@ -108,82 +32,7 @@ async function main(): Promise<void> {
...
-        id: request.id,
-        error: {
-          code: errorCode,
-          message: error.toString(),
-        },
-      };
-      return errorResponse;
-    }
-  }
+  const server = new JsonRpcSigningServer(new SigningServerCore(profile, signer));
```
ethanfrey: I like this example. All logic is gone, just configuration, as should be in a test case :+1:

packages/iov-core/src/jsonrpcsigningserver.spec.ts
webmaster128: response is of type `JsonRpcResponse | JsonRpcErrorResponse`. I need to throw here to allow accessing members of `JsonRpcResponse` that do not exist in `JsonRpcErrorResponse`

packages/iov-core/src/jsonrpcsigningserver.ts
webmaster128: It is used internally only. I need to convert different exceptions to different JSON-RPC error codes.

Both ways work. I choose the way you do it in all typed languages, that is unfortunatly poorly supported by JavaScript, since you cannot catch by type directly. Using properties on Error should work as long as you don't use TypeScript, which disallows adding new members. Using TypeScript you need a custom type anyway. So yes, we could also have a single custom type that holds the type information.

Since the scope of this is only a single file, we can change at any time if we have a reason to

packages/iov-core/src/jsonrpcsigningserver.ts
webmaster128: Added some more: https://github.com/iov-one/iov-core/pull/654/commits/084c472c35f4efbb720efa25c3c06e2c39448c0f

packages/iov-core/src/jsonrpcsigningserver.ts
webmaster128: Jap, you are right, it's a little nicer: https://github.com/iov-one/iov-core/pull/654/commits/2c26a417ab0e644f12e056771b12e3f5ecdd9601

PR #657, opened 2019-01-03
Add HdPaths.bip44Like and HdPaths.iov

### LINKED ISSUE #634: Add BIP44-like HD path helpers for account based coins
There are _BIP44-like_ HD paths for account based coins:

```
m/44'/<coin_type>'/<account index>'
```

following [Trezor's recommendation](https://github.com/trezor/trezor-core/tree/master/docs/coins) that

> If the coin is UTXO-based the path should have all five parts, precisely as defined in BIP-32. If it is account-based we follow Stellar's SEP-0005 - paths have only three parts 44'/c'/a'.

We call them _BIP44-like_ because they are not [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) compatible since they use 3 instead of 5 components.

Derivation paths wee need for our software

```
m/44'/234'/a'   IOV
m/44'/134'/a'   Lisk
m/44'/1120'/a'  RISE

// maybe later
m/44'/148'/a'   Stellar
```

webmaster128:
Closes #634

### DIFF
--- a/packages/iov-keycontrol/src/hdpaths.ts
+++ b/packages/iov-keycontrol/src/hdpaths.ts
@@ -35,6 +35,39 @@ export class HdPaths {
   }
 
   /**
+   * BIP44-like HD paths for account based coins
+   *
+   * This is following Trezor's recommendation that
+   * > If the coin is UTXO-based the path should have all five parts,
+   * > precisely as defined in BIP-32. If it is account-based we follow
+   * > Stellar's SEP-0005 - paths have only three parts 44'/c'/a'.
+   *
+   * Example paths to use this for
+   *
+   *     m/44'/234'/a'   IOV
+   *     m/44'/134'/a'   Lisk
+   *     m/44'/1120'/a'  RISE
+   *     m/44'/148'/a'   Stellar
+   *
+   * This is called "BIP44-like" because it follows the idea of BIP44 but is not
+   * compatible to the 5 component BIP44 standard.
+   */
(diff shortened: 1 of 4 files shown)


packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -34,6 +34,39 @@ export class HdPaths {
     ];
   }
 
+  /**
```
ethanfrey: Great comment here.
Love it when I learn context in the code reviews

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -34,6 +34,39 @@ export class HdPaths {
...
+   *
+   *     m/44'/234'/a'   IOV
+   *     m/44'/134'/a'   Lisk
+   *     m/44'/1120'/a'  RISE
+   *     m/44'/148'/a'   Stellar
+   *
+   * This is called "BIP44-like" because it follows the idea of BIP44 but is not
+   * compatible to the 5 component BIP44 standard.
+   */
+  public static bip44Like(coinType: number, account: number): ReadonlyArray<Slip10RawIndex> {
```
ethanfrey: Nice helper functions

packages/iov-keycontrol/src/hdpaths.ts
webmaster128: :) I'm glad it helps

PR #658, opened 2019-01-03
Add prefix "ethereum-eip155-" to Ethereums BCP chain IDs

### LINKED ISSUE #649: Add prefix to Ethereum chain IDs
For Ethereum we want to use BCP chain IDs in the format `ethereum-eip155-%d` where `%d` is a numeric chain ID from https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#list-of-chain-ids

webmaster128:
Closes #649

### DIFF
--- a/packages/iov-ethereum/src/utils.spec.ts
+++ b/packages/iov-ethereum/src/utils.spec.ts
@@ -1,4 +1,4 @@
-import { Nonce } from "@iov/bcp-types";
+import { ChainId, Nonce } from "@iov/bcp-types";
 import { Int53 } from "@iov/encoding";
 
 import {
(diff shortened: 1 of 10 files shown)


packages/iov-ethereum/src/utils.spec.ts
```diff
@@ -142,4 +144,55 @@ describe("Ethereum utils", () => {
       expect(hexPadToEven("0x4a817c800")).toEqual("04a817c800");
     });
   });
+
+  describe("toBcpChainId", () => {
```
ethanfrey: Nice util functions and test vectors

PR #662, opened 2019-01-03
Refactorings of @iov/jsonrpc

webmaster128:
Refactorings of @iov/jsonrpc to allow reusing it for Ethereum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 * @iov/bcp-type: `BcpConnection.getNonce` and `.watchNonce` now return
   `Promise<Nonce>`/`Stream<Nonce>` and implementations set a default value on
   their own.
+* @iov/jsonrpc: Add new package for type-safe JSON-RPC 2.0 interfaces and
+  response parsers. This is used for out-of-process signing and can be reused
+  in @iov/ethereum and @iov/tendermint-rpc later on.
 * @iov/keycontrol: Add `HdPaths.bip44Like` and `HdPaths.iov`
 
 Breaking changes
--- a/packages/iov-core/src/signingservice.spec.ts
+++ b/packages/iov-core/src/signingservice.spec.ts
@@ -20,7 +20,7 @@ import {
   JsonCompatibleDictionary,
   JsonRpcClient,
   JsonRpcErrorResponse,
-  JsonRpcResponse,
+  JsonRpcSuccessResponse,
   parseJsonRpcError,
   parseJsonRpcResponse,
   SimpleMessagingConnection,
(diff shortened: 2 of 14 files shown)


packages/iov-core/src/signingservice.spec.ts
```diff
@@ -53,6 +62,33 @@ async function randomBnsAddress(): Promise<Address> {
   return bnsCodec.identityToAddress(randomIdentity);
 }
 
+function makeSimpleMessagingConnection(worker: Worker): SimpleMessagingConnection {
```
ethanfrey: Streams, streams, everywhere :)

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * @iov/bcp-type: `BcpConnection.getNonce` and `.watchNonce` now return
   `Promise<Nonce>`/`Stream<Nonce>` and implementations set a default value on
   their own.
+* @iov/jsonrpc: Add new package for type-safe JSON-RPC 2.0 interfaces and
+  response parsers. This is used for out-of-process signing and can be reused
+  in @iov/ethereum and @iov/tendermint-rpc later on.
```
ethanfrey: Yes, I saw tendermint rpc is very similar to ethereum (probably based the design on it).
Would be great to reuse this package

packages/iov-jsonrpc/src/types.ts
```diff
@@ -11,7 +11,7 @@ export interface JsonRpcRequest {
   readonly params: JsonCompatibleArray | JsonCompatibleDictionary;
 }
 
-export interface JsonRpcResponse {
+export interface JsonRpcSuccessResponse {
   readonly jsonrpc: "2.0";
   readonly id: number;
```
ethanfrey: If you want to support tendermint, etc... id is a string
Is it defined to be a number in the jsonrpc spec?

packages/iov-jsonrpc/src/types.ts
```diff
@@ -32,9 +32,9 @@ export interface JsonRpcErrorResponse {
   readonly error: JsonRpcError;
 }
 
-export function isJsonRpcErrorResponse(
-  response: JsonRpcResponse | JsonRpcErrorResponse,
-): response is JsonRpcErrorResponse {
+export type JsonRpcResponse = JsonRpcSuccessResponse | JsonRpcErrorResponse;
```
ethanfrey: Yes, looks more like the tendermnt jsonrpc stuff

CHANGELOG.md
webmaster128: Fortunatly JSON-RPC 2.0 is a standard that both use. And as far as I can see, both Tendermint and Ethereum are spec-complient. Locally I could already use @iov/jsonrpc for Ethereum

packages/iov-jsonrpc/src/types.ts
webmaster128: Oh, seems like both are valid:

> An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [1] and Numbers SHOULD NOT contain fractional parts [2]

https://www.jsonrpc.org/specification#request_object

Will generalize when using this for tendermint

PR #664, opened 2019-01-04
Implement buffering stream concat and use in BNS

webmaster128:
This splits the logic of liveTx() into

1. Stream concatenation
2. Event deduplication

where 1. is implemented in a reusable fashion as `concat` in @iov/stream.

The last commit (https://github.com/iov-one/iov-core/pull/664/commits/0d14198a30751f609b83518d5c9b3ce5f9832780) is not strictly related to the PR topic but something I found by failing tests while working on the rest.

### DIFF
--- a/packages/iov-bns/src/bnsconnection.ts
+++ b/packages/iov-bns/src/bnsconnection.ts
@@ -40,7 +40,7 @@ import {
   TxReadCodec,
 } from "@iov/bcp-types";
 import { Encoding, Int53, Uint53 } from "@iov/encoding";
-import { concat, DefaultValueProducer, toListPromise, ValueAndUpdates } from "@iov/stream";
+import { concat, DefaultValueProducer, fromListPromise, toListPromise, ValueAndUpdates } from "@iov/stream";
 import {
   broadcastTxSyncSuccess,
   Client as TendermintClient,
@@ -467,61 +467,17 @@ export class BnsConnection implements BcpAtomicSwapConnection {
    * and then continuing with live feeds
    */
   public liveTx(txQuery: BcpTxQuery): Stream<ConfirmedTransaction> {
-    let updatesSubscription: Subscription | undefined;
+    const historyStream = fromListPromise(this.searchTx(txQuery));
+    const updatesStream = this.listenTx(txQuery);
+    const combinedStream = concat(historyStream, updatesStream);
 
-    const producer: Producer<ConfirmedTransaction> = {
-      start: listener => {
-        // called as soon as anyone subscribes to the stream
+    // remove duplicates
+    const alreadySent = new Set<TransactionId>();
+    const deduplicatedStream = combinedStream
+      .filter(ct => !alreadySent.has(ct.transactionId))
+      .debug(ct => alreadySent.add(ct.transactionId));
 
-        // 1. subscribe to updates before queying history to ensure we don't miss something; store updates in queue
-        // 2. when history is there, send history and process queue
-        // 3. skip queue and send updates directly
-
-        let doneSendingHistory = false;
-        const queue = new Array<ConfirmedTransaction>();
-
-        updatesSubscription = this.listenTx(txQuery).subscribe({
-          next: value => {
-            if (doneSendingHistory) {
-              listener.next(value);
-            } else {
-              queue.push(value);
-            }
-          },
-          error: error => listener.error(error),
-        });
-
-        this.searchTx(txQuery)
-          .then(history => {
-            for (const transaction of history) {
-              listener.next(transaction);
-            }
-
-            const historyIds = history.map(transaction => transaction.transactionId);
-
-            let element: ConfirmedTransaction | undefined;
-            // tslint:disable-next-line:no-conditional-assignment
-            while ((element = queue.shift())) {
-              const elementId = element.transactionId;
-              if (historyIds.indexOf(elementId) !== -1) {
-                // only do this for elements not already sent
-                listener.next(element);
-              }
-            }
-
-            doneSendingHistory = true;
-          })
-          .catch(error => listener.error(error));
-      },
-      stop: () => {
-        if (!updatesSubscription) {
-          throw new Error("stop() was clled before start(). This is not expected to happen.");
-        }
-        updatesSubscription.unsubscribe();
-      },
-    };
-
-    return Stream.create(producer);
+    return deduplicatedStream;
   }
 
   /**
--- a/packages/iov-bns/src/bnsconnection.ts
+++ b/packages/iov-bns/src/bnsconnection.ts
@@ -40,7 +40,7 @@ import {
   TxReadCodec,
 } from "@iov/bcp-types";
 import { Encoding, Int53, Uint53 } from "@iov/encoding";
-import { DefaultValueProducer, toListPromise, ValueAndUpdates } from "@iov/stream";
+import { concat, DefaultValueProducer, toListPromise, ValueAndUpdates } from "@iov/stream";
 import {
   broadcastTxSyncSuccess,
   Client as TendermintClient,
@@ -638,7 +638,7 @@ export class BnsConnection implements BcpAtomicSwapConnection {
       return acct.data.length < 1 ? undefined : acct.data[0];
     };
 
-    return Stream.merge(
+    return concat(
       Stream.fromPromise(oneAccount()),
       this.changeBalance(account.address)
         .map(() => Stream.fromPromise(oneAccount()))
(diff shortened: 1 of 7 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -467,61 +475,17 @@ export class BnsConnection implements BcpAtomicSwapConnection {
    * and then continuing with live feeds
    */
   public liveTx(txQuery: BcpTxQuery): Stream<ConfirmedTransaction> {
-    let updatesSubscription: Subscription | undefined;
+    const historyStream = fromListPromise(this.searchTx(txQuery));
+    const updatesStream = this.listenTx(txQuery);
+    const combinedStream = concat(historyStream, updatesStream);
```
ethanfrey: nice

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -467,61 +475,17 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-            while ((element = queue.shift())) {
-              const elementId = element.transactionId;
-              if (historyIds.indexOf(elementId) !== -1) {
-                // only do this for elements not already sent
-                listener.next(element);
-              }
-            }
+    // remove duplicates
+    const alreadySent = new Set<TransactionId>();
+    const deduplicatedStream = combinedStream
```
ethanfrey: I like this simplified logic quite a bit.
I don't quite understand what debug does, but it seems like it runs that code for each item and then passes the original item.

`debug(ct => alreadySent.add(ct.transactionId))` 
would be the same as
`filter(ct => {alreadySent.add(ct.transactionId); return true})`
but easier to read. Correct?

ethanfrey: So much code gone with this refactor. Like it!

packages/iov-stream/src/concat.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+      next: value => expect(value).toEqual(expected.shift()!),
+      complete: () => {
+        expect(expected.length).toEqual(0);
+        done();
+      },
+      error: done.fail,
+    });
+  });
+
+  it("changes output order when order of streams switch", done => {
```
ethanfrey: this test case is duplicated above.
either remove it or make it unique

packages/iov-stream/src/concat.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+      next: value => expect(value).toEqual(expected.shift()!),
+      complete: () => {
+        expect(expected.length).toEqual(0);
+        done();
+      },
+      error: done.fail,
+    });
+  });
+
+  it("should append a synchronous stream after an asynchronous stream", done => {
```
ethanfrey: Very nice case

packages/iov-stream/src/concat.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+      next: value => expect(value).toEqual(expected.shift()!),
+      complete: () => {
+        expect(expected.length).toEqual(0);
+        done();
+      },
+      error: done.fail,
+    });
+  });
+
+  it("unsubscribes and re-subscribes from source streams", done => {
```
ethanfrey: Very nice test case...
I just started wondering about that, would producers be registered properly.
Actually, my thought was more like `const combined = concat(stream1, stream2)`, where stream1 is a periodic one (like above), and stream 2 is a producer, that will only produce events when a listener is active.

So...
In the case we listen to stream 1 until done, then listen to stream2, some event will be missing.

Like a bit more complex take on this test.
That events are triggered at certain times and if no listener is registered at time X, then the event gets dropped.

packages/iov-stream/src/concat.ts
```diff
@@ -0,0 +1,102 @@
...
+  let activeStreamIndex = 0;
+
+  function reset(): void {
+    while (subscriptions.length > 0) {
+      const subscription = subscriptions.shift()!;
+      subscription.unsubscribe();
+    }
+
+    // tslint:disable-next-line:no-object-mutation
+    queues.length = 0;
```
ethanfrey: Huh? Setting the length empties the array?
Why not just reset it to a new Array (and use let above). Seems cleaner

packages/iov-stream/src/concat.ts
```diff
@@ -0,0 +1,102 @@
...
+      function isDone(): boolean {
+        return activeStreamIndex >= streams.length;
+      }
+
+      if (isDone()) {
+        listener.complete();
+        return;
+      }
+
+      streams.forEach((stream, index) => {
```
ethanfrey: Okay, from inspection, I think the code is correct (wrt the test case I mentioned above).
You do subscribe to all streams at once and add them to queues to be delivered later

packages/iov-stream/src/concat.ts
```diff
@@ -0,0 +1,102 @@
...
+        subscriptions.push(
+          stream.subscribe({
+            next: value => {
```
ethanfrey: :+1:

packages/iov-stream/src/concat.ts
```diff
@@ -0,0 +1,102 @@
...
+              if (index === activeStreamIndex) {
+                listener.next(value);
+              } else {
+                queues[index].push(value);
+              }
+            },
+            complete: () => {
+              completedStreams.add(index);
+
+              while (completedStreams.has(activeStreamIndex)) {
```
ethanfrey: :+1:

packages/iov-stream/src/concat.ts
```diff
@@ -0,0 +1,102 @@
...
+                // this stream completed: emit all and move on
+                emitAllQueuesEvents(activeStreamIndex);
+                activeStreamIndex++;
+              }
+
+              if (isDone()) {
+                listener.complete();
+              } else {
+                // now active stream can have some events queued but did not yet complete
+                emitAllQueuesEvents(activeStreamIndex);
```
ethanfrey: :+1:

packages/iov-stream/src/concat.ts
```diff
@@ -0,0 +1,102 @@
...
+              }
+            },
+            error: error => {
```
ethanfrey: So, error on any stream aborts the entire combined stream.
That seems like very proper behavior

packages/iov-stream/src/concat.spec.ts
webmaster128: Oh, indeed a copy/paste error. Thanks!

packages/iov-stream/src/concat.ts
webmaster128: This is the only way I am aware of to clear an Array object. It is dirty as hell compared to a `clear()` method like you have for [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)s or [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)s.

But in this case I don't want to create another object (by assigning a new array), which makes it harder to track which object is used in which code block.

packages/iov-bns/src/bnsconnection.ts
webmaster128: debug() simply runs a function for every event, without changing the stream. It was created for debugging purposes and has some simpler ways to call it:

* `.debug()` logs the event to console
* `.debug("label123")` logs the event to console, adding the debug label "label123"

For out use case `.debug(function_call)` would better be called `.forEarch(function_call)`.

My mid-term idea is to create a deduplication helper in @iov/stream and push the debug() call down the stack.

packages/iov-stream/src/concat.spec.ts
webmaster128: What you describe is the test case that differentiates xstream's concat and out concat: https://github.com/staltz/xstream/issues/170#issuecomment-451305237

You can do the same thing with two streams that are both powered by a shared source.

```ts
    const sourceStream = Stream.periodic(25);
    const stream1 = sourceStream.take(3);
    const stream2 = sourceStream.take(3);
    const concatenatedStream = concat(stream1, stream2);
    const expected = [0, 1, 2, 0, 1, 2];
```

If subscription time to `stream2` was not correct, you would get other values than `[0, 1, 2]`.

This is implemented as test "buffers asynchonous events of second stream until first stream completes"

packages/iov-stream/src/concat.ts
webmaster128: I just checked again: this is what Stream.merge does as well: https://github.com/staltz/xstream/blob/master/src/index.ts#L246-L249

As far as I understood, any streams ends with completion or error and now two error events can be emitted on a stream. So as soon as concat emits an error, there is nothing more to be done.

ethanfrey: I agree with it. Saying I think this is proper solution

packages/iov-stream/src/concat.spec.ts
ethanfrey: Yes. Great example

packages/iov-stream/src/concat.ts
ethanfrey: Fair enough

PR #666, opened 2019-01-08
Fix calculation of confirmations in Ethereum searchTx

webmaster128:


packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -374,7 +374,7 @@ export class EthereumConnection implements BcpConnection {
 
       const currentHeight = decodeHexQuantity(lastBlockNumberResponse.result);
 
-      const confirmations = currentHeight - transactionHeight;
+      const confirmations = currentHeight - transactionHeight + 1;
```
ethanfrey: Yes, good call. In a block is considered 1 confirmation

PR #685, opened 2019-01-11
Migrate UnsignedTransaction.chainId and .signer into .creator 

### LINKED ISSUE #640: Migrate transaction .chainId and .signer into .creator
Convert `UnsignedTransaction.chainId` + `UnsignedTransaction.signer` into

```ts
interface UnsignedTransaction {
  creator: PublicIdentity;
  // ...
}
```

This consistes of two aspects:

* Merging `chainId` and `signer` into a combined object of type `PublicIdentity`
* Use the term "creator" as a better name when you think of multi signatures

### LINKED ISSUE #670: Expose publicKeyBundleEquals
To avoid this in caller code:

```
const keysEqual = (a: PublicKeyBundle, b: PublicKeyBundle): boolean =>
  a.algo === b.algo && arraysEqual(a.data, b.data);
const arraysEqual = (a: Uint8Array, b: Uint8Array): boolean => {
  if (a.length !== b.length) {
    return false;
  }
  for (let i = 0; i < a.length; i++) {
    if (a[i] !== b[i]) {
      return false;
    }
  }
  return true;
};
```

We need this functionality in `publicIdentityEquals` anyway

webmaster128:
Closes #670
Closes #640


ethanfrey:
Why is this on hold?

webmaster128:
Depends on #693

PR #687, opened 2019-01-11
Add publicKeyBundleEquals

### LINKED ISSUE #670: Expose publicKeyBundleEquals

webmaster128:
To be resused by callers projects

Closes #670

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * @iov/bcp-type: `BcpConnection.getNonce` now returns a `Promise<Nonce>` and
   implementations set a default value on their own.
+* @iov/bcp-type: Expose `publicKeyBundleEquals`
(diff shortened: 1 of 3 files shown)


webmaster128:
Because there is no testing infrastructure at all in bcp-types ;) bcp-types is just slowly transitioning from types-only to types+helpers and will get probably get testing one day or another.

PR #692, opened 2019-01-15
Avoid extra query in block header events

webmaster128:
This was creating some annoying race condition test fails in the wallet:

The execution order is:
- header `n` is produced
- missing information of header `n` is requested
- full BlockHeader `n` event arrives in test code
- test code unsubscribes
- test code disconnects
- socket is disconnected
- header `n + 1` is produced
- missing information of header `n + 1`  is requested
- Boooom! Trying to send on disconnected WebSocket

### DIFF
--- a/packages/iov-bns/src/bnsconnection.ts
+++ b/packages/iov-bns/src/bnsconnection.ts
@@ -1,5 +1,5 @@
 import equal from "fast-deep-equal";
-import { Producer, Stream, Subscription } from "xstream";
+import { Stream, Subscription } from "xstream";
 
 import {
   Address,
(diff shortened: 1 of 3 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -558,28 +558,20 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-      stop: () => {
-        headersSubscription.unsubscribe();
-      },
-    };
-    return Stream.create(producer);
+    // TODO: ID unavailable because
+    // - we cannot trivially calculate it https://github.com/iov-one/iov-core/issues/618 and
+    // - want to avoid an extra query to the node which causes issues when trying to send to a disconnected socket
+    // Leave it empty as long as no application strictly requires the ID
+    const dummyBlockId = "" as BlockId;
```
ethanfrey: I would make it clear it is dummy.
If a caller tries it and it is empty, they will wonder
If they try and it say `Id not yet implemented` or such, it will be clear

webmaster128: jap, good idea, done

webmaster128:
>  Also, no changes needed in ethereum code?

We need to be a bit careful for Ethereum where the same could happen at some point. But right now, all non-subscription RPC calls are done via HTTP, and HTTP call after disconnected Websocket is no problem.

PR #694, opened 2019-01-16
Remove LocalIdentity/LocalIdentityId

### LINKED ISSUE #693: Separate LocalIdentity and PublicIdentity
Right now, every LocalIdentity is a PublicIdentity, making it type compatible. This means you can assign a LocalIdentity to all the places where a PublicIdentity is expected. However, at runtime the extra information of LocalIdentity memain. This makes it very easy to leak private information accidentally by copying a LocalIdentity into public space. This happened twice during IOV development.

To avoid this kind of potential security issue, we should not derive sensitive types from public types.

webmaster128:
Closes #693

### DIFF
--- a/packages/iov-keycontrol/src/wallets/slip10wallet.ts
+++ b/packages/iov-keycontrol/src/wallets/slip10wallet.ts
@@ -1,4 +1,5 @@
 import PseudoRandom from "random-js";
+import { As } from "type-tagger";
 
 import {
   Algorithm,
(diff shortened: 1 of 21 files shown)


packages/iov-keycontrol/src/wallets/slip10wallet.ts
```diff
@@ -171,12 +167,16 @@ export class Slip10Wallet implements Wallet {
...
   private readonly secret: EnglishMnemonic;
   private readonly curve: Slip10Curve;
-  private readonly identities: LocalIdentity[];
-  private readonly privkeyPaths: Map<string, ReadonlyArray<Slip10RawIndex>>;
   private readonly labelProducer: DefaultValueProducer<string | undefined>;
 
+  // identities
+  private readonly identities: PublicIdentity[];
+  private readonly privkeyPaths: Map<IdentityId, ReadonlyArray<Slip10RawIndex>>;
+  private readonly labels: Map<IdentityId, string | undefined>;
```
ethanfrey: This repeats the Map mostly...
Maybe something like 
```ts
private readonly idData: Map<IdentityId, IdentityData>

interface IdentityData {
  readonly paths: ReadonlyArray<Slip10RawIndex>;
  readonly label: string | undefined;
}
```

I think this will make it easier to add more in the future.
But maybe not worth the change... fine as it is, you can take or leave the suggestion

webmaster128: Good idea. Will update as soon as we have 3 properties per IdentityId

PR #700, opened 2019-01-16
Add BcpConnection.getNonces 

### LINKED ISSUE #669: Add Bcp.getNonces
`.getNonces(query, count)` should return a list of nonces to sign the next `count` transactions. This becomes very valuable for Ethereum, where `.getNonce()` can only be used a second time as soon as the first nonce is in a block.

This code from #668 shows how this is needed:

```ts
        const nonceA = await connection.getNonce({ pubkey: sender.pubkey });
        const nonceB = new Int53(nonceA.toNumber() + 1) as Nonce;
        const nonceC = new Int53(nonceA.toNumber() + 2) as Nonce;

        const signedA = await profile.signTransaction(wallet.id, sender, sendA, ethereumCodec, nonceA);
        const signedB = await profile.signTransaction(wallet.id, sender, sendB, ethereumCodec, nonceB);
        const signedC = await profile.signTransaction(wallet.id, sender, sendC, ethereumCodec, nonceC);
        const bytesToPostA = ethereumCodec.bytesToPost(signedA);
        const bytesToPostB = ethereumCodec.bytesToPost(signedB);
        const bytesToPostC = ethereumCodec.bytesToPost(signedC);

        // Post A and B
        const postResultA = await connection.postTx(bytesToPostA);
        await connection.postTx(bytesToPostB);

        // Wait for a block
        await postResultA.blockInfo.waitFor(info => info.state === BcpTransactionState.InBlock);

        // Post C
        await connection.postTx(bytesToPostC);
```

The caller should not need to care about how a Nonce is designed internally. It may not be incremented.

This code should look like

```ts
const [nonceA, noncB, nonceC] = await connection.getNonces({ pubkey: sender.pubkey }, 3);
```

webmaster128:
For Ethereum!

Closes #669

### DIFF
--- a/packages/iov-ethereum/src/ethereumconnection.spec.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.spec.ts
@@ -248,13 +248,15 @@ describe("EthereumConnection", () => {
       {
         const nonces = await connection.getNonces({ address: testConfig.address }, 1);
         expect(nonces.length).toEqual(1);
+        expect(nonces[0].toNumber()).toEqual(testConfig.expectedNonce.toNumber());
       }
 
       // by address, 2 nonces
       {
         const nonces = await connection.getNonces({ address: testConfig.address }, 2);
         expect(nonces.length).toEqual(2);
-        expect(nonces[1].toNumber()).toEqual(nonces[0].toNumber() + 1);
+        expect(nonces[0].toNumber()).toEqual(testConfig.expectedNonce.toNumber());
+        expect(nonces[1].toNumber()).toEqual(testConfig.expectedNonce.toNumber() + 1);
       }
 
       // by pubkey, 0 nonces
@@ -267,13 +269,15 @@ describe("EthereumConnection", () => {
       {
         const nonces = await connection.getNonces({ pubkey: testConfig.pubkey }, 1);
         expect(nonces.length).toEqual(1);
+        expect(nonces[0].toNumber()).toEqual(testConfig.expectedNonce.toNumber());
       }
 
       // by pubkey, 2 nonces
       {
         const nonces = await connection.getNonces({ pubkey: testConfig.pubkey }, 2);
         expect(nonces.length).toEqual(2);
-        expect(nonces[1].toNumber()).toEqual(nonces[0].toNumber() + 1);
+        expect(nonces[0].toNumber()).toEqual(testConfig.expectedNonce.toNumber());
+        expect(nonces[1].toNumber()).toEqual(testConfig.expectedNonce.toNumber() + 1);
       }
 
       connection.disconnect();
--- a/packages/iov-ethereum/src/ethereumconnection.spec.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.spec.ts
@@ -221,13 +221,13 @@ describe("EthereumConnection", () => {
       // by address
       {
         const nonce = await connection.getNonce({ address: testConfig.address });
-        expect(nonce).toEqual(testConfig.nonce);
+        expect(nonce).toEqual(testConfig.expectedNonce);
       }
 
       // by pubkey
       {
         const nonce = await connection.getNonce({ pubkey: testConfig.pubkey });
-        expect(nonce).toEqual(testConfig.nonce);
+        expect(nonce).toEqual(testConfig.expectedNonce);
       }
       connection.disconnect();
     });
--- a/packages/iov-ethereum/src/ethereumconnection.spec.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.spec.ts
@@ -234,6 +234,53 @@ describe("EthereumConnection", () => {
     });
   });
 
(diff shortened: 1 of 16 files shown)


packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -215,22 +214,71 @@ describe("EthereumConnection", () => {
...
+      {
+        const nonces = await connection.getNonces({ address: testConfig.address }, 1);
+        expect(nonces.length).toEqual(1);
+      }
+
+      // by address, 2 nonces
+      {
+        const nonces = await connection.getNonces({ address: testConfig.address }, 2);
+        expect(nonces.length).toEqual(2);
+        expect(nonces[1].toNumber()).toEqual(nonces[0].toNumber() + 1);
```
ethanfrey: How about checking like above, not just difference...
eg. 
```ts
expect(nonces[0]).toEqual(testConfig.nonce)
expect(nonces[1].toNumber()).toEqual(testConfig.nonce,toNumber() + 1)
```

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -792,9 +838,7 @@ describe("EthereumConnection", () => {
           memo: `liveTx() test C ${Math.random()}`,
         };
 
-        const nonceA = await connection.getNonce({ pubkey: sender.pubkey });
-        const nonceB = new Int53(nonceA.toNumber() + 1) as Nonce;
```
ethanfrey: And here I see the inspiration

packages/iov-lisk/src/liskconnection.ts
```diff
@@ -206,8 +206,13 @@ export class LiskConnection implements BcpConnection {
...
 
-  public getNonce(_: BcpAddressQuery | BcpPubkeyQuery): Promise<Nonce> {
-    return Promise.resolve(generateNonce());
+  public async getNonce(_: BcpAddressQuery | BcpPubkeyQuery): Promise<Nonce> {
+    return generateNonce();
+  }
+
+  public async getNonces(_: BcpAddressQuery | BcpPubkeyQuery, count: number): Promise<ReadonlyArray<Nonce>> {
+    const checkedCount = new Uint53(count).toNumber();
+    return Array.from({ length: checkedCount }).map(_1 => generateNonce());
```
ethanfrey: Is there an issue with using the same nonce in multiple transactions?
Like should one insert a second offset between each one?
It seems this will generate the same value multiple times

packages/iov-rise/src/riseconnection.spec.ts
```diff
@@ -182,33 +182,91 @@ describe("RiseConnection", () => {
...
+      {
+        const nonces = await connection.getNonces(pubkeyQuery, 1);
+        expect(nonces.length).toEqual(1);
+      }
+
+      // by pubkey, 2 nonces
+      {
+        const nonces = await connection.getNonces(pubkeyQuery, 2);
+        expect(nonces.length).toEqual(2);
+        expect(nonces[1].toNumber()).toEqual(nonces[0].toNumber());
```
ethanfrey: Yup, same number.
If I want to send the same person the same amount in two separate transactions in the same block, this will fail.

Maybe super edge case, I still like a minor change on each returned nonce

packages/iov-lisk/src/liskconnection.ts
webmaster128: Indeed, same transaction content in the same second will lead to the same transaction ID. I actually had this issue when working with the Lisk faucet.

Incrementing the nonce is also not optimal in this case it changes the creation timestamp (which has no real value anyway as you can put anything in there). However, Lisk nodes do not accept timestamps in the future. But it can be [now-3, now-2, now-1, now] for 4 nonces.

ethanfrey: > But it can be [now-3, now-2, now-1, now] for 4 nonces.

Yes, I prefer something like that to avoid hard to debug errors further down the line

webmaster128:
Rebased, feedback not yet implemented. Will do soon.

packages/iov-lisk/src/liskconnection.ts
webmaster128: This feature makes `getNonces(query, 2)` more advanced than `getNonce(); gerNonce()` because the later gives you two equal nonces

packages/iov-ethereum/src/ethereumconnection.spec.ts
webmaster128: done

packages/iov-lisk/src/liskconnection.ts
webmaster128: done

packages/iov-rise/src/riseconnection.spec.ts
webmaster128: done, thanks

PR #709, opened 2019-01-22
Add MultiChainSigner.isValidAddress

### LINKED ISSUE #703: Add isValidAddress to MultiChainSigner
In wallet-demo we need to have a find to validate addresses when sending tokens, so the proposed method was:

```MultiChainSigner.isValidAddress(chainId: ChainId, address: string)```

webmaster128:
Closes #703


webmaster128:
> If you want to expose a getCodec(chainid) on MultichainSigner, feel free.

Right now I try to avoid it and I think we're doing good. I think of this as follows: either you want to use IOV-Core high level, then MultiChainSigner is the way to do it. Or you use IOV-Core for very specific tasks (like [this HD discovery example](https://github.com/iov-one/iov-core/tree/master/packages/iov-lisk#address-discovery)), in which case you should are able to access the components from the sub-packages directly.

PR #710, opened 2019-01-22
Add BcpTxQuery.address in favour of package-specific address tags

webmaster128:
To address

> I also agree that we need to rethink how we do the "query by account" type api

and to prepare transaction search by ID


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -812,7 +811,7 @@ describe("BnsConnection", () => {
       await tendermintSearchIndexUpdated();
 
       // finds transaction using tag
-      const results = await connection.searchTx({ tags: [bnsFromOrToTag(rcptAddress)] });
+      const results = await connection.searchTx({ address: rcptAddress });
```
ethanfrey: Definitely cleaner API usage

packages/iov-bcp-types/src/connection.ts
```diff
@@ -135,6 +135,8 @@ export interface BcpQueryTag {
 
 export interface BcpTxQuery {
   readonly id?: TransactionId;
+  /** transaction from or to this address */
```
ethanfrey: This is a bit of a question. "any transaction" is quite hard

"send transaction to or from this address" is current implementation of `fromOrToTag`
"atomic swap address with this as sender or recipient" would be other tag
"all transactions signed by this address/pubkey" would be another tag
etc.

I would update the comment to `"send transaction to or from this address"` and maybe modify the name to match

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -521,14 +520,14 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-    return this.changeTx([bnsFromOrToTag(addr)]);
+    return this.changeTx({ address: addr });
   }
 
   /**
    * A helper that triggers if the nonce every changes
    */
   public changeNonce(addr: Address): Stream<number> {
-    return this.changeTx([bnsNonceTag(addr)]);
+    return this.changeTx({ tags: [bnsNonceTag(addr)] });
```
ethanfrey: what about adding a nonce tag? Or maybe a `signedBy: Address` field to the TxQuery, which this really is trying to do

packages/iov-bns/src/util.ts
```diff
@@ -116,7 +116,15 @@ export function isConfirmedWithSwapClaimOrTimeoutTransaction(
...
+  const key = Encoding.toHex(id).toUpperCase();
+  const value = "s"; // "s" for "set"
+  return `${key}='${value}'`;
+}
+
 export function buildTxQuery(query: BcpTxQuery): QueryString {
+  const addressConponents = query.address !== undefined ? [bnsFromOrToTag(query.address)] : [];
   const tagComponents = query.tags !== undefined ? query.tags.map(tag => `${tag.key}='${tag.value}'`) : [];
   // In Tendermint, hash can be lower case for search queries but must be upper case for subscribe queries
   const hashComponents = query.id !== undefined ? [`tx.hash='${query.id}'`] : [];
```
ethanfrey: Reading this code and remembering the bug, I am surprised that you do not auto-uppercase the tx.hash to avoid those annoying issues

webmaster128: This is because the BCP representation (type `TransactionId`) of a Tendermint transaction hash is uppercase hex. non-uppercase hex is corrupted data. When there is corrupted data, we should never hide the corruption by auto-correction but instead find the root of the corruption.

We cannot allow mixed hex case in `TransactionId` as we're comparing BCP `TransactionId` by string comparison and `'aabb' != 'AABB'`.

packages/iov-bcp-types/src/connection.ts
webmaster128: Thanks for bringing this up. I missed that since the 3 other blockchain implementations do more or less only support sent transactions. Will adapt. I think the term "address" is not adding value as it just describes the type of the identity, not the relation of the identity to the transaction. I like:

```
readonly sentFromOrTo?: Address;
readonly signedBy?: Address;
```

packages/iov-bns/src/util.ts
ethanfrey: Okay, so this is handled above, and any errors here mean mis-use of the api (force asserting garbage as TransactionId). Fair enough

packages/iov-bcp-types/src/connection.ts
ethanfrey: :+1:
That is nice.

Even if `signedBy` is only supported by bns right now, it should actually be doable in most chains, and is a useful thing to query

packages/iov-bns/src/util.ts
webmaster128: The only thing we could do here is check the format of `TransactionId` and error out if broken. However, since `TransactionId` always comes from up, I want to avoid checking it in every single API unless necessary.

packages/iov-bns/src/bnsconnection.ts
webmaster128: Postponed this for now: https://github.com/iov-one/iov-core/issues/714

packages/iov-bcp-types/src/connection.ts
webmaster128: Renamed: https://github.com/iov-one/iov-core/pull/710/commits/dc6ba49cd43dc9ab67e9dea8417fc374e999bea5

PR #713, opened 2019-01-23
Let searchTx/listenTx/liveTx handle failed transactions

### LINKED ISSUE #677: Waiting on committed transaction hangs forever on error
I use this on the result of `signAndPost`. It errors when the call errors, and resolved when the delivertx works, but when the delivertx errors, it never ends... no error is thrown.

Not sure if this an issue in ValueForUpdate.waitFor or that the updates on PostTx don't contain errors.

```ts
// this waits for one commit to be writen, then returns the response
export async function waitForCommit(req: Promise<PostTxResponse>): Promise<PostTxResponse> {
  const res = await req;
  await res.blockInfo.waitFor(info => info.state === BcpTransactionState.InBlock);
  return res;
}
```

### LINKED ISSUE #711: Move transaction ID search from searchTx/listenTx/liveTx into waitForTransaction
- [ ] Add `waitForTransaction`
- [ ] Remove `BcpTxQuery.id`

webmaster128:
~Based on #710~

Closes #677 
Closes #711

### DIFF
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -148,6 +148,8 @@ export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTr
 }
 
 export interface FailedTransaction {
+  /** a unique identifier (hash of the transaction) */
+  readonly transactionId: TransactionId;
   /**
    * Application specific error code
    */
@@ -276,22 +278,14 @@ export interface BcpConnection {
 
   // transactions
   readonly postTx: (tx: PostableBytes) => Promise<PostTxResponse>;
-  // TODO: make non-toptional
-  /**
-   * Looks up transaction in history and if not found, waits until it is available.
-   *
-   * As a consequence, this sends an event for non-existing transaction IDs. Stream completes
-   * after 1 element was sent.
-   */
-  readonly waitForTransaction?: (id: TransactionId) => Stream<ConfirmedTransaction | FailedTransaction>;
   readonly searchTx: (query: BcpTxQuery) => Promise<ReadonlyArray<ConfirmedTransaction | FailedTransaction>>;
   /**
    * Subscribes to all newly added transactions that match the query
    */
-  readonly listenTx: (query: BcpTxQuery) => Stream<ConfirmedTransaction>;
+  readonly listenTx: (query: BcpTxQuery) => Stream<ConfirmedTransaction | FailedTransaction>;
   /**
    * Returns a stream for all historical transactions that match
    * the query, along with all new transactions arriving from listenTx
    */
-  readonly liveTx: (txQuery: BcpTxQuery) => Stream<ConfirmedTransaction>;
+  readonly liveTx: (txQuery: BcpTxQuery) => Stream<ConfirmedTransaction | FailedTransaction>;
 }
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -82,19 +82,21 @@ export interface BcpTicker {
   readonly fractionalDigits: number;
 }
 
-export enum BcpTransactionState {
+export enum TransactionState {
   /** accepted by a blockchain node and in mempool */
   Pending,
   /** successfully written in a block, but cannot yet guarantee it won't be reverted */
-  InBlock,
+  Succeeded,
+  /** executing the transaction failed */
+  Failed,
 }
 
-export interface BcpBlockInfoPending {
-  readonly state: BcpTransactionState.Pending;
+export interface BlockInfoPending {
+  readonly state: TransactionState.Pending;
 }
 
-export interface BcpBlockInfoInBlock {
-  readonly state: BcpTransactionState.InBlock;
+export interface BlockInfoSucceeded {
+  readonly state: TransactionState.Succeeded;
   /** block height, if the transaction is included in a block */
   readonly height: number;
   /** depth of the transaction's block, starting at 1 as soon as transaction is in a block */
@@ -103,12 +105,25 @@ export interface BcpBlockInfoInBlock {
   readonly result?: Uint8Array;
 }
 
+export interface BlockInfoFailed {
+  readonly state: TransactionState.Failed;
+  /**
+   * Application specific error code
+   */
+  readonly code: number;
+  /**
+   * Application specific logging output in an arbitrary text format that
+   * may change at any time.
+   */
+  readonly log?: string;
+}
+
 /** Information attached to a signature about its state in a block */
-export type BcpBlockInfo = BcpBlockInfoPending | BcpBlockInfoInBlock;
+export type BlockInfo = BlockInfoPending | BlockInfoSucceeded | BlockInfoFailed;
 
 export interface PostTxResponse {
   /** Information about the block the transaction is in */
-  readonly blockInfo: ValueAndUpdates<BcpBlockInfo>;
+  readonly blockInfo: ValueAndUpdates<BlockInfo>;
   /** a unique identifier (hash of the transaction) */
   readonly transactionId: TransactionId;
   /** a human readable debugging log */
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -107,6 +107,8 @@ export interface BlockInfoSucceeded {
 
 export interface BlockInfoFailed {
   readonly state: TransactionState.Failed;
+  /** height of the block that contains the transaction */
+  readonly height: number;
   /**
    * Application specific error code
    */
@@ -148,6 +150,8 @@ export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTr
 }
 
 export interface FailedTransaction {
+  /** height of the block that contains the transaction */
+  readonly height: number;
   /** a unique identifier (hash of the transaction) */
   readonly transactionId: TransactionId;
   /**
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -284,7 +284,7 @@ export interface BcpConnection {
    * after 1 element was sent.
    */
   readonly waitForTransaction?: (id: TransactionId) => Stream<ConfirmedTransaction | FailedTransaction>;
-  readonly searchTx: (query: BcpTxQuery) => Promise<ReadonlyArray<ConfirmedTransaction>>;
+  readonly searchTx: (query: BcpTxQuery) => Promise<ReadonlyArray<ConfirmedTransaction | FailedTransaction>>;
   /**
    * Subscribes to all newly added transactions that match the query
    */
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -112,10 +112,10 @@ export interface BlockInfoFailed {
    */
   readonly code: number;
   /**
-   * Application specific logging output in an arbitrary text format that
-   * may change at any time.
+   * Application specific, human-readable, non-localized error message
+   * in an arbitrary text format that may change at any time.
    */
-  readonly log?: string;
+  readonly message?: string;
 }
 
 /** Information attached to a signature about its state in a block */
@@ -155,10 +155,10 @@ export interface FailedTransaction {
    */
   readonly code: number;
   /**
-   * Application specific logging output in an arbitrary text format that
-   * may change at any time.
+   * Application specific, human-readable, non-localized error message
+   * in an arbitrary text format that may change at any time.
    */
-  readonly log?: string;
+  readonly message?: string;
 }
 
 export function isConfirmedTransaction(
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -123,6 +123,18 @@ export interface BlockInfoFailed {
 /** Information attached to a signature about its state in a block */
 export type BlockInfo = BlockInfoPending | BlockInfoSucceeded | BlockInfoFailed;
 
+export function isBlockInfoPending(info: BlockInfo): info is BlockInfoPending {
+  return info.state === TransactionState.Pending;
+}
+
+export function isBlockInfoSucceeded(info: BlockInfo): info is BlockInfoSucceeded {
+  return info.state === TransactionState.Succeeded;
+}
+
+export function isBlockInfoFailed(info: BlockInfo): info is BlockInfoFailed {
+  return info.state === TransactionState.Failed;
+}
+
 export interface PostTxResponse {
   /** Information about the block the transaction is in */
   readonly blockInfo: ValueAndUpdates<BlockInfo>;
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -124,6 +124,10 @@ export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTr
   readonly transactionId: TransactionId;
   /** application specific data from executing tx (result, code, tags...) */
   readonly result?: Uint8Array;
+  /**
+   * Application specific logging output in an arbitrary text format that
+   * may change at any time.
+   */
   readonly log?: string;
   // readonly tags: ReadonlyArray<Tag>;
 }
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -280,9 +280,10 @@ export interface BcpConnection {
   /**
    * Looks up transaction in history and if not found, waits until it is available.
    *
-   * As a consequence, this never resolves for non-existing transaction IDs.
+   * As a consequence, this sends an event for non-existing transaction IDs. Stream completes
+   * after 1 element was sent.
    */
-  readonly waitForTransaction?: (id: TransactionId) => Promise<ConfirmedTransaction | FailedTransaction>;
+  readonly waitForTransaction?: (id: TransactionId) => Stream<ConfirmedTransaction | FailedTransaction>;
   readonly searchTx: (query: BcpTxQuery) => Promise<ReadonlyArray<ConfirmedTransaction>>;
   /**
    * Subscribes to all newly added transactions that match the query
--- a/packages/iov-bcp-types/src/connection.ts
+++ b/packages/iov-bcp-types/src/connection.ts
@@ -132,6 +132,30 @@ export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTr
   // readonly tags: ReadonlyArray<Tag>;
 }
 
+export interface FailedTransaction {
+  /**
+   * Application specific error code
+   */
+  readonly code: number;
+  /**
+   * Application specific logging output in an arbitrary text format that
+   * may change at any time.
+   */
+  readonly log?: string;
+}
+
+export function isConfirmedTransaction(
+  transaction: ConfirmedTransaction | FailedTransaction,
+): transaction is ConfirmedTransaction {
+  return typeof (transaction as any).transaction !== "undefined";
+}
+
+export function isFailedTransaction(
+  transaction: ConfirmedTransaction | FailedTransaction,
+): transaction is FailedTransaction {
+  return !isConfirmedTransaction(transaction);
+}
+
 export interface BcpQueryTag {
   readonly key: string;
   readonly value: string;
@@ -235,8 +259,15 @@ export interface BcpConnection {
   /** @deprecated use watchBlockHeaders().map(header => header.height) */
   readonly changeBlock: () => Stream<number>;
 
-  // transaction
+  // transactions
   readonly postTx: (tx: PostableBytes) => Promise<PostTxResponse>;
+  // TODO: make non-toptional
+  /**
+   * Looks up transaction in history and if not found, waits until it is available.
+   *
+   * As a consequence, this never resolves for non-existing transaction IDs.
+   */
+  readonly waitForTransaction?: (id: TransactionId) => Promise<ConfirmedTransaction | FailedTransaction>;
   readonly searchTx: (query: BcpTxQuery) => Promise<ReadonlyArray<ConfirmedTransaction>>;
   /**
    * Subscribes to all newly added transactions that match the query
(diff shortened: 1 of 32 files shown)


webmaster128:
Test fail in "reports DeliverTx errors for existing transactions" is because `await response.blockInfo.waitFor(info => info.state === BcpTransactionState.InBlock);` never resolves, as its implementation does not work with DeliverTx errors .

packages/iov-bcp-types/types/connection.d.ts
```diff
@@ -154,6 +185,12 @@ export interface BcpConnection {
     /** @deprecated use watchBlockHeaders().map(header => header.height) */
     readonly changeBlock: () => Stream<number>;
     readonly postTx: (tx: PostableBytes) => Promise<PostTxResponse>;
+    /**
+     * Looks up transaction in history and if not found, waits until it is available.
+     *
+     * As a consequence, this never resolves for non-existing transaction IDs.
+     */
+    readonly waitForTransaction?: (id: TransactionId) => Promise<ConfirmedTransaction | FailedTransaction>;
```
webmaster128: This interface has one significant flaw: there is not way to cancel the waiting externally. If we use it as part of a stream producer (on postTx for example), we cannot stop this, even if noone listens to the stream anymore.

So we might be forced to convert it into a single event stream, which allows cancellation.

packages/iov-bcp-types/src/connection.ts
```diff
@@ -280,9 +280,10 @@ export interface BcpConnection {
   /**
    * Looks up transaction in history and if not found, waits until it is available.
    *
-   * As a consequence, this never resolves for non-existing transaction IDs.
+   * As a consequence, this sends an event for non-existing transaction IDs. Stream completes
+   * after 1 element was sent.
    */
-  readonly waitForTransaction?: (id: TransactionId) => Promise<ConfirmedTransaction | FailedTransaction>;
+  readonly waitForTransaction?: (id: TransactionId) => Stream<ConfirmedTransaction | FailedTransaction>;
```
ethanfrey: What's up with this change?
Seems to make it harder to use, and I don't see the upside.

webmaster128: Have a look at the 2 resolves conversation in here. The issue is you cannot cancel a Promise.

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -192,23 +196,33 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-    let lastEventSent: BcpBlockInfo = firstEvent;
-    const blockInfoProducer = new DefaultValueProducer<BcpBlockInfo>(firstEvent, {
+    const firstEvent: BlockInfo = { state: TransactionState.Pending };
+    let lastEventSent: BlockInfo = firstEvent;
+    const blockInfoProducer = new DefaultValueProducer<BlockInfo>(firstEvent, {
       onStarted: async () => {
         try {
-          // we utilize liveTx to implement a _search or watch_ mechanism since we do not know
-          // if the transaction is already committed when the producer is started
-          const searchResult = await toListPromise(this.liveTx({ id: transactionId }), 1);
```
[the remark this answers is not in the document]
ethanfrey: Hmmm... I see the point of using streams internally, but do we have to expose it?

For example, if we use something like this (from the test code):
```
      const pendingResult = toListPromise(connection.waitForTransaction(transactionId), 1);

      // send transaction
      await connection.postTx(liskCodec.bytesToPost(signed));

      // wait for transaction
      const result = (await pendingResult)[0];
```

we will block forever and never unsubscribe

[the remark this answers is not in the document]
ethanfrey: To make it easy for the client to use the library properly, I would propose using the streams and unsubscribe *internally* as you did here, but exposing with a timeout parameter, where it would unsubscribe all (potentially polling) subscriptions and reject the promise.

eg. `const result = await connection.waitForTransaction(transactionId, 15)` will cancel all subscriptions and reject the promise after 15 seconds. 

If you want to expose all this power to the users, assuming they want real reactive, maybe leave this as is, and add `waitForTransactionWithTimeout`, which returns a promise, and would be the simple API for the non power-user

packages/iov-bcp-types/src/connection.ts
webmaster128: > toListPromise(..., 1) and waiting for stream completion will also block in the same manner.

Right, every promise can be converted to a stream without losing information. However, not every stream can be converted into a promise. What `toListPromise` does is removing the power to unsibscribe, which is okay for tests that will timeout.

webmaster128:
@ethanfrey What about this alternative structure for cancleable promises:

```ts
interface CancelablePromise<T> {
    readonly result: Promise<T>;
    readonly cancel: () => void;
}
```

That keeks 0 or 1 semantics but still allows cancellation.

webmaster128:
an additional `.result` is for sure easier than `(await toListPromise(..., 1))[0]`

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -190,55 +193,68 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-            next: async blockHeader => {
-              const event: BcpBlockInfo = {
-                state: BcpTransactionState.InBlock,
+    const firstEvent: BlockInfo = { state: TransactionState.Pending };
+    let lastEventSent: BlockInfo = firstEvent;
+    const blockInfoProducer = new DefaultValueProducer<BlockInfo>(firstEvent, {
+      onStarted: () => {
+        transactionSubscription = this.liveTx({ id: transactionId }).subscribe({
+          next: searchResult => {
+            if (isFailedTransaction(searchResult)) {
```
webmaster128: This is the core of this PR. Due to an improved liveTx that can handle failed transactions, postTx now supports propagating delivertx errors.

ethanfrey:
Interesting idea with Cancellable Promises. I like.

But maybe a different PR.
I will review this tomorrow morning.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -113,7 +114,7 @@ describe("BnsConnection", () => {
     const nonce = await connection.getNonce({ pubkey: identity.pubkey });
     const signed = await profile.signTransaction(firstWalletId, identity, sendTx, bnsCodec, nonce);
     const response = await connection.postTx(bnsCodec.bytesToPost(signed));
-    await response.blockInfo.waitFor(info => info.state === BcpTransactionState.InBlock);
+    await response.blockInfo.waitFor(info => info.state !== TransactionState.Pending);
```
ethanfrey: This is a very common idiom, maybe expose this as a helper function (like you do with `isFailedTransaction`).

`function transactionProcessed(info: BlockInfo): boolean { return info.state !== TransactionState.Pending }`

It's not so much for the typing, but to avoid people using other checks (like `info => info.state === TransactionState.Success`) when they want this. They can use the other checks when they really only want the other case.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -436,7 +437,7 @@ describe("BnsConnection", () => {
       await tendermintSearchIndexUpdated();
 
       // now verify we can query the same tx back
-      const search = await connection.searchTx({ sentFromOrTo: faucetAddr });
+      const search = (await connection.searchTx({ sentFromOrTo: faucetAddr })).filter(isConfirmedTransaction);
```
ethanfrey: :+1:

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -869,13 +879,13 @@ describe("BnsConnection", () => {
       const nonce = await connection.getNonce({ pubkey: faucet.pubkey });
       const signed = await profile.signTransaction(mainWalletId, faucet, sendTx, bnsCodec, nonce);
       const response = await connection.postTx(bnsCodec.bytesToPost(signed));
-      const blockInfo = await response.blockInfo.waitFor(info => info.state === BcpTransactionState.InBlock);
-      const txHeight = (blockInfo as BcpBlockInfoInBlock).height;
+      const blockInfo = await response.blockInfo.waitFor(info => info.state === TransactionState.Succeeded);
```
ethanfrey: What about `.waitFor(transactionProcessed)` and then assert?

As written, if the transaction fails, it will hang forever.
Using my suggestion, we can check the state and provide a much clearer error message in case the tx ever fails (including the log/msg from the blockchain to help debug)

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1014,6 +1020,45 @@ describe("BnsConnection", () => {
...
+        recipient: await randomBnsAddress(),
+        memo: "Sending from empty",
+        amount: defaultAmount,
+      };
+
+      const nonce = await connection.getNonce({ pubkey: brokeIdentity.pubkey });
+      const signed = await profile.signTransaction(mainWalletId, brokeIdentity, sendTx, bnsCodec, nonce);
+      const response = await connection.postTx(bnsCodec.bytesToPost(signed));
+      const transactionIdToSearch = response.transactionId;
+      await response.blockInfo.waitFor(info => info.state !== TransactionState.Pending);
```
ethanfrey: Same point with waitForCondition (`transactionProcessed`)

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1014,6 +1020,45 @@ describe("BnsConnection", () => {
...
+
+      await tendermintSearchIndexUpdated();
```
ethanfrey: This is an interesting issue.
We use sleep now.
Digging into those issues with bugs in wallet-demo, I found the dispatch code. Maybe we can get a tendermint event for this case (`search_index_updated`) shortly after all headers and tx events are emitted.

ethanfrey: (Anton, tendermint core dev who build the indexing system, also proposed this `IndexedBlock` event... so maybe in the not too distance future)

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1064,6 +1110,158 @@ describe("BnsConnection", () => {
...
+        amount: defaultAmount,
+      };
+
+      const nonce = await connection.getNonce({ pubkey: faucet.pubkey });
+      const signed = await profile.signTransaction(mainWalletId, faucet, sendTx, bnsCodec, nonce);
+      const response = await connection.postTx(bnsCodec.bytesToPost(signed));
+      const transactionIdToSearch = response.transactionId;
+      await response.blockInfo.waitFor(info => info.state !== TransactionState.Pending);
+
+      await tendermintSearchIndexUpdated();
```
ethanfrey: Actually, this brings up an interesting point for me....

If we call `liveTx` 2 ms after the tx event was emitted, but 10 ms before the search index was updated, then it may get no event at all.

For true robustness, we should update blockInfo not after the Header event, but after the propsoed `IndexedBlock` event (of course this waits for some future version of tendermint, just wanted to write down idea as I was noticing it in code)

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1064,6 +1110,158 @@ describe("BnsConnection", () => {
...
+
+      const nonce = await connection.getNonce({ pubkey: brokeIdentity.pubkey });
+      const signed = await profile.signTransaction(mainWalletId, brokeIdentity, sendTx, bnsCodec, nonce);
+      const response = await connection.postTx(bnsCodec.bytesToPost(signed));
+      const transactionIdToSearch = response.transactionId;
+      await response.blockInfo.waitFor(info => info.state !== TransactionState.Pending);
+
+      await tendermintSearchIndexUpdated();
+
+      const result = (await toListPromise(connection.liveTx({ id: transactionIdToSearch }), 1))[0];
```
ethanfrey: Just another convenience function we might want:
`await firstEvent(stream)` 
to be the same as
`(await toListPromise(stream, 1)[0])`

ethanfrey: Not cuz it is hard, but to reduce typos, etc for clients.
No stress, I can add all helpers in wallet-demo if you don't like them in iov-core. Just proposing.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1556,19 +1703,27 @@ describe("BnsConnection", () => {
...
 
-    const txByRecipient = await connection.searchTx({ tags: [bnsSwapQueryTags(querySwapRecipient)] });
+    const txByRecipient = (await connection.searchTx({
+      tags: [bnsSwapQueryTags(querySwapRecipient)],
+    })).filter(isConfirmedTransaction);
     expect(txByRecipient.length).toEqual(1);
     expect(txByRecipient[0].transactionId).toEqual(transactionId);
 
-    const txByHash = await connection.searchTx({ tags: [bnsSwapQueryTags(querySwapHash)] });
+    const txByHash = (await connection.searchTx({ tags: [bnsSwapQueryTags(querySwapHash)] })).filter(
```
ethanfrey: If swapQueryTags are related to state defined in `bcp/*` transactions not just `bns/*` specific ones, we should probably define a generic way to create them in the `searchTx` query parameters. Just like `sentToOrFrom`.

This can be a different PR, but I figure a minor update for 0.11 (or 0.12)

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1705,20 +1860,20 @@ describe("BnsConnection", () => {
...
+      await post.blockInfo.waitFor(info => info.state !== TransactionState.Pending);
     }
 
     // start to watch
     const liveView = asArray(connection.watchSwap(rcptQuery));
 
     const post3 = await openSwap(connection, profile, faucet, recipientAddr, preimage3);
-    const blockInfo3 = await post3.blockInfo.waitFor(info => info.state === BcpTransactionState.InBlock);
-    const id3 = (blockInfo3 as BcpBlockInfoInBlock).result! as SwapIdBytes;
+    const blockInfo3 = await post3.blockInfo.waitFor(info => info.state === TransactionState.Succeeded);
```
ethanfrey: Again, I don't like this, as it will hang forever if our call errors, with no useful diagnostics.
How about waiting for `Succeeded|Failure`, and then asserting it was a success, where we can easily dump out the tendermint logs for debugging.

There are dozens of such cases, I will stop commenting on them, but I do see as it is, if there are any DeliverTx failures when I do the weave v0.10 upgrade, I will end up with hanging tests and no error messages

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -190,55 +193,68 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+              const errorEvent: BlockInfoFailed = {
+                state: TransactionState.Failed,
+                code: searchResult.code,
+                log: searchResult.log,
+              };
+              blockInfoProducer.update(errorEvent);
```
ethanfrey: Maybe this is handled somewhere else, but shouldn't we do `transactionSubscription.unsubscribe()` here, as well as completing the `blockInfoProducer` stream. We should expect no new events after a failure

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -190,55 +193,68 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+              },
+              complete: () => blockInfoProducer.error("Block header stream stopped. This must not happen."),
+              error: error => blockInfoProducer.error(error),
+            });
+          },
+          error: error => blockInfoProducer.error(error),
+        });
       },
       onStop: () => {
+        if (transactionSubscription) {
```
ethanfrey: I guess we rely on the client unsubscribing after a failure is returned for the proper processing as I mentioned above. I guess this works, seems a bit harder to follow logic, as it depends on the actions of other parts.

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -444,46 +460,68 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+            height: height,
+            confirmations: currentHeight - height + 1,
+            transactionId: transactionId,
+            log: result.log,
+            result: result.data,
+            ...this.codec.parseBytes(new Uint8Array(tx) as PostableBytes, chainId),
+          };
+        } else {
+          return {
+            transactionId: transactionId,
```
ethanfrey: We could fit height in here as I mentioned somewhere above.
I like to provide extra context that may be useful.

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -444,46 +460,68 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-        }),
-      );
+    return this.tmClient.subscribeTx(rawQuery).map(
+      (transaction): ConfirmedTransaction | FailedTransaction => {
+        const transactionId = Encoding.toHex(transaction.hash).toUpperCase() as TransactionId;
+
+        if (transaction.result.code === 0) {
+          return {
+            height: transaction.height,
+            confirmations: 1, // assuming block height is current height when listening to events
```
ethanfrey: this is a safe assumption, the height on the TxEvent is the height of the block it was contained in.

packages/iov-lisk/src/liskconnection.ts
```diff
@@ -388,6 +369,43 @@ export class LiskConnection implements BcpConnection {
     }
   }
 
+  private waitForTransaction(id: TransactionId): Stream<ConfirmedTransaction | FailedTransaction> {
```
ethanfrey: looks good.
simpler than the tendermint event model but less efficient.

packages/iov-bcp-types/src/connection.ts
```diff
@@ -103,12 +105,25 @@ export interface BcpBlockInfoInBlock {
...
+  readonly state: TransactionState.Failed;
+  /**
+   * Application specific error code
+   */
+  readonly code: number;
+  /**
+   * Application specific logging output in an arbitrary text format that
+   * may change at any time.
+   */
+  readonly log?: string;
```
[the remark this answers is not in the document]
webmaster128: Right, good point. At BCP level we should not blindly copy Tendermint terminology, which is implementation detail. Will rename.

packages/iov-bcp-types/src/connection.ts
```diff
@@ -124,10 +139,40 @@ export interface ConfirmedTransaction<T extends UnsignedTransaction = UnsignedTr
...
   readonly result?: Uint8Array;
+  /**
+   * Application specific logging output in an arbitrary text format that
+   * may change at any time.
+   */
   readonly log?: string;
   // readonly tags: ReadonlyArray<Tag>;
 }
 
+export interface FailedTransaction {
```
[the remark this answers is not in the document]
webmaster128: Right. There more common fields `ConfirmedTransaction | FailedTransaction`, the easier it is to work on the union type instead of handling case-by-case.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Hmm, the helper function already exists in the wallet and I think we should be very careful with the amount of APIs we expose.

I wanted to suggest type checkers `isBlockInfoPending`,  `isBlockInfoSucceeded`, `isBlockInfoFailed` like we have for many union types. However, we can only use them directly if we expect success:

```ts
// ignore failures
await response.blockInfo.waitFor(isBlockInfoSucceeded)

// what we want
await response.blockInfo.waitFor(info => !isBlockInfoPending(info));
```

The last one is probably not what you desire, but better than what we have using a very generic API.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: > As written, if the transaction fails, it will hang forever.

Yes, I'm aware of this. This is fine for tests which time out after 15 seconds

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: What's the issue here? This is the default statement for _wait for block_

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Right now we consistently delay 50ms between block event and searches. This is done in all Tendermint and BNS tests. Before we had many race condition issues and tests that ran on a dev machine but occasionally failed on Travis. Since we use the 50 ms, everything is fine.

But of course, having an event that fires after the new block is indexed would be much more convenient.

packages/iov-bns/src/bnsconnection.ts
webmaster128: Hmm good question. The return type ValueAndUpdates does not support any kind of completion. There is no _value will not change anymore_ state.

packages/iov-bns/src/bnsconnection.spec.ts
ethanfrey: I am fine with not exposing more and I will have fun in the client.
My major issue is that many tests use success/fail checks that may hang forever when the test would otherwise fail. You could make `transactionProcessed` unexported and use it in tests. But some way that broken tests report good errors.

packages/iov-bns/src/bnsconnection.ts
webmaster128: Nope, the client does not need to unsubscribe. In case of FailedTransaction, there is one event, an open stream but no further events.

We don't waste resources in this case because `blockHeadersSubscription = this.watchBlockHeaders().subscribe({` is never started for FailedTransaction.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: > My major issue is that many tests use success/fail checks that

Right, I will get rid of all `=== TransactionState.Success` in favour of `!== TransactionState.Pending` and check type explicitly

webmaster128:
All comments resolved :)

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1143,7 +1143,7 @@ describe("BnsConnection", () => {
       await tendermintSearchIndexUpdated();
 
       // finds transaction using id
-      const result = (await toListPromise(connection.liveTx({ id: transactionIdToSearch }), 1))[0];
+      const result = await firstEvent(connection.liveTx({ id: transactionIdToSearch }));
```
ethanfrey: :smile: 
Thank you!

PR #724, opened 2019-01-25
Assert 8 fractional digits in dpos serialization

### LINKED ISSUE #723: Lisk codec does not consider fractionalDigits of amount
```
  bytesToPost: (signed: SignedTransaction): PostableBytes => {
    const unsigned = signed.transaction;
    if (isSendTransaction(unsigned)) {
      const timestamp = signed.primarySignature.nonce.toNumber();
      const liskTimestamp = timestamp - 1464109200;
      const id = Serialization.transactionId(
        unsigned,
        new ReadonlyDate(timestamp * 1000),
        signed.primarySignature,
        constants.transactionSerializationOptions,
      );

      const postableObject = {
        type: 0,
        amount: unsigned.amount.quantity,
        recipientId: unsigned.recipient,
        senderPublicKey: Encoding.toHex(signed.primarySignature.pubkey.data),
        timestamp: liskTimestamp,
        fee: "10000000", // 0.1 LSK fixed
        asset: {
          data: unsigned.memo,
        },
        signature: Encoding.toHex(signed.primarySignature.signature),
        id: id,
      };
      return Encoding.toUtf8(JSON.stringify(postableObject)) as PostableBytes;
    } else {
      throw new Error("Unsupported kind of transaction");
    }
```

When paired with a default 9 fractionalDigits in the wallet (from the bns default), this leads to 10x too much LSK being sent.

Will enforce 8, as we enforce 9 in bns.

ethanfrey:
Closes #723 

Also addresses the same issue in Rise

### DIFF
--- a/packages/iov-dpos/src/serialization.spec.ts
+++ b/packages/iov-dpos/src/serialization.spec.ts
@@ -143,6 +143,29 @@ describe("Serialization", () => {
       );
     });
 
+    it("throws error is fractionalDigits are not correct", () => {
+      const pubkey = fromHex("00112233445566778899aabbccddeeff00112233445566778899aabbccddeeff");
+      const tx: SendTransaction = {
+        kind: "bcp/send",
+        creator: {
+          chainId: riseTestnet,
+          pubkey: {
+            algo: Algorithm.Ed25519,
+            data: pubkey as PublicKeyBytes,
+          },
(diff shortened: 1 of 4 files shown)


ethanfrey:
Just added a few more tests to assert this in higher-level api

PR #725, opened 2019-01-25
Add Tendermint 0.27

webmaster128:


packages/iov-tendermint-rpc/README.md
```diff
@@ -23,7 +23,7 @@ const status = await client.status();
 
 | IOV-Core version | Supported tendermint versions |
 |------------------|-------------------------------|
-| 0.11             | 0.25.x                        |
+| 0.11             | 0.25.x, 0.27.x                |
```
ethanfrey: :+1:

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -29,20 +30,29 @@ function pendingWithoutTendermint(): void {
...
-  // {
-  //   url: "localhost:11121",
-  //   version: "0.21.x",
-  // },
   {
     url: "localhost:12345",
     version: "0.25.x",
+    appCreator: "jae",
   },
+  // {
```
ethanfrey: Commented out?
This looks like a great config with appCreator and all

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -92,6 +102,18 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor): void {
     client.disconnect();
   });
 
+  it("gets the same tx hash from backend as calculated locally", async () => {
```
ethanfrey: Yes, this changed between versions, right?

packages/iov-tendermint-rpc/src/requests.ts
```diff
@@ -63,7 +63,13 @@ export interface AbciQueryParams {
   readonly path: string;
   readonly data: Uint8Array;
   readonly height?: number;
-  readonly trusted?: boolean;
+  /**
+   * A flag that defines if proofs are included in the response or not.
+   *
+   * Internally this is mapped to the old name `trusted` for Tendermint < 0.26.
+   * Starting with Tendermint 0.26, the default value changed from true to false.
+   */
```
ethanfrey: And before v0.12.0, it was also prove....
Oh, I remember the debate to change this

https://github.com/tendermint/tendermint/blob/v0.11.0/rpc/core/abci.go#L44-L49

packages/iov-tendermint-rpc/src/v0-25/requests.ts
```diff
@@ -96,12 +96,13 @@ interface RpcAbciQueryParams {
...
   readonly trusted?: boolean;
 }
+
 function encodeAbciQueryParams(params: requests.AbciQueryParams): RpcAbciQueryParams {
   return {
     path: notEmpty(params.path),
     data: Encoding.toHex(params.data) as HexString,
     height: may(Integer.encode, params.height),
-    trusted: params.trusted,
+    trusted: params.prove,
```
ethanfrey: `trusted: !params.prove`
they are opposites

packages/iov-tendermint-rpc/src/client.spec.ts
webmaster128: The stop script only supports one container labeled "tendermint/tendermint" right now. I think it is out of scope to run both in the CI. It can be done, but it is some work.

What I commented out is for local development. It works fine!

packages/iov-tendermint-rpc/src/client.spec.ts
webmaster128: Correct. I made hash test in the test abolve less specific (was fixed to length 20) and here I test hasher against backend. It is 20/32 bytes depending on the version but the test is generic and supports both

packages/iov-tendermint-rpc/src/v0-25/requests.ts
webmaster128: oh! oh oh! well spotted

webmaster128: Like this? https://github.com/iov-one/iov-core/pull/725/commits/8b5771288f65bef022799f87c91d6ca235c17fa8

ethanfrey: Yes, that looks like a good fix

PR #729, opened 2019-01-27
Run Tendermint 25 and 27 tests by default; many Tendermint improvements

webmaster128:
* Run Tendermint 25 and 27 tests by default
* Avoid exporting internal, version specific interfaces
* Stronger, more consistent run time checks on return values
* Multiple API fixes
* Some new fields now exposed


webmaster128:
By running Tendermint 27 in the CI I discovered a missing CORS configuration when running the browser tests (good we found it). However, looks like we cannot set the value we need: https://github.com/tendermint/tendermint/issues/3216

.travis.yml
```diff
@@ -2,6 +2,8 @@ os:
   - linux
   - osx
 
+dist: xenial
```
ethanfrey: :smile:

packages/iov-tendermint-rpc/src/client.spec.ts
```diff
@@ -143,12 +112,26 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor): void {
...
     client.disconnect();
   });
 
+  it("can call status", async () => {
+    pendingWithoutTendermint();
+    const client = new Client(rpcFactory(), adaptor);
+
+    const status = await client.status();
+    expect(status.nodeInfo.other.size).toBeGreaterThanOrEqual(2);
+    expect(status.nodeInfo.other.get("tx_index")).toEqual("on");
```
ethanfrey: great check

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -9,30 +8,88 @@ export type HexString = string & As<"hex">;
...
+  return value;
+}
+
+/**
+ * A runtime checker that ensures a given value is a boolean
+ *
+ * This is used when you want to verify that data at runtime matches the expected type.
+ * This implies assertSet.
+ */
+export function assertBoolean(value: boolean): boolean {
```
ethanfrey: I like these deeper asserts

scripts/socketserver/echo.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
```
ethanfrey: Interesting.... this adds another requirement of python to the dev setup....
I'd prefer to have this in node to reduce our requirements, but for a quick setup this is fine.

I don't know if all mac or windows installs have python3 by default (but Ubuntu does).

Maybe a PR in 0.12.0 to reduce this requirement (doesn't block 0.11 and not urgent til a dev complains, but a general feedback)

scripts/travis.sh
```diff
@@ -33,14 +33,22 @@ function fold_end() {
 source ./scripts/retry.sh
 retry 3 yarn install
 
+# For socketserver
+pip3 install websockets
```
ethanfrey: This is yet another requirement, that pip3 is installed on dev machine.
No issue for me, but something I am wary of.

webmaster128:
I'd like to keep testing infrastructure out of our nodejs dependency tree, to keep that manageable. What we could do is wrap the socket server in a docker container to use a consistent setup.

ethanfrey:
Docker solution sounds good.
But as a different release. No need to put any more into 0.11... frontend is ready to build on it

PR #743, opened 2019-01-30
Split Travis matrix in tests, docs, lint

webmaster128:


ethanfrey:
Nice and much simpler than expected.

Did you notice any change in ci response time

webmaster128:
> Did you notice any change in ci response time

The main builds are now 2 1/2 minutes faster (1 minue lint and 1 1/2 minutes docs). However, given the recent ci time increases this reduction is negligible. Recent CI time increase comes from

* Mac build start very delayed (maybe temporary Travis problem but mac on Travis is problematic for years)
* All there event based stream tests are killing CI times
* Recent addition of liveTx/listenTx/searchTx tests for blockchains with 10-30 seconds block time. One RISE test needs a 90 second timeout and still times out occasionally.

I see this PR mostly as a proof of concept for how to split out tests into independent, parallizable parts.

PR #747, opened 2019-01-31
Test the encoding and use of fees; Upgrade bnsd

### LINKED ISSUE #686: Add support for fees
We have a fees field in all transactions, but I don't know if we actually use it when sending transactions.

We will soon (https://github.com/iov-one/weave/issues/264) start enforcing minimum fees server side and putting them in a "tip jar" Let's make sure we properly set the fees field if the client sends it (this can also be tested with current codebase... server handles it since May, just doesn't enforce a minimum).

Relevant code in weave if you want to look at it:

Implementation: https://github.com/iov-one/weave/blob/master/x/cash/decorator.go
Location in transaction: https://github.com/iov-one/weave/blob/master/cmd/bnsd/app/codec.proto#L29 (different field number, but same format as in 0.9)

### LINKED ISSUE #744: Test master using BNSd 0.11.x and Tendermint 0.29.x
Should work without code changes. @alpe do we have docker images ready to use for this setup?

webmaster128:
Closes #686
Closes #744

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -459,6 +459,53 @@ describe("BnsConnection", () => {
       connection.disconnect();
     });
 
+    it("can send transaction with fees", async () => {
+      pendingWithoutBnsd();
+      const connection = await BnsConnection.establish(bnsdTendermintUrl);
+      const chainId = connection.chainId();
+
+      const { profile, mainWalletId, faucet } = await userProfileWithFaucet(chainId);
+
+      const initialBalance = (await connection.getAccount({ pubkey: faucet.pubkey }))!.balance.find(
+        coin => coin.tokenTicker === cash,
+      )!;
+
+      const sendTx: SendTransaction = {
+        kind: "bcp/send",
+        creator: faucet,
+        recipient: await randomBnsAddress(),
+        memo: "This time I pay my bills",
+        amount: {
+          quantity: "100",
+          fractionalDigits: 9,
+          tokenTicker: cash,
+        },
+        fee: {
+          quantity: "2",
+          fractionalDigits: 9,
+          tokenTicker: cash,
+        },
+      };
+      const nonce = await connection.getNonce({ pubkey: faucet.pubkey });
+      const signed = await profile.signTransaction(mainWalletId, faucet, sendTx, bnsCodec, nonce);
+      const response = await connection.postTx(bnsCodec.bytesToPost(signed));
+      const blockInfo = await response.blockInfo.waitFor(info => !isBlockInfoPending(info));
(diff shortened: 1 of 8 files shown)


scripts/bnsd/bnsd_app.sh
```diff
@@ -16,6 +16,7 @@ if [ ! -d "${BNSD_DIR}" ]; then
 fi
 
 exec docker run --user="$UID" \
+  --name "bnsd-app" \
```
ethanfrey: add `--rm` there for better cleanup

scripts/bnsd/bnsd_tm.sh
```diff
@@ -23,6 +23,7 @@ fi
 
 # tx indexing set in init
 exec docker run --user="$UID" \
+  --name "bnsd-tendermint" \
```
ethanfrey: As above, with `--rm`

scripts/bnsd/start.sh
```diff
@@ -3,9 +3,9 @@ set -o errexit -o nounset -o pipefail
 command -v shellcheck > /dev/null && shellcheck "$0"
 
 # Choose from https://hub.docker.com/r/iov1/tendermint/tags/
-export BNSD_TM_VERSION=v0.27.4
+export BNSD_TM_VERSION=v0.29.1
```
ethanfrey: Nice

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -459,6 +459,53 @@ describe("BnsConnection", () => {
...
     });
 
+    it("can send transaction with fees", async () => {
+      pendingWithoutBnsd();
+      const connection = await BnsConnection.establish(bnsdTendermintUrl);
+      const chainId = connection.chainId();
+
+      const { profile, mainWalletId, faucet } = await userProfileWithFaucet(chainId);
+
+      const initialBalance = (await connection.getAccount({ pubkey: faucet.pubkey }))!.balance.find(
```
ethanfrey: nice one-liner

packages/iov-bns/src/encode.spec.ts
```diff
@@ -138,15 +139,97 @@ describe("Encode", () => {
     expect(encoded.signature!.ed25519).toEqual(fromHex("aabbcc22334455"));
   });
 
+  describe("buildUnsignedTx", () => {
```
ethanfrey: good tests

scripts/bnsd/bnsd_app.sh
webmaster128: Hmm, I now changed the stop script to delete the container. We should be consistent. Any good reasons for one or the other solution?

webmaster128: Postponed as this affects all containers we start: https://github.com/iov-one/iov-core/issues/748

PR #749, opened 2019-01-31
Move firefox test from mac to Linux

webmaster128:


webmaster128:
Baseline times, let see how good we get
bildschirmfoto 2019-01-31 um 18 13 32

webmaster128:
Moving firefox to Linux:
* Mac build: -5 minutes
* Linux build: +7 minutes

Great!

webmaster128:
[Remove all non-safari tests from Mac](https://travis-ci.com/iov-one/iov-core/builds/99315281): -8:30 minutes

webmaster128:
Okay, we now have 5 jobs running <= 12 minutes, which is the best I see right now given the tests we want to perform. Many smaller jobs come with the additional advantage that restarting false positives can be done quicker and one CI run can detect multiple issues at the same time.

PR #757, opened 2019-02-02
Remove package @iov/ledger-bns from this monorepo

### LINKED ISSUE #756: Extract package @iov/ledger-bns from monorepo
This is a follow-up on https://github.com/iov-one/iov-core/issues/389. Not that APIs are stabilizing a bit and the Ledger app will not be the first thing to be released, we concluded to extract the package into its own repo

webmaster128:
It is now available at https://github.com/iov-one/iov-ledger-bns

Closes #756


webmaster128:
Added a link to IOV-Core users for now: https://github.com/iov-one/iov-core/blob/master/README.md#users We can make it more prominent as the project stabilized.

PR #754, opened 2019-02-01
Allow atomic swap merger to receive claim/timeout before open

webmaster128:
Closes #556

PR #761, opened 2019-02-03
Create cross-chain atomic swap example

### LINKED ISSUE #6: "Manual" atomic swap
CLI should allow for easy command-line execution of Atomic Swaps

webmaster128:
Based on #754

Closes #6


webmaster128:
Rebased onto current master and ready for review

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ Breaking changes
   `AtomicSwapRecipientQuery`, `AtomicSwapSenderQuery`, `AtomicSwapIdQuery`,
   `AtomicSwapHashlockQuery`, `isAtomicSwap*Query`.
 * @iov/bcp-types: Rename `bnsSwapQueryTags` to `bnsSwapQueryTag`
+* @iov/bcp-types: Let `SwapOfferTransaction` take a `hash` instead of a `preimage`.
```
ethanfrey: Doing this means SwapOffer and SwapCounter are the same and can be combined.

I guess this means we need a standard preimage calculator exposed, which upon reflection is important. If I use a different hash algorithm on chain A and chain B, then I can claim this, while giving a preimage that is useless.

We should also standardize maximum allowed size of the preimage.

But that is a different issue...

packages/iov-bcp-types/src/atomicswapmerger.spec.ts
```diff
@@ -158,15 +158,15 @@ describe("AtomicSwapMerger", () => {
     const alice = "tiov1u8syu9juwx668k4vqfwl5vtm8j6yz89wamkcda" as Address;
     const bobAddress = "tiov1lpzdluzsq3u7tqkfkp3rmrfavkhv0ly56gjexe" as Address;
     const preimage = fromHex("00110011");
-    const hashLock = new Sha256(preimage).digest();
+    const hash = new Sha256(preimage).digest();
```
ethanfrey: Can we export a function to do this?

There is one in iov-bns/src/utils.ts:
```ts
export function preimageIdentifier(preimage: Uint8Array): HashId {
  const hash = new Sha256(preimage).digest();
  return hashIdentifier(hash);
}
```

maybe we can just move this into bcp(-types)?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1874,27 +1874,27 @@ describe("BnsConnection", () => {
 
     // create the preimages for the three swaps
     const preimage1 = Encoding.toAscii("the first swap is easy");
-    // const hash1 = new Sha256(preimage1).digest();
+    const hash1 = new Sha256(preimage1).digest();
```
ethanfrey: Yeah, I really don't want to have to leak this all over the place, especially if we start putting checks and validation in there... see above comment about preimageIdentifier

packages/iov-bns/src/context.ts
```diff
@@ -60,44 +58,36 @@ export class Context {
...
 
-  public swapOffer(swap: codecImpl.escrow.Escrow & Keyed): AtomicSwap {
-    // TODO: get and check hashlock
-    let hashlock: Uint8Array;
-    if (isHashIdentifier(swap.arbiter)) {
-      hashlock = hashFromIdentifier(swap.arbiter);
-    } else {
+  /** Decode within a Context to have the chain ID available */
+  public decodeOpenSwap(swap: codecImpl.escrow.Escrow & Keyed): OpenSwap {
+    if (!isHashIdentifier(swap.arbiter)) {
```
ethanfrey: Nice cleanup and clearer flow

packages/iov-bns/src/encode.ts
```diff
@@ -162,22 +162,23 @@ function buildSendTransaction(tx: SendTransaction): codecImpl.app.ITx {
...
 
 function buildSwapOfferTx(tx: SwapOfferTransaction): codecImpl.app.ITx {
-  const hashed: SwapCounterTransaction = {
-    kind: "bcp/swap_counter",
-    creator: tx.creator,
-    recipient: tx.recipient,
-    amounts: tx.amounts,
-    timeout: tx.timeout,
-    hashCode: preimageIdentifier(tx.preimage),
+  return {
```
ethanfrey: This is a copy of `buildSwapCounterTx`

If we no longer have the preimage/hash distinction, I wonder why we need two types here...
They are both "offers" on the respective chains, the interpretation of them is only whether the client produced the preimage for the hash, or whether it copied the hash without knowing the preimage.

Can we simplify by removing types?

packages/iov-core/src/integrationtests/atomicswap.spec.ts
```diff
@@ -0,0 +1,313 @@
+import BN = require("bn.js");
```
ethanfrey: I like the multi-chain integration test

CHANGELOG.md
webmaster128: You are right, SwapOfferTransaction and SwapCounterTransaction contains the same fields (probably they are implemented as the same transaction in BNS). But maybe they'll diverge at some point? For me the differentiation right now helps to express the intent from a user's perspective.

> If I use a different hash algorithm on chain A and chain B, then I can claim this, while giving a preimage that is useless.

The hashing algorithm must be the same on both chains (for the reason you gave). Right now this algorithm is hardcoded to SHA256, which is kind of globally known. We could make it configurable probably but I see this as a pre-launch thing.

> We should also standardize maximum allowed size of the preimage.

Very valid and non-trivial point, especially since the blockchain cannot check the preimage before the swap happens. So I guess we need to ensure the preimage-length is limited by the same amount of bytes on all supported chains to avoid half atomic swaps.

packages/iov-bcp-types/src/atomicswapmerger.spec.ts
webmaster128: The whole idea of this identifier is BNS-specific. What we can do is expose a preimage->hash helper. And at that point the question arises if it is worth exposing a SHA56 hasher.

I think this starts to make sense as soon as miltiple algorithms are supported.

packages/iov-bns/src/encode.ts
webmaster128: What was importent for me was to not pass the preimage as part of a transaction, as it suggests the secret becomes public data at a point where it should be a client-side secret.

So yeah, we ended up with two transaction types that have the same fields and are implemented as the same thing on BNS. But will they always be the same transaction from the blockchain's point of view?

But even if we keep the two BCP transaction types, we can combine the implementation into

```ts
function buildSwapCounterTx(tx: SwapOfferTransaction | SwapCounterTransaction): codecImpl.app.ITx
```

ethanfrey: > So yeah, we ended up with two transaction types that have the same fields and are implemented as the same thing on BNS. But will they always be the same transaction from the blockchain's point of view?

The issue is that they are encoded the same on the blockchain, so when parsing you have to assume type A or B, regardless of what was written.

And since each chain is a word unto itself, what we have is basically:

`Offer -> Claim` OR `Offer -> Return`

on each side.

ethanfrey: There is no way to differentiate the two without client side knowledge of both chains and intent, which we cannot codify

packages/iov-bcp-types/src/atomicswapmerger.spec.ts
ethanfrey: Yeah, sorry, this Id is bns specific.
You are write about the hash being different.

But I would like to make something like 

```ts
export function hashPreimage(preimage: Uint8Array): Uint8Array {
  return new Sha256(preimage).digest();
}
``

Where we can add some client-side checks for the preimage length later, when we start enforcing them in the blockchain.

But sure, I guess we can leave it as is for now.

CHANGELOG.md
ethanfrey: > You are right, SwapOfferTransaction and SwapCounterTransaction contains the same fields (probably they are implemented as the same transaction in BNS). But maybe they'll diverge at some point?

The transactions on the blockchain cannot diverge at any point. I used this for different intents when adding a pre-processing, but I agree that added confusion. As mentioned below, once you make them the same fields, please make them the same type. It will also make parsing more clear

webmaster128:
All comments addressed. "Remove SwapCounterTransaction in favour of SwapOfferTransaction" is the first new commit. Please have another look @ethanfrey

PR #763, opened 2019-02-06
Upgrade bnsd version and genesis to v0.11.1

alpe:
* Adds required `gconf` to genesis
* Bnds version upgrade


webmaster128:
@alpe could you look into the CI fails here?

PR #781, opened 2019-02-27
Update iov-core README with bov-hugnet etc

### LINKED ISSUE #689: Wrong testnet URL
https://github.com/iov-one/iov-core/blob/master/packages/iov-core/README.md contains wrong link to testnet (https://bov.friendnet-fast.iov.one/). It shows "Server not found" error.

willclarktech:
Quite a lot was out of date, so I updated some other things too as it was useful to get familiar with the codebase, however I left the following out as they seemed very out of scope:

- `const acct = await connection.getAccount({ address: bert });` fails because of a bech32 issue
- history didn't seem to work
- reactive clients uses a `client` which hasn't previously been defined

Closes #689

### DIFF
--- a/packages/iov-core/README.md
+++ b/packages/iov-core/README.md
@@ -74,7 +74,8 @@ profile.setWalletLabel(wallet2.id, "24 words");
 Create identies on the two wallets:
 
 ```ts
-import { ChainId, HdPaths } from '@iov/keycontrol';
+import { ChainId } from '@iov/bcp-types';
+import { HdPaths } from '@iov/keycontrol';
 import { Encoding } from '@iov/encoding';
 const { fromHex, toHex } = Encoding;
 
@@ -186,10 +187,9 @@ Query the testnet for some existing genesis accounts:
 // this is pulled from the genesis account
 import { Address } from "@iov/bcp-types"
 
-const bert = "E28AE9A6EB94FC88B73EB7CBD6B87BF93EB9BEF0" as Address;
+const bert = "tiov1u29wnfhtjn7g3de7kl9adwrmlyltn0hsjckecc" as Address;
 const acct = await connection.getAccount({ address: bert });
 console.log(acct);
-console.log(acct.data[0])
 ```
 
 If you are running the testnet faucet, just ask for some free money.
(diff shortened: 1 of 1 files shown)


packages/iov-core/README.md
```diff
@@ -74,21 +74,22 @@ profile.setWalletLabel(wallet2.id, "24 words");
 Create identies on the two wallets:
 
 ```ts
-import { HdPaths } from '@iov/keycontrol';
+import { ChainId, HdPaths } from '@iov/keycontrol';
```
webmaster128: Does this work? `ChainId` should come from @iov/bcp-types

packages/iov-core/README.md
```diff
@@ -186,14 +185,11 @@ Query the testnet for some existing genesis accounts:
...
 import { Address } from "@iov/bcp-types"
-const bert = "E28AE9A6EB94FC88B73EB7CBD6B87BF93EB9BEF0" as Address;
-const faucet = await connection.getAccount({ address: bert });
-console.log(faucet);
-console.log(faucet.data[0])
 
-// you can also query by registered name
-const byName = await connection.getAccount({ name: "bert" });
-console.log(byName.data[0])
+const bert = "E28AE9A6EB94FC88B73EB7CBD6B87BF93EB9BEF0" as Address;
```
webmaster128: Could you use the https://iov-one.github.io/token-finder/ to convert the old hex address format into bech32 using the "tiov" prefix?

packages/iov-core/README.md
```diff
@@ -186,14 +185,11 @@ Query the testnet for some existing genesis accounts:
 ```ts
 // this is pulled from the genesis account
 import { Address } from "@iov/bcp-types"
-const bert = "E28AE9A6EB94FC88B73EB7CBD6B87BF93EB9BEF0" as Address;
-const faucet = await connection.getAccount({ address: bert });
-console.log(faucet);
-console.log(faucet.data[0])
 
-// you can also query by registered name
-const byName = await connection.getAccount({ name: "bert" });
```
webmaster128: Good removal. getAccount by name does not exist anymore.

packages/iov-core/README.md
willclarktech: I've updated this, but the account is not found.

webmaster128: Yeah, no problem, thanks

Issue #400, opened 2018-10-05
Verify "Signer" proxy works out of process with simple rest calls


ethanfrey:
This builds on #399 and tests more realistic case where key management is in another process.

We launch a subprocess exposing an http server and containing a profile (which can auto-create in other process). We reproduce the flow described in #399 but using the http(s) api not in-process calls.

webmaster128:
I built a HTTP server that exposes the JSON-RPC interface.

Interesting observasion: since we currently use our types directly in the JSON-RPC interface, pubkeys are represented as lists of bytes:

```
curl -sS -X POST -d '{"jsonrpc":"2.0", "id": 1, "method": "getIdentities", "params": { "reason": "Who are you?", "chainIds": ["bov-hugnet"]}}' -H "Content-Type: application/json" http://localhost:8822/ | jq
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    {
      "chainId": "bov-hugnet",
      "pubkey": {
        "algo": "ed25519",
        "data": {
          "0": 1,
          "1": 118,
          "2": 9,
          "3": 140,
          "4": 187,
          "5": 227,
          "6": 59,
          "7": 55,
          "8": 170,
          "9": 178,
          "10": 104,
          "11": 111,
          "12": 31,
          "13": 175,
          "14": 110,
          "15": 216,
          "16": 168,
          "17": 159,
          "18": 38,
          "19": 100,
          "20": 63,
          "21": 174,
          "22": 244,
          "23": 167,
          "24": 11,
          "25": 144,
          "26": 58,
          "27": 163,
          "28": 9,
          "29": 73,
          "30": 53,
          "31": 227
        }
      }
    },
    {
      "chainId": "bov-hugnet",
      "pubkey": {
        "algo": "ed25519",
        "data": {
          "0": 51,
          "1": 178,
          "2": 31,
          "3": 3,
          "4": 34,
          "5": 230,
          "6": 70,
          "7": 157,
          "8": 25,
          "9": 183,
          "10": 39,
          "11": 37,
          "12": 25,
          "13": 49,
          "14": 127,
          "15": 84,
          "16": 253,
          "17": 124,
          "18": 40,
          "19": 225,
          "20": 160,
          "21": 148,
          "22": 201,
          "23": 127,
          "24": 218,
          "25": 202,
          "26": 193,
          "27": 17,
          "28": 173,
          "29": 2,
          "30": 92,
          "31": 167
        }
      }
    },
    {
[...]

PR #783, opened 2019-02-27
Implement confirm/deny logic for signing server

### LINKED ISSUE #651: Implement confirm/deny logic for signing server
> Maybe we pass in a "authorize" callback in the constructor, like

```
constructor(
  profile: UserProfile, 
  authorizeView: (reason: string, chains: ReadonlyArray<ChainId>) => Promise<boolean>,
  authorizeSign: (reason: string, transaction: UnsignedTransaction) => Promise<boolean>)
```

webmaster128:
Closes #651

### DIFF
--- a/docs/out-of-process-rpc.md
+++ b/docs/out-of-process-rpc.md
@@ -8,6 +8,7 @@ and can be implemented for example in JavaScript messages or deep-link URLs.
 
 | Name                | Definition                        |
 |---------------------|-----------------------------------|
+| `null`              | A representation of _no value_    |
 | List{T}             | A list of elements of type T      |
 | String              | A list of unicode codepoints      |
 | Identity            | An identity on a blockchain, defined as a pair of chain ID and public key |
@@ -18,7 +19,7 @@ and can be implemented for example in JavaScript messages or deep-link URLs.
 | Name          | Input parameters   | Return type   | Description   |
 |---------------|--------------------|---------------|---------------|
 | getIdentities | `reason`: String<br>`chainIds`: List{String} | `identities`: List{Identity} | Request available identities for the specified chains. The user can choose which of the matching identities they want to reveal to the client. |
-| signAndPost   | `reason`: String<br>`transaction`: UnsignedTransaction | `transactionId`: string | Request signing and posting a transaction prepared by the client. User is asked to confirm the signing after reviewing the transaction content. In case of successful post to the blockchain, the transaction ID is returned to the client to allow tracking the confirmation status. |
+| signAndPost   | `reason`: String<br>`transaction`: UnsignedTransaction | `transactionId`: String or `null` | Request signing and posting a transaction prepared by the client. User is asked to confirm the signing after reviewing the transaction content. In case of successful post to the blockchain, the transaction ID is returned to the client to allow tracking the confirmation status. When the request is denied, `null` is retruned |
 
 ## TODO
 
--- a/packages/iov-core/src/signingservercore.spec.ts
+++ b/packages/iov-core/src/signingservercore.spec.ts
@@ -59,16 +59,14 @@ describe("SigningServerCore", () => {
   });
 
   it("can get identities", async () => {
-    const chainId = "some-network" as ChainId;
-
     const profile = new UserProfile();
     const wallet = profile.addWallet(
       Ed25519HdWallet.fromMnemonic(
         "option diagram plastic million educate they arrow fat comic excite abandon green",
       ),
     );
-    const identity0 = await profile.createIdentity(wallet.id, chainId, HdPaths.iov(0));
-    const identity1 = await profile.createIdentity(wallet.id, chainId, HdPaths.iov(1));
+    const identity0 = await profile.createIdentity(wallet.id, defaultChainId, HdPaths.iov(0));
+    const identity1 = await profile.createIdentity(wallet.id, defaultChainId, HdPaths.iov(1));
 
     const signer = new MultiChainSigner(profile);
     const core = new SigningServerCore(
@@ -78,8 +76,8 @@ describe("SigningServerCore", () => {
       defaultSignAndPostCallback,
     );
 
-    const xnetIdentities = await core.getIdentities("Login to XY service", [chainId]);
-    expect(xnetIdentities).toEqual([identity0, identity1]);
+    const revealedIdentities = await core.getIdentities("Login to XY service", [defaultChainId]);
+    expect(revealedIdentities).toEqual([identity0, identity1]);
 
     core.shutdown();
   });
--- a/packages/iov-core/src/signingservercore.spec.ts
+++ b/packages/iov-core/src/signingservercore.spec.ts
@@ -223,7 +223,8 @@ describe("SigningServerCore", () => {
       recipient: await randomBnsAddress(),
     };
     const transactionId = await core.signAndPost("Please sign now", send);
-    expect(transactionId).toBeTruthy();
+    expect(transactionId).toBeDefined();
+    expect(transactionId).toMatch(/^[0-9A-F]{64}$/);
 
     core.shutdown();
   });
--- a/packages/iov-core/src/signingservercore.spec.ts
+++ b/packages/iov-core/src/signingservercore.spec.ts
@@ -7,6 +7,7 @@ import {
   PublicKeyBytes,
   SendTransaction,
   TokenTicker,
+  UnsignedTransaction,
 } from "@iov/bcp";
 import { bnsCodec, bnsConnector } from "@iov/bns";
 import { Ed25519, Random } from "@iov/crypto";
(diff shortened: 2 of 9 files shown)


packages/iov-core/src/signingservercore.ts
```diff
@@ -9,29 +9,57 @@ import { UserProfile, WalletId } from "@iov/keycontrol";
 
 import { MultiChainSigner } from "./multichainsigner";
 
+export interface GetIdentitiesAuthorization {
```
ethanfrey: I like this design, called after checking request against what we will show.

I would like to add a real identifier, like "session", so we can keep track of who is asking for what. But that is another issue.

packages/iov-core/src/signingservercore.ts
```diff
@@ -9,29 +9,57 @@ import { UserProfile, WalletId } from "@iov/keycontrol";
...
   ): Promise<ReadonlyArray<PublicIdentity>> {
     const matchingIdentities = this.allIdentities().filter(identity => {
       return chainIds.some(chainId => identity.chainId === chainId);
     });
 
-    // TODO: ask user for permission and allow selection of identities
+    const authorizedIdentities = this.authorizeGetIdentities(reason, matchingIdentities);
 
-    return matchingIdentities;
+    return authorizedIdentities;
```
ethanfrey: Why the const and not just return directly above?

packages/iov-core/src/signingservercore.ts
```diff
@@ -9,29 +9,57 @@ import { UserProfile, WalletId } from "@iov/keycontrol";
...
-    _: string,
+    reason: string,
     chainIds: ReadonlyArray<ChainId>,
   ): Promise<ReadonlyArray<PublicIdentity>> {
     const matchingIdentities = this.allIdentities().filter(identity => {
       return chainIds.some(chainId => identity.chainId === chainId);
     });
 
-    // TODO: ask user for permission and allow selection of identities
+    const authorizedIdentities = this.authorizeGetIdentities(reason, matchingIdentities);
```
ethanfrey: So, I assume we can return the whole list or a filtered list for success.

On failure, what is proper, return `[]` or throw and error (reject the promise)?

It would be good to add this to the docs above, what is expected from the callback

packages/iov-core/src/signingservercore.ts
```diff
@@ -49,10 +77,12 @@ export class SigningServerCore {
         throw new Error("More than one wallets contain the identity to sign this transaction");
     }
 
-    // TODO: ask user for permission
-
-    const response = await this.signer.signAndPost(transaction, walletId);
-    return response.transactionId;
+    if (this.authorizeSignAndPost(reason, transaction)) {
```
ethanfrey: Isn't this a promise??
Don't you want: `if (await this.authorizeSignAndPost(reason, transaction))`...

packages/iov-core/src/signingservercore.ts
```diff
@@ -49,10 +77,12 @@ export class SigningServerCore {
...
+      const response = await this.signer.signAndPost(transaction, walletId);
+      return response.transactionId;
+    } else {
+      return undefined;
```
ethanfrey: I prefer promise rejection on error, rather than return undefined.

DIfference between `Maybe<TransactionId>` and `Result<TransactionId, Error>` in functional languages

packages/iov-core/src/workers/signingservice.worker.ts
```diff
@@ -16,6 +16,10 @@ const bnsdFaucetMnemonic = "degree tackle suggest window test behind mesh extra
 const ethereumUrl = "http://localhost:8545";
 const ganacheMnemonic = "oxygen fall sure lava energy veteran enroll frown question detail include maximum";
 
+async function sleep(ms: number): Promise<void> {
```
ethanfrey: Isn't this function in some util package now? I've seen it a dozen places

packages/iov-core/src/signingservercore.spec.ts
```diff
@@ -41,11 +41,18 @@ describe("SigningServerCore", () => {
     fractionalDigits: 9,
     tokenTicker: "CASH" as TokenTicker,
   };
+  const defaultGetIdentitiesCallback: GetIdentitiesAuthorization = async (_, matching) => matching;
```
ethanfrey: This is good to show positive case.
Can you add tests for eg. filtering out all but one or returning an error?
Likewise for the SignAndPost callback with `true`, `false`, `throw Error`... or specify we should never throw an error if that is what you want.

It would be good to see how errors go across the rest boundary as well.

packages/iov-core/src/signingservercore.ts
webmaster128: No specific reason, both ways are fine. Having a variable name allowed me to express the meaning of the callback's return value

packages/iov-core/src/signingservercore.ts
webmaster128: Whatever the user decides is not an error. Right now, revealing 0 identities means either
- the user rejected the identities request
- the user deselected all available addresses and confirmed

I don't think there is a lot of value in having different return values for those cases.

> It would be good to add this to the docs above, what is expected from the callback

Yes, will improve

packages/iov-core/src/signingservercore.ts
webmaster128: Well spotted, this is a horrible bug!

Will add a test for the other case that finds this bug

packages/iov-core/src/workers/signingservice.worker.ts
webmaster128: Yes it exists in a dozen places across multiple packages and there is no package for that

packages/iov-core/src/signingservercore.ts
webmaster128: This is not an error but a decision made by the user, which must have one explicit handleable result that is transmitted via a JSON-RPC response.

Will improve documentation of this function as well

webmaster128:
Improved test coverage and docs. Everything starting from commit 'Add "can get identities" baseline test' is new.

> It would be good to see how errors go across the rest boundary as well.

I'm building a HTTP server implementation for this as an external project right now (as part of https://github.com/iov-one/iov-core/issues/400), so I'd keep this out of this PR.

packages/iov-core/src/signingservercore.ts
```diff
@@ -7,14 +7,28 @@ export interface GetIdentitiesAuthorization {
   /**
    * Called by the signing server and lets user decide which of the
    * available identities should be revealed to the application.
+   *
```
ethanfrey: Thank you for these docs. This is very helpful to clarify rejection

packages/iov-core/src/signingservercore.spec.ts
```diff
@@ -228,4 +229,41 @@ describe("SigningServerCore", () => {
...
+    const identities = await core.getIdentities("Please select signer", [bnsChain]);
+    const signingIdentity = identities[0];
+    const send: SendTransaction = {
+      kind: "bcp/send",
+      creator: signingIdentity,
+      amount: defaultAmount,
+      recipient: await randomBnsAddress(),
+    };
+    const transactionId = await core.signAndPost("Please sign now", send);
+    expect(transactionId).toBeUndefined();
```
ethanfrey: Okay, so we check for undefined in the client as rejection. An error would be like a failed network connection, right?

packages/iov-core/src/signingservercore.spec.ts
```diff
@@ -196,6 +196,37 @@ describe("SigningServerCore", () => {
...
+        throw new Error("Something broken in here!");
+      }
+
+      const signer = new MultiChainSigner(profile);
+      const core = new SigningServerCore(profile, signer, throwingCallback, defaultSignAndPostCallback);
+
+      await core
+        .getIdentities("Login to XY service", [defaultChainId])
+        .then(() => fail("must not resolve"))
+        .catch(error => expect(error).toMatch(/internal server error/i));
```
ethanfrey: So we guarantee an error that does not leak internal info of the server. This is good (we also had a long discussion of redacting errors in weave), but can make it hard to debug. It would be good if the server (separate project #400 right?) logs any errors, so it is possible at least for the server code to report them somehow, if we don't share them with the untrusted client.

docs/out-of-process-rpc.md
```diff
@@ -18,7 +19,7 @@ and can be implemented for example in JavaScript messages or deep-link URLs.
...
 |---------------|--------------------|---------------|---------------|
 | getIdentities | `reason`: String<br>`chainIds`: List{String} | `identities`: List{Identity} | Request available identities for the specified chains. The user can choose which of the matching identities they want to reveal to the client. |
-| signAndPost   | `reason`: String<br>`transaction`: UnsignedTransaction | `transactionId`: string | Request signing and posting a transaction prepared by the client. User is asked to confirm the signing after reviewing the transaction content. In case of successful post to the blockchain, the transaction ID is returned to the client to allow tracking the confirmation status. |
+| signAndPost   | `reason`: String<br>`transaction`: UnsignedTransaction | `transactionId`: String or `null` | Request signing and posting a transaction prepared by the client. User is asked to confirm the signing after reviewing the transaction content. In case of successful post to the blockchain, the transaction ID is returned to the client to allow tracking the confirmation status. When the request is denied, `null` is retruned |
```
ethanfrey: Okay, null not undefined now. I guess this makes it clearer, as undefined can be a forgotten return value, while null must be explicitly returned

packages/iov-core/src/signingservercore.spec.ts
webmaster128: I think the signing server (i.e. the browser extension) should log those errors. That way we have a clear separation of internal and external and a developer can always check the browser extension's logs for debugging.

docs/out-of-process-rpc.md
webmaster128: Good question. These docs are programming language agnostic so `null` is `nil` in Swift, `nullptr` in C++, `nil` in Golang, `null` or `undefined` in JavaScript. But yeah, `null` is probably the better option in case of JSON.

packages/iov-core/src/signingservercore.spec.ts
webmaster128: Right

PR #794, opened 2019-02-28
Replace bnsNonceTag with BcpTxQuery.signedBy

### LINKED ISSUE #714: Convert bnsNonceTag into BcpTxQuery.signedBy
This becomes important as soon as other transaction types then bcp/send are supported for non-BNS chains.

Interface:

```ts
export interface BcpTxQuery {
  readonly id?: TransactionId;
  /** send transaction to or from this address */
  readonly sentFromOrTo?: Address;
  /** any transaction signed by this address */
  readonly signedBy?: Address;
  /** chain-specific key value pairs that encode a query */
  readonly tags?: ReadonlyArray<BcpQueryTag>;
  readonly height?: number;
  readonly minHeight?: number;
  readonly maxHeight?: number;
}
```

willclarktech:
Closes #714 
Not 100% the errors are thrown in the right places or with the right messages, but seemed like the most appropriate to me.

### DIFF
--- a/packages/iov-bcp/src/connection.ts
+++ b/packages/iov-bcp/src/connection.ts
@@ -168,8 +168,10 @@ export interface BcpQueryTag {
 
 export interface BcpTxQuery {
   readonly id?: TransactionId;
-  /** send transaction to or from this address */
+  /** any transaction sent to or from this address */
   readonly sentFromOrTo?: Address;
+  /** any transaction signed by this address */
+  readonly signedBy?: Address;
   /** chain-specific key value pairs that encode a query */
   readonly tags?: ReadonlyArray<BcpQueryTag>;
   readonly height?: number;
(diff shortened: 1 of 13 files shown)


packages/iov-bcp/src/connection.ts
```diff
@@ -168,8 +168,10 @@ export interface BcpQueryTag {
 
 export interface BcpTxQuery {
   readonly id?: TransactionId;
-  /** send transaction to or from this address */
+  /** any transaction sent to or from this address */
```
webmaster128: Hmm, does this change help? What I wanted to say is:

> any SendTransaction with this address as the recipient or the sender

willclarktech: Ah yeah, I changed this before we talked about the transaction types. I'll adjust.

Issue #652, opened 2019-01-02
Make PublicIdentity unique in UserProfile/Keyring


webmaster128:
Many methods of UserProfile require a `id: WalletId` argument because it is not guaranteed that a PublicIdentity is unique. Now that PublicIdentity contains a chain ID it is reasonable to enforce that a PublicIdentity is not part of multiple wallets.

PR #797, opened 2019-03-04
Create ReadonlyWallet

webmaster128:
First step of #652

With this change, all wallet mutation goes through

1. `Keyring.add(wallet: Wallet)`
2. `Keyring.setWalletLabel(walletId: WalletId, label: string | undefined)`
3. `Keyring.createIdentity(walletId: WalletId, chainId: ChainId, options: Ed25519Keypair | ReadonlyArray<Slip10RawIndex> | number)`
4. `Keyring.setIdentityLabel(walletId: WalletId, identity: PublicIdentity, label: string | undefined)`

Given that we know that only 1. and 3. change the keyring's identities. So we have only 2 places where we need to ensure an identity is unique within a Keyring.

### DIFF
--- a/packages/iov-keycontrol/src/wallet.ts
+++ b/packages/iov-keycontrol/src/wallet.ts
@@ -8,44 +8,14 @@ export type WalletId = string & As<"wallet-id">;
 export type WalletImplementationIdString = string & As<"wallet-implementation-id">;
 export type WalletSerializationString = string & As<"wallet-serialization">;
 
-/**
- * A is a generic interface for managing a set of keys and signing
- * data with them.
- *
- * A Wallet is responsible for storing private keys securely and
- * signing with them. Wallet can be implemented in software or as
- * a bridge to a hardware wallet.
- */
-export interface Wallet {
+/** An interface that does not allow mutating any internal state of the Wallet */
+export interface ReadonlyWallet {
   readonly label: ValueAndUpdates<string | undefined>;
 
   // id is a unique identifier based on the content of the keyring
   // the same implementation with same seed/secret should have same identifier
   // otherwise, they will be different
   readonly id: WalletId;
-
-  /**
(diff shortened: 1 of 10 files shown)


packages/iov-keycontrol/src/wallet.ts
```diff
@@ -94,3 +64,36 @@ export interface Wallet {
 
   readonly clone: () => Wallet;
```
ethanfrey: Clone a ReadonlyWallet gives you a mutable Wallet?
Is that safe?
Shall it return a ReadonlyWallet instead?

packages/iov-keycontrol/src/wallet.ts
```diff
@@ -94,3 +64,36 @@ export interface Wallet {
...
+
+/**
+ * A is a generic interface for managing a set of keys and signing
+ * data with them.
+ *
+ * A Wallet is responsible for storing private keys securely and
+ * signing with them. Wallet can be implemented in software or as
+ * a bridge to a hardware wallet.
+ */
+export interface Wallet extends ReadonlyWallet {
```
ethanfrey: I like this separation by the way, to make it easier to reason about everything

packages/iov-keycontrol/src/wallet.ts
webmaster128: Yes, because cloning the ReadonlyWallet does not change the original. So this is a way to make a mutable copy of the immutable object.

This property is required when adding a wallet to the UserProfile:

```
  public addWallet(wallet: Wallet): WalletInfo {
    if (!this.keyring) {
      throw new Error("UserProfile is currently locked");
    }

    const copy = wallet.clone();
    const info = this.keyring.add(copy);
    this.walletsProducer.update(this.walletInfos());
    return info;
  }
```

If the `copy` was immutable, the keyring could not perform the required changes on it.

PR #798, opened 2019-03-04
Create send transaction test for MultiChainSigner

### LINKED ISSUE #769: signAndPost does not propagate error
When adding ethereum to wallet-demo, `signAndPost` does not get "throw new Error("gasPrice must be set");" from `signTransaction`. So there is a missed error propagation, hard to notice when something is missed.

webmaster128:
Using this test and commenting out the definition of gasPrice or gasLimit, I could verify the error propagation works.

Closes #769

### DIFF
--- a/packages/iov-core/src/multichainsigner.spec.ts
+++ b/packages/iov-core/src/multichainsigner.spec.ts
@@ -9,12 +9,13 @@ import {
   PublicKeyBytes,
   SendTransaction,
   TokenTicker,
+  TransactionState,
 } from "@iov/bcp";
 import { bnsCodec, bnsConnector } from "@iov/bns";
 import { Ed25519, Random } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
 import { ethereumConnector } from "@iov/ethereum";
-import { Ed25519HdWallet, HdPaths, UserProfile, WalletId } from "@iov/keycontrol";
+import { Ed25519HdWallet, HdPaths, Secp256k1HdWallet, UserProfile, WalletId } from "@iov/keycontrol";
 
 import { MultiChainSigner } from "./multichainsigner";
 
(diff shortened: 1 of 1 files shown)


packages/iov-core/src/multichainsigner.spec.ts
```diff
@@ -173,6 +174,80 @@ describe("MultiChainSigner", () => {
...
+    it("can add two chains and send on both of them", async () => {
+      pendingWithoutBnsd();
+      pendingWithoutEthereum();
+
+      const profile = new UserProfile();
+      const signer = new MultiChainSigner(profile);
+      expect(signer.chainIds().length).toEqual(0);
+
+      await signer.addChain(bnsConnector(bnsdTendermintUrl));
+      await signer.addChain(ethereumConnector(httpEthereumUrl, undefined));
```
willclarktech: This is a bit strange. It would be more idiomatic if the type was `null`able than `undefined`able.

packages/iov-core/src/multichainsigner.spec.ts
```diff
@@ -173,6 +174,80 @@ describe("MultiChainSigner", () => {
...
+          "oxygen fall sure lava energy veteran enroll frown question detail include maximum",
+        ),
+      );
+      const ganacheMainIdentity = await profile.createIdentity(
+        secpWallet.id,
+        ethereumChainId,
+        HdPaths.ethereum(0),
+      );
+
+      {
```
willclarktech: Are these blocks just for visual delimitation or is there some other purpose?

packages/iov-core/src/multichainsigner.spec.ts
```diff
@@ -173,6 +174,80 @@ describe("MultiChainSigner", () => {
...
+        // Send on BNS
+        const sendOnBns: SendTransaction = {
+          kind: "bcp/send",
+          creator: bnsFaucet,
+          recipient: await randomBnsAddress(),
+          memo: `MultiChainSigner style (${Math.random()})`,
```
willclarktech: Is there a good reason to use a random value in this test?

packages/iov-core/src/multichainsigner.spec.ts
webmaster128: Oh the debate over the two versions of null in JS are probably going to live longer than us :) In all the places where we have TS support, I guess both ways are probably equally good.

I have no good reasons for one or the other and use `undefined` by default

packages/iov-core/src/multichainsigner.spec.ts
webmaster128: Visual and for scoping of the variables. I tend to do it when I want to make sure the variables are not accessed anymore afterwards and to be able to reuse the varable names

packages/iov-core/src/multichainsigner.spec.ts
webmaster128: No. It was copied from the test above where we use a random memo to ensure we find the same transaction afterwards using searchTx. But in this case we don't do that.

webmaster128:
> Should we have tests for error propagation though?

I think can take error propagation for granted. Otherwise we'd need to create thousands of tests verifying that all high level interfaces propagate all possible lower level errors

packages/iov-core/src/multichainsigner.spec.ts
willclarktech: The rule I go by is something like: `undefined` is where you don't have knowledge of the value, `null` is where you know there is no value. I find it a helpful distinction, didn't realise it was controversial!

PR #806, opened 2019-03-06
Make identities unique in Keyring

webmaster128:
Closes #652

With this merged, we know that an identity (`PublicIdentity`) is unique within a Keyring. Since there is only one keyring in a `UserProfile`, we know that identities are unique within a profile. This allows us to drop the wallet ID argument in

* Keyring.setIdentityLabel
* UserProfile.setIdentityLabel
* UserProfile.getIdentityLabel
* UserProfile.signTransaction
* UserProfile.appendSignature
* MultiChainSigner.signAndPost

PR #808, opened 2019-03-06
Remove identity argument from UserProfile.signTransaction

### LINKED ISSUE #641: Remove identity arguments from UserProfile.signTransaction
The identity we pass must match the transaction creator. Thus the data is redundant and the `identity: PublicIdentity` argument can be removed.

Current interface

```ts
  public async signTransaction(
    id: WalletId,
    identity: PublicIdentity,
    transaction: UnsignedTransaction,
    codec: TxCodec,
    nonce: Nonce,
  ): Promise<SignedTransaction>
```

webmaster128:
Based on #806

Closes #641

### DIFF
--- a/packages/iov-keycontrol/src/userprofile.ts
+++ b/packages/iov-keycontrol/src/userprofile.ts
@@ -7,7 +7,6 @@ import {
   FullSignature,
   Nonce,
   PublicIdentity,
-  publicIdentityEquals,
   SignedTransaction,
   TxCodec,
   UnsignedTransaction,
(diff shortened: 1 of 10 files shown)


ethanfrey:
I was excited to see the walletId removed, then realized it was the identity.
Yes, this is a duplicate of info already in the transaction, so good change.

Waiting for walletId to disappear from the arg list too :crossed_fingers:

webmaster128:
Rebased and ready for reaview

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -192,27 +191,23 @@ export class UserProfile {
...
   public async signTransaction(
     id: WalletId,
-    identity: PublicIdentity,
     transaction: UnsignedTransaction,
     codec: TxCodec,
     nonce: Nonce,
   ): Promise<SignedTransaction> {
-    // identity and the pair (transaction.chainId, transaction.signer) are redundant
-    // but we keep both in the interface to be consistent with appendSignature() where
-    // the original transaction creator is not the signer
```
willclarktech: Why exactly is this no longer valid?

webmaster128: We decided that it is not important to keep the signature of signTransaction() and appendSignature() the same. In the simple cases (signTransaction), the transaction creator is always the one who signs the transaction. This saves us an additional, redundant argument.

PR #823, opened 2019-03-06
Add `BcpConnection.getFeeQuote`

### LINKED ISSUE #766: Query fees per transaction type and blockchain in BCP

willclarktech:
Closes #766 

Opening a draft PR to check this is going in the right direction before turning to the more complicated Ethereum case.

### DIFF
--- a/packages/iov-bcp/src/transactions.ts
+++ b/packages/iov-bcp/src/transactions.ts
@@ -164,6 +164,13 @@ export function isAmount(data: any): data is Amount {
   );
 }
 
+/** A general interface for blockchain fees */
+export interface Fee {
+  readonly tokens?: Amount;
+  readonly gasPrice?: Amount;
+  readonly gasLimit?: Amount;
+}
+
 /** The basic transaction type all transactions should extend */
 export interface UnsignedTransaction {
   /**
@@ -184,21 +191,21 @@ export interface UnsignedTransaction {
    * This implicitly fixes the chain ID this transaction can be used on.
    */
   readonly creator: PublicIdentity;
-  readonly fee?: Amount;
-  readonly gasPrice?: Amount;
-  readonly gasLimit?: Amount;
+  readonly fee?: Fee;
 }
 
 export function isUnsignedTransaction(data: any): data is UnsignedTransaction {
+  const isObject = typeof data === "object" && data !== null;
+  if (!isObject) {
+    return false;
+  }
+  const tx = data as UnsignedTransaction;
   return (
-    typeof data === "object" &&
-    data !== null &&
-    typeof (data as UnsignedTransaction).kind === "string" &&
-    isPublicIdentity((data as UnsignedTransaction).creator) &&
-    ((data as UnsignedTransaction).fee === undefined || isAmount((data as UnsignedTransaction).fee)) &&
-    ((data as UnsignedTransaction).gasPrice === undefined ||
-      isAmount((data as UnsignedTransaction).gasPrice)) &&
-    ((data as UnsignedTransaction).gasLimit === undefined || isAmount((data as UnsignedTransaction).gasLimit))
+    typeof tx.kind === "string" &&
+    isPublicIdentity(tx.creator) &&
+    (tx.fee === undefined ||
+      isAmount(tx.fee.tokens) ||
+      (isAmount(tx.fee.gasPrice) && isAmount(tx.fee.gasLimit)))
   );
 }
 
--- a/packages/iov-bcp/src/transactions.ts
+++ b/packages/iov-bcp/src/transactions.ts
@@ -171,6 +171,14 @@ export interface Fee {
   readonly gasLimit?: Amount;
 }
 
+export function isFee(data: any): data is Fee {
+  return (
+    typeof data === "object" &&
+    data !== null &&
+    (isAmount(data.tokens) || (isAmount(data.gasPrice) && isAmount(data.gasLimit)))
+  );
+}
+
 /** The basic transaction type all transactions should extend */
 export interface UnsignedTransaction {
   /**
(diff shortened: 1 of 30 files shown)


ethanfrey:
One note, we have completely overhauled the implementation of fees im Bnsd 0.12, not yet released. Nothing to do here, but just a note to keep in mind

packages/iov-bcp/src/transactions.ts
```diff
@@ -164,6 +164,13 @@ export function isAmount(data: any): data is Amount {
   );
 }
 
+/** A general interface for blockchain fees */
+export interface Fee {
```
webmaster128: Yeah, nice, super simple

packages/iov-bcp/src/transactions.ts
```diff
@@ -184,21 +191,21 @@ export interface UnsignedTransaction {
...
-    isPublicIdentity((data as UnsignedTransaction).creator) &&
-    ((data as UnsignedTransaction).fee === undefined || isAmount((data as UnsignedTransaction).fee)) &&
-    ((data as UnsignedTransaction).gasPrice === undefined ||
-      isAmount((data as UnsignedTransaction).gasPrice)) &&
-    ((data as UnsignedTransaction).gasLimit === undefined || isAmount((data as UnsignedTransaction).gasLimit))
+    typeof tx.kind === "string" &&
+    isPublicIdentity(tx.creator) &&
+    (tx.fee === undefined ||
+      isAmount(tx.fee.tokens) ||
+      (isAmount(tx.fee.gasPrice) && isAmount(tx.fee.gasLimit)))
```
webmaster128: This logic is good. Could be extracted into an isFee function very similar to isAmount

packages/iov-bcp/types/connection.d.ts
```diff
@@ -194,4 +194,5 @@ export interface BcpConnection {
      * the query, along with all new transactions arriving from listenTx
      */
     readonly liveTx: (txQuery: BcpTxQuery) => Stream<ConfirmedTransaction | FailedTransaction>;
+    readonly getFeeQuote: (tx: UnsignedTransaction) => Promise<Fee | null>;
```
webmaster128: Good choice using a promise. This allows us to query data from the network, like this one: https://web3js.readthedocs.io/en/1.0/web3-eth.html#getgasprice

However, I don't think we need the `null` case. In case of BNS, we can return a zero amount or even better, change to a small anti-spam fee

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -619,6 +619,10 @@ export class BnsConnection implements BcpAtomicSwapConnection {
     return nfts;
   }
 
+  public async getFeeQuote(): Promise<null> {
+    return null;
```
webmaster128: We can use

```
         tokens: {
            quantity: "0",
            fractionalDigits: 9,
            tokenTicker: "IOV" as TokenTicker,
          },
```

However, not all chains that use the BnsConnection have an IOV token. So the question is: in which currency do I have to pay the fees? Let's discuss with backend team

packages/iov-bns/src/testdata.spec.ts
```diff
@@ -155,9 +155,11 @@ const randomMsg: SendTransaction = {
     tokenTicker: "FOO" as TokenTicker,
   },
   fee: {
-    quantity: "5432",
-    fractionalDigits: 9,
-    tokenTicker: "PSQL" as TokenTicker,
+    tokens: {
+      quantity: "5432",
+      fractionalDigits: 9,
+      tokenTicker: "PSQL" as TokenTicker,
```
webmaster128: Haha, that was most likely Isabella and her beloved Postgres port number :)

packages/iov-ethereum/src/serialization.spec.ts
```diff
@@ -147,10 +153,12 @@ describe("Serialization", () => {
...
           kind: "bcp/send",
-          creator: creator,
-          amount: amount,
-          gasPrice: gasPrice,
-          gasLimit: undefined,
+          creator,
+          amount,
+          fee: {
+            gasPrice,
+            gasLimit: undefined,
```
webmaster128: I'm not really a fan of those shortcut notations where property name equals variable name. I find it very hard to parse in the case where some properties are key/value and some are variable only. Also it is a strange mix of concerns: the local variable name on the one hand, that should be changeable at any point in time and the external property name on the other hand that must not be changed.

So if you don't have a strong opinion here, I'd like to stick with explicit property names. We can also use tslint to enforce consistency.

packages/iov-lisk/src/liskconnection.ts
```diff
@@ -396,6 +397,16 @@ export class LiskConnection implements BcpConnection {
     }
   }
 
+  public async getFeeQuote(): Promise<Fee> {
+    return {
```
webmaster128: I think we should wrap that in a transaction type switch (see `function buildMsg(tx: UnsignedTransaction): codecImpl.app.ITx` in @iov/bns) and default to an error, such that we don't return the wrong fee accidentally when new transactions types are added.

packages/iov-rise/src/riseconnection.ts
```diff
@@ -383,6 +384,16 @@ export class RiseConnection implements BcpConnection {
     }
   }
 
+  public async getFeeQuote(): Promise<Fee> {
+    return {
```
webmaster128: same here: switch over transaction type

packages/iov-ethereum/src/serialization.spec.ts
willclarktech: I like them because they can reduce reading quite a lot, but the most important thing for me is consistency. I suggest a separate issue to add a tslint rule.

willclarktech:
This PR leaves out:
- BNS fees (since they are changing in 0.12)
- Ethereum fees (since hopefully we'll be in a better position to implement this after next week's workshop)
- object shorthand notation/tslint rule

packages/iov-bcp/src/transactions.ts
```diff
@@ -164,6 +164,20 @@ export function isAmount(data: any): data is Amount {
...
+export interface Fee {
+  readonly tokens?: Amount;
+  readonly gasPrice?: Amount;
+  readonly gasLimit?: Amount;
+}
+
+export function isFee(data: any): data is Fee {
+  return (
+    (typeof data === "object" && data !== null && isAmount(data.tokens)) ||
+    (isAmount(data.gasPrice) && isAmount(data.gasLimit))
```
webmaster128: the way this is grouped now should lead to a crash for `data` that is no object, since `data.gasPrice` is referenced after the first line returns false.

willclarktech: Oh! Not sure how this happened, will fix.

webmaster128: Most likely the currect CI fail is because of this

PR #837, opened 2019-03-18
Rename SwapTimeoutTransaction -> SwapAbortTransaction

webmaster128:


packages/iov-bns/src/decode.ts
```diff
@@ -224,13 +224,13 @@ function parseSwapClaimTx(
   };
 }
 
-function parseSwapTimeoutTx(
+function parseSwapAbortTransaction(
```
willclarktech: Do we have any rules about when to use `tx` and when to use `transaction`? This change is not consistent with how it was before or the `parseSwapClaimTx` function.

webmaster128: My idea is to ban the term "tx" in all non-Tendermint code. It seems to be very popular in the Tendermint community but not necessarily in all blockchain world. This is why in BCP everything is *transaction but there is still a lot of legacy.

The whole naming build* and parse* in encode.ts/decode.ts is inconsitent. It should probably be "encode<transaction type>" and "decode<transaction type>". I started this renaming as part of this PR but it became too much so I stopped the internal renamings to focus on external facing changes only.

willclarktech: OK, I'm happy to incrementally switch to `transaction` only.

webmaster128:
> I find a few more occurrences

thanks, renamed

ethanfrey:
If you are complaining about my internal function names, it means, you have tackled all my code.

Please feel free to rename it all. I just care that the functionality works with good test coverage. You two are much better at naming than I

webmaster128:
No problem, we'll do Ethan :) It is much easier to cleanup existing functionality as clarity is improving than coming up with all that from scratch!

PR #838, opened 2019-03-18
Ethereum updates (IOV-Core patch 0.12.3)

### LINKED ISSUE #833: Export toChecksumAddress from @iov/ethereum

### LINKED ISSUE #834: Add missing @iov/socket dependency to @iov/ethereum

webmaster128:
Closes #834 
Closes #833

Also adds a nonces/ folder for each package to ensure releases are made for every package


packages/iov-ethereum/src/address.spec.ts
```diff
@@ -47,28 +47,58 @@ describe("derivation", () => {
...
       );
     });
+
+    it("does not change checksummed addresses", () => {
+      expect(toChecksummedAddress("0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359")).toEqual(
+        "0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359",
+      );
+    });
+
+    it("throws for invalud input addresses", () => {
```
willclarktech: `invalid`

packages/iov-ethereum/src/address.spec.ts
```diff
@@ -47,28 +47,58 @@ describe("derivation", () => {
...
+      // invalid length
+      expect(() => toChecksummedAddress("0x")).toThrowError(/not a valid Ethereum address/i);
+      expect(() => toChecksummedAddress("0xaa")).toThrowError(/not a valid Ethereum address/i);
+      expect(() => toChecksummedAddress("0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")).toThrowError(
+        /not a valid Ethereum address/i,
+      );
+
+      // incorrectly checksummed
+      expect(() => toChecksummedAddress("0xFB6916095ca1df60bB79Ce92cE3Ea74c37c5d359")).toThrowError(
+        /not a valid Ethereum address/i,
```
willclarktech: Shouldn't we have a more informative error message for this case?

packages/iov-ethereum/src/address.ts
```diff
@@ -28,9 +28,18 @@ export function isValidAddress(address: string): boolean {
   }
 }
 
-export function toChecksumAddress(address: string): Address {
+/**
+ * Converts Ethereum adddress to checksummed address according to EIP-55.
```
willclarktech: Would be nice to have the link here instead of below.

willclarktech: `address` spelling

scripts/add_nonce.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# This must be called from repo root as current directory
+
+NONCE=$(date +%s)
+
+for packagedir in ./packages/*; do
+  if [[ -d "$packagedir" ]]; then
+    echo "$packagedir"
+    mkdir -p "$packagedir/nonces"
+    echo "Directory used to trigger lerna package updates for all packages" \
```
willclarktech: I guess we can remove this now.

scripts/add_nonce.sh
```diff
@@ -0,0 +1,17 @@
...
+      > "$packagedir/nonces/README.txt"
+    touch "$packagedir/nonces/$NONCE"
```
willclarktech: Do we also want to remove the old nonce(s)? Alternatively we could just have a file which changes.

scripts/add_nonce.sh
```diff
@@ -0,0 +1,17 @@
...
+set -o errexit -o nounset -o pipefail
+command -v shellcheck > /dev/null && shellcheck "$0"
+
+# This must be called from repo root as current directory
+
+NONCE=$(date +%s)
+
+for packagedir in ./packages/*; do
+  if [[ -d "$packagedir" ]]; then
+    echo "$packagedir"
```
willclarktech: Would be nice to expand a little. Suggest `"Added nonce to $packagedir" and putting it at the end of the loop.

scripts/add_nonce.sh
webmaster128: The command can remain included to cover new packages or README updates as well

scripts/add_nonce.sh
webmaster128: The idea here is that multiple files in a directory can be merged automatically. This saves time when merging e.g. the branch 0.12 into master when there are 0.12.x releases (and potentially 0.13.x on master)

scripts/add_nonce.sh
webmaster128: Good point, this is an old debugging log

packages/iov-ethereum/src/address.spec.ts
webmaster128: Hmm, this is already way better than what we had before and we don't have better data available easily (since isValidAddress returns boolean only). So I'd keep it simple until we need something better.

webmaster128:
updated, +3 commits

PR #840, opened 2019-03-18
Convert Nonce to number

### LINKED ISSUE #799: Convert all implementations of UnsignedTransaction to simple types
In the first step, all types must be in the [List of supported types of the structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#Supported_types). Later we need to be able to JSON encode them, so prefer simple types over complex types.

As far as I can see this only affects the type `Nonce`, which is a complex class at the moment. Nonce can be converted to a number with the restriction that it must be in the safe integer range.

webmaster128:
Closes #799

### DIFF
--- a/packages/iov-lisk/src/liskcodec.ts
+++ b/packages/iov-lisk/src/liskcodec.ts
@@ -20,7 +20,7 @@ import {
   UnsignedTransaction,
 } from "@iov/bcp";
 import { Parse, Serialization } from "@iov/dpos";
-import { Encoding } from "@iov/encoding";
+import { Encoding, Int53 } from "@iov/encoding";
 
 import { constants } from "./constants";
 import { isValidAddress, pubkeyToAddress } from "./derivation";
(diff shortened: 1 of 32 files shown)


packages/iov-lisk/src/liskcodec.ts
```diff
@@ -31,8 +31,8 @@ export const liskCodec: TxCodec = {
    * https://github.com/prolina-foundation/snapshot-validator/blob/35621c7/src/transaction.cpp#L36
    */
   bytesToSign: (unsigned: UnsignedTransaction, nonce: Nonce): SigningJob => {
-    const creationTimestamp = nonce.toNumber();
-    const creationDate = new ReadonlyDate(creationTimestamp * 1000);
+    const creationTimestamp = new Int53(nonce);
+    const creationDate = new ReadonlyDate(creationTimestamp.toNumber() * 1000);
```
willclarktech: Can we leave the safe range here (and below)?

webmaster128: Yes, we can, good point. Unfortunatly as far as I know, we cannot construct a JS Date using seconds directly.

But this can only happen for extreme times that are not in the Lisk time range (Lisk epoch plus int32)

PR #841, opened 2019-03-18
Implement `getFeeQuote` method on `BnsConnection` 

webmaster128:


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -620,8 +622,19 @@ export class BnsConnection implements BcpAtomicSwapConnection {
     return nfts;
   }
 
-  public async getFeeQuote(): Promise<Fee> {
-    throw new Error("Not implemented");
+  public async getFeeQuote(transaction: UnsignedTransaction): Promise<Fee> {
```
ethanfrey: Yes, good enough filler for now.
I will add real logic here when I add weave 0.12 support

webmaster128: Jap. This current explicit 0 fee logic should work for the wallet/hugnet and shows the path for future changes.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1907,4 +1908,85 @@ describe("BnsConnection", () => {
 
     connection.disconnect();
   });
+
+  describe("getFeeQuote", () => {
+    it("works for send transaction", async () => {
+      pendingWithoutBnsd();
+      const connection = await BnsConnection.establish(bnsdTendermintUrl);
```
willclarktech: Might want to consider putting this connection stuff in a `beforeEach`/`afterEach` hook since it's used so often.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1907,4 +1908,85 @@ describe("BnsConnection", () => {
...
+          pubkey: {
+            algo: Algorithm.Ed25519,
+            data: fromHex("aabbccdd") as PublicKeyBytes,
+          },
+        },
+      };
+      await connection
+        .getFeeQuote(otherTransaction)
+        .then(() => fail("must not resolve"))
+        .catch(error => expect(error).toMatch(/transaction of unsupported kind/i));
```
willclarktech: Is there something similar to `chai-as-promised` we can use?

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Yeah, right, this is a little annoying. In our experience, pending() does not work properly with beforeEach(). Also this adds some top level mutable state, which is not great as well.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: The only thing I am aware of is this issue: https://github.com/jasmine/jasmine/issues/1625

Looks like strict error message equality can be archived but not the looser checks that allow slight variation of the implementation.

packages/iov-bns/src/bnsconnection.spec.ts
willclarktech: I've pretty much accepted top level mutable state as a necessary evil for concise mocha tests and it seems the same in jasmine. The pending/beforeEach conflict's annoying though.

willclarktech: The `beforeEach` hook shouldn't run for pending tests though, right? In that case we could at least move the connection stuff out of the test.

Another slightly weirder thought: conditionally call `it()` or `it.skip()` depending on whether the test should be executed. (Probably involves a lot of ugly bracketing.)

webmaster128: pendingWithoutBnsd can only know at runtime if the test is pending. So beforeEach is called in any case. However, a [call to pending in beforeEach does not stop the test execution](https://github.com/jasmine/jasmine/issues/1579).

Using `xit` instead of `it` does not allow to secify a pending reason: https://jasmine.github.io/api/edge/global.html#xit

ethanfrey:
Do I smell some PRs to jasmine coming up?

webmaster128:
> Do I smell some PRs to jasmine coming up?

Unfortunatly I'm not so optimistic about the room for improvement

PR #842, opened 2019-03-19
Convert signing server from JSON interface to JS interface

### LINKED ISSUE #800: Rename JsonRpcSigningServer to JsSigningServer
Requires #799

The current JsonRpcSigningServer misses the functionality of JSON encoding/decoding. Since we do not strictly need that right now, we can vonvert it to a JsSigningServer that takes UnsignedTransaction as a JS object.

webmaster128:
Closes #800

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * @iov/keycontrol: Add `Keyring.getAllIdentities` and `Keyring.getWalletByIdentity`
 * @iov/keycontrol: Add `UserProfile.getAllIdentities`
 * @iov/keycontrol: Add `.previewIdentity` to the `ReadonlyWallet`/`Wallet` interfaces
+* @iov/jsonrpc: Let `isJsonCompatibleDictionary` not accept non-simple objects
 * @iov/lisk: Implement `watchBlockHeaders` method on `LiskConnection`.
 * @iov/lisk: Implement `getFeeQuote` method on `LiskConnection`.
 * @iov/rise: Implement `getFeeQuote` method on `RiseConnection`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Breaking changes
 * @iov/encoding: Remove `Uint32.asNumber`. Use `Uint32.toNumber` instead.
 * @iov/ethereum: The `options` parameter of `ethereumConnector` does not allow
   strings anymore. Use `{ wsUrl: myWebsocketUrl }` instead.
+* @iov/jsonrpc: Make request and response types generic in `SimpleMessagingConnection`.
 * @iov/keycontrol: Remove `HdPaths.metamaskHdKeyTree`. Use `HdPaths.ethereum` instead.
 * @iov/keycontrol: `Keyring.getWallet` and `Keyring.getWallets` return the
   immutable type ReadonlyWallet now. New functions to mutate wallets are added:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ Breaking changes
   JavaScript and JSON interfaces.
 * @iov/bns: Remove `bnsNonceTag` in favour of `BcpTxQuery.signedBy`.
 * @iov/core: Remove wallet ID argument from `MultiChainSigner.signAndPost`
+* @iov/core: Rename `JsonRpcSigningServer` to `JsRpcSigningServer` and convert
+  interface from JSON to JavaScript.
 * @iov/encoding: Remove `Uint32.asNumber`. Use `Uint32.toNumber` instead.
 * @iov/ethereum: The `options` parameter of `ethereumConnector` does not allow
   strings anymore. Use `{ wsUrl: myWebsocketUrl }` instead.
--- a/packages/iov-core/src/jsrpc.ts
+++ b/packages/iov-core/src/jsrpc.ts
@@ -63,14 +63,9 @@ export function isJsRpcCompatibleDictionary(data: unknown): data is JsRpcCompati
     return false;
   }
 
-  for (const key of Object.getOwnPropertyNames(data)) {
-    const value = (data as any)[key];
-    if (!isJsRpcCompatibleValue(value)) {
-      return false;
-    }
-  }
-
-  return true;
+  // replace with Object.values when available (ES2017+)
+  const values = Object.getOwnPropertyNames(data).map(key => (data as any)[key]);
+  return values.every(isJsRpcCompatibleValue);
 }
 
 export interface JsRpcRequest {
--- a/packages/iov-core/src/jsrpc.ts
+++ b/packages/iov-core/src/jsrpc.ts
@@ -1,0 +1,242 @@
+import {
+  jsonRpcCodeInternalError,
+  jsonRpcCodeInvalidParams,
+  jsonRpcCodeInvalidRequest,
+  jsonRpcCodeMethodNotFound,
+  jsonRpcCodeParseError,
+  jsonRpcCodeServerErrorDefault,
+  SimpleMessagingConnection,
+} from "@iov/jsonrpc";
+import { firstEvent } from "@iov/stream";
+
+// An RPC interface very similar to JSON-RPC but supporting
+// all the data types from the structured clone algorithm
+// https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#Supported_types
+
+/** A single structured clone algorithm compatible value */
+export declare type JsRpcCompatibleValue =
+  | string
+  | number
+  | boolean
+  | null
+  | Uint8Array
+  | JsRpcCompatibleArray
+  | JsRpcCompatibleDictionary;
+
+/** An array of JsRpcCompatibleValue */
+export interface JsRpcCompatibleArray extends ReadonlyArray<JsRpcCompatibleValue> {}
+
+/** A string to JsRpcCompatibleValue dictionary. */
+export interface JsRpcCompatibleDictionary {
+  readonly [key: string]: JsRpcCompatibleValue;
+}
+
+export function isJsRpcCompatibleValue(value: unknown): value is JsRpcCompatibleValue {
+  return (
+    typeof value === "string" ||
+    typeof value === "number" ||
+    typeof value === "boolean" ||
+    value === null ||
+    value instanceof Uint8Array ||
+    isJsRpcCompatibleArray(value) ||
+    isJsRpcCompatibleDictionary(value)
+  );
+}
+
+export function isJsRpcCompatibleArray(value: unknown): value is JsRpcCompatibleArray {
+  if (!Array.isArray(value)) {
+    return false;
+  }
+  return value.every(isJsRpcCompatibleValue);
+}
+
+export function isJsRpcCompatibleDictionary(data: unknown): data is JsRpcCompatibleDictionary {
+  if (typeof data !== "object" || data === null) {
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * @iov/keycontrol: Add `Keyring.getAllIdentities` and `Keyring.getWalletByIdentity`
 * @iov/keycontrol: Add `UserProfile.getAllIdentities`
 * @iov/keycontrol: Add `.previewIdentity` to the `ReadonlyWallet`/`Wallet` interfaces
+* @iov/jsonrpc: Let `isJsonCompatibleDictionary` not accept non-simple objects
```
ethanfrey: Huh? Not really clear what this is until diving into the issues.

ethanfrey: Looking at tests, I assume you mean you accept objects and Uint8Array?

ethanfrey: Also name changed to `isJsRpcCompatibleDictionary`, right?

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+  SimpleMessagingConnection,
+} from "@iov/jsonrpc";
+import { firstEvent } from "@iov/stream";
+
+// An RPC interface very similar to JSON-RPC but supporting
+// all the data types from the structured clone algorithm
+// https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#Supported_types
+
+/** A single structured clone algorithm compatible value */
+export declare type JsRpcCompatibleValue =
```
ethanfrey: This is nice and good. Types plus link above adds clarity, changelog was more confusing

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+export function isJsRpcCompatibleArray(value: unknown): value is JsRpcCompatibleArray {
+  if (!Array.isArray(value)) {
+    return false;
+  }
+  return value.every(isJsRpcCompatibleValue);
+}
+
+export function isJsRpcCompatibleDictionary(data: unknown): data is JsRpcCompatibleDictionary {
+  if (typeof data !== "object" || data === null) {
+    // data must be a non-null object
```
ethanfrey: Is it ever possible that `typeof data === "object" && data === null` returns true?
Seems like a needless second check unless you can show me a case where needed (maybe a link)

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  // Exclude special kind of objects like Array, Date or Uint8Array
+  // Object.prototype.toString() returns a specified value:
+  // http://www.ecma-international.org/ecma-262/7.0/index.html#sec-object.prototype.tostring
+  if (Object.prototype.toString.call(data) !== "[object Object]") {
+    return false;
+  }
+
+  for (const key of Object.getOwnPropertyNames(data)) {
```
ethanfrey: looks good. but what about

`return Object.values(data).every(isJsRpcCompatibleValue)`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+    isJsRpcCompatibleArray(value) ||
+    isJsRpcCompatibleDictionary(value)
+  );
+}
+
+export function isJsRpcCompatibleArray(value: unknown): value is JsRpcCompatibleArray {
+  if (!Array.isArray(value)) {
+    return false;
+  }
+  return value.every(isJsRpcCompatibleValue);
```
ethanfrey: Very nice and compact

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }
+
+  return true;
+}
+
+export interface JsRpcRequest {
+  readonly id: number;
+  readonly method: string;
+  readonly params: JsRpcCompatibleArray | JsRpcCompatibleDictionary;
```
ethanfrey: :smile: supporting both formats from the spec

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export class JsRpcClient {
+  private readonly connection: SimpleMessagingConnection<JsRpcRequest, JsRpcResponse>;
+
+  constructor(connection: SimpleMessagingConnection<JsRpcRequest, JsRpcResponse>) {
+    this.connection = connection;
+  }
+
+  public async run(request: JsRpcRequest): Promise<JsRpcSuccessResponse> {
+    const filteredStream = this.connection.responseStream.filter(r => r.id === request.id);
```
ethanfrey: Love these two lines, to easily handle dispatching responses coming in any order. Oh, the power of streams.

packages/iov-core/src/jsrpc.ts
```diff
@@ -0,0 +1,242 @@
...
+    const pendingResponses = firstEvent(filteredStream);
+    this.connection.sendRequest(request);
+
+    const response = await pendingResponses;
+    if (isJsRpcErrorResponse(response)) {
+      const error = response.error;
+      throw new Error(`JS RPC error: code=${error.code}; message='${error.message}'`);
```
ethanfrey: This `JsRpcError` to `new Error` conversion? Is it only here?
If it is done multiple places, it would be good to capture it in a function for consistency

packages/iov-core/src/jsrpcsigningserver.spec.ts
```diff
@@ -155,24 +156,23 @@ describe("JsonRpcSigningServer", () => {
...
       method: "getIdentities",
       params: {
         reason: "Who are you?",
         chainIds: [ethereumChainId, bnsConnection.chainId()],
       },
     });
-    expect(response.jsonrpc).toEqual("2.0");
     expect(response.id).toEqual(123);
-    if (isJsonRpcErrorResponse(response)) {
+    if (isJsRpcErrorResponse(response)) {
```
ethanfrey: Ahh... I was confused why this was not using the client interface, but this is calling the server directly.

Maybe extract these lines to a function, same logic as in JsRpcClient.run.

Or even just use that client for these tests, as it seems a very light convenience wrapper around the server.
`await client.run(...)` instead of `await server.handleChecked(..)`. But this may break the test isolation, so feel free to leave as is... code to convert responses to `new Error` should be in one place though

packages/iov-core/src/jsrpcsigningserver.spec.ts
```diff
@@ -190,21 +190,21 @@ describe("JsonRpcSigningServer", () => {
...
       params: {
         reason: "Who are you?",
         chainIds: [bnsConnection.chainId()],
       },
     });
-    if (isJsonRpcErrorResponse(identitiesResponse)) {
+    if (isJsRpcErrorResponse(identitiesResponse)) {
       throw new Error(`Response must not be an error, but got '${identitiesResponse.error.message}'`);
     }
     const signer: PublicIdentity = identitiesResponse.result[0];
```
ethanfrey: Expect the signer to be defined and valid?

packages/iov-core/src/jsrpcsigningserver.ts
```diff
@@ -112,41 +108,38 @@ export class JsonRpcSigningServer {
...
 
     try {
-      let response: JsonRpcSuccessResponse;
+      let response: JsRpcSuccessResponse;
       switch (call.name) {
         case "getIdentities":
           response = {
-            jsonrpc: "2.0",
             id: request.id,
             result: await this.core.getIdentities(call.reason, call.chainIds),
```
ethanfrey: I need to dig deeper in the code, but since JsRpcRequest.params can be Array or Dictionary, does parseRpcCall handle the array format properly, as you refer to x.reason below

packages/iov-core/src/signingservice.spec.ts
```diff
@@ -238,12 +237,13 @@ describe("signingservice.worker", () => {
...
 
     const signAndPostResponse = await client.run({
-      jsonrpc: "2.0",
       id: 2,
       method: "signAndPost",
       params: {
         reason: "Please sign",
-        transaction: (send as unknown) as JsonCompatibleDictionary,
+        // Cast needed since type of indices of transaction is not string at compile time.
+        // see https://stackoverflow.com/a/37006179/2013738
```
ethanfrey: Good link and explanation

CHANGELOG.md
webmaster128: Before `isJsonCompatibleDictionary` accepted Uint8Array, Date, RegEx, etc as value. Those objects are not JSON compatible.

Now we have two different things
- `isJsonCompatibleDictionary` (fixed to accept JSON only)
- `isJsRpcCompatibleDictionary` a new function that allows more data types and can consume a full transaction object.

packages/iov-core/src/jsrpc.ts
webmaster128: The thing is: in JavaScript, `null` is an object:

```
> typeof null
'object'
```

This is what I wanted to express with that comment.

packages/iov-core/src/jsrpc.ts
webmaster128: Great find. This is exactly what the code tries to say. However, `Object.values` is only available in the ES2017 standard library and does not compile with our setup:

```
src/jsrpc.ts:66:17 - error TS2339: Property 'values' does not exist on type 'ObjectConstructor'.

66   return Object.values(data).every(isJsRpcCompatibleValue);
                   ~~~~~~


Found 2 errors.
```

webmaster128: But this works

```ts
  // replace with Object.values when available (ES2017+)
  const values = Object.getOwnPropertyNames(data).map(key => (data as any)[key]);
  return values.every(isJsonCompatibleValue);
```

packages/iov-core/src/jsrpcsigningserver.spec.ts
webmaster128: For the server, `JsRpcResponse = JsRpcSuccessResponse | JsRpcErrorResponse` are valid, non-error responses as described in https://www.jsonrpc.org/specification#response_object

It is the client's job to interprete both possible responses. In this test I expect a `JsRpcSuccessResponse` but in another test I might expect a `JsRpcErrorResponse`.

packages/iov-core/src/jsrpc.ts
webmaster128: Yes, only here. The client can choose how to interprete an error response. This client decided to convert error responses into promise rejections. A different client could handle error responses differently (like pass them as equally valid responses to the caller).

packages/iov-core/src/jsrpc.ts
webmaster128: Correct. We only use `JsRpcCompatibleDictionary` in our RPC interface but due to separation of concerns, the type allows both.

packages/iov-core/src/jsrpcsigningserver.ts
webmaster128: `parseRpcCall` ensures the dictionary parameters are used and checks the type of those fields

packages/iov-core/src/jsrpc.ts
ethanfrey: Wow. I didn't realize that... Maybe update the comment to read:

`// the thing is: in JavaScript, null is an object`

PR #846, opened 2019-03-25
Query ERC20 tokens using raw eth_call

### LINKED ISSUE #812: Query ERC20 balance
Builds on #809 and #810 

**Acceptance**

- [ ] Upload sample ERC20 contract with initial token distribution to ganache (test script)
- [ ] Query balances of accounts (present and not present)
- [ ] Consider generic "smart contract method call" api as well as erc20 specific api (that can wrap it)

webmaster128:
Based on #845 but with no web3x runtime dependency. Let's see if CI likes it

Closes #812


webmaster128:
About this PR: 

- Sorry for the huge diff! It includes the API file of our contracts and a new yarn.lock
- I was able to use raw Ethereum calls to query ERC20 balances, no additional dependency in IOV-Core
- There is AshToken (all fields set) and TrashToken (decimals, name, symbol not set) for local testing
- scripts/ethereum/deployment contains a sub project that uses web3x to deploy out contracts to the local ganache blockchain. This can at some point be moved out of this repo to be reusable.
- Request parameter and response encoding/decoding is in `Abi.ts`. What we need is pretty simple.
- Focus should go to the changes in `interface Erc20Options`, `class Erc20` and `class EthereumConnection`

packages/iov-ethereum/src/abi.ts
```diff
@@ -0,0 +1,89 @@
...
+import { isValidAddress } from "./address";
+
+export interface HeadTail {
+  /** An array of start positions within the original data */
+  readonly head: ReadonlyArray<number>;
+  /** Arguments split by positions as defined by head */
+  readonly tail: ReadonlyArray<Uint8Array>;
+}
+
+export class Abi {
```
ethanfrey: You really want to start re-implementing all the abi code?

Why not just copy in relevant files from ether.js?

packages/iov-ethereum/src/erc20.ts
```diff
@@ -0,0 +1,103 @@
...
+  readonly name?: string;
+  /** Override on-chain decimals. Use this of contract does not define value on-chain */
+  readonly decimals?: number;
+}
+
+export class Erc20 {
+  private readonly client: EthereumRpcClient;
+  private readonly options: Erc20Options;
+
+  constructor(client: EthereumRpcClient, options: Erc20Options) {
```
ethanfrey: Good design

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -86,11 +90,29 @@ export class EthereumConnection implements BcpConnection {
...
   private readonly socket: StreamingSocket | undefined;
   private readonly scraperApiUrl: string | undefined;
+  private readonly erc20Tokens = new Map<TokenTicker, Erc20>();
 
   constructor(baseUrl: string, chainId: ChainId, options?: EthereumConnectionOptions) {
     this.rpcClient = new HttpJsonRpcClient(baseUrl);
     this.myChainId = chainId;
 
+    const ethereumClient = {
+      ethCall: async (contractAddress: Address, data: Uint8Array): Promise<Uint8Array> => {
```
ethanfrey: :+1:

packages/iov-ethereum/src/abi.ts
webmaster128: The general ABI encoder/decoder is a complex and spread across mutiple files and packages. It contains a functionality that we don't need: encoding/decoding based on the ABI file. With ERC20 and Atomic swap we have less than 10 calls to covers and can easily compose the call and decompose the result in those places. I'm very happy that non of the code in IOV-Core requires the abi JSON of a contrect (it just exists in the deployment part). Thus I don't think we need a fully featured ABI coder.

ethanfrey: This is okay for erc20 (all tokens) and atomic swap support (only our contract).

There are requests for general smart contract calls on iov-core (various sales/marketing contracts that consider using iov-core to connect to their private blockchains). They would need some sort of `etherum/raw` transaction type that accepted a `json abi` and `any[]` data.

We don't need it for these stories, but will in the future, and I assumed easier to add that as a basis. However, if you can implement the needed calls easily hard-coded, just go for it, and general smart-contract support will be done in some months when MvP is finished and the requirement officially comes from business.

PR #852, opened 2019-03-26
ERC 20 support for getAllTickers(); Test ERC20 on rinkeby

### LINKED ISSUE #815: Integrate query erc20 balances with get ethereum account
When we query an account in ethereum, we can also query for the balances of the erc20 tokens. To start with we can use the same map as #814 to do the lookup, though at some point we need to limit which ones we look at (only query 5 not 500 each time).

**Acceptance**
- [ ] `getToken` and `getTokens` should list available erc20 tokens
- [ ] `getAccount` should show balance of eth and all available erc20 tokens in one list

webmaster128:
Closes #815

### DIFF
--- a/packages/iov-ethereum/src/ethereumconnection.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.ts
@@ -221,8 +221,9 @@ export class EthereumConnection implements BcpConnection {
     const erc20s = await Promise.all(
       [...this.erc20Tokens.entries()].map(
         async ([ticker, contract]): Promise<BcpTicker> => {
-          if (ticker !== (await contract.symbol())) {
-            throw new Error("Inconsistent ticker configured and in contract");
+          const symbol = await contract.symbol();
+          if (ticker !== symbol) {
+            throw new Error(`Configured ticker '${ticker}' does not match contract symbol '${symbol}'`);
           }
 
           return {
@@ -263,8 +264,9 @@ export class EthereumConnection implements BcpConnection {
 
     const erc20Balances: ReadonlyArray<BcpCoin> = await Promise.all(
       [...this.erc20Tokens.entries()].map(async ([ticker, contract]) => {
-        if (ticker !== (await contract.symbol())) {
-          throw new Error("Inconsistent ticker configured and in contract");
+        const symbol = await contract.symbol();
+        if (ticker !== symbol) {
+          throw new Error(`Configured ticker '${ticker}' does not match contract symbol '${symbol}'`);
         }
 
         return {
--- a/packages/iov-ethereum/src/ethereumconnection.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.ts
@@ -218,12 +218,29 @@ export class EthereumConnection implements BcpConnection {
   }
 
   public async getAllTickers(): Promise<ReadonlyArray<BcpTicker>> {
+    const erc20s = await Promise.all(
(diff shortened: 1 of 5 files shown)


packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -218,12 +218,29 @@ export class EthereumConnection implements BcpConnection {
   }
 
   public async getAllTickers(): Promise<ReadonlyArray<BcpTicker>> {
+    const erc20s = await Promise.all(
+      [...this.erc20Tokens.entries()].map(
+        async ([ticker, contract]): Promise<BcpTicker> => {
+          if (ticker !== (await contract.symbol())) {
+            throw new Error("Inconsistent ticker configured and in contract");
```
willclarktech: Weird grammar. Suggest `Ticker ${ticker} does not match contract symbol ${symbol}`.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -218,12 +218,29 @@ export class EthereumConnection implements BcpConnection {
...
+          }
+
+          return {
+            tokenTicker: (await contract.symbol()) as TokenTicker,
```
willclarktech: Is this an expensive query? No need to do it twice if so.

packages/iov-ethereum/src/testconfig.spec.ts
```diff
@@ -20,14 +22,27 @@ export interface EthereumNetworkConfig {
...
-  readonly accountState: {
-    readonly pubkey: PublicKeyBundle;
-    readonly address: Address;
-    /** expected balance for the `address` */
-    readonly expectedBalance: Amount;
-    /** expected nonce for the `address` */
-    readonly expectedNonce: Nonce;
+  readonly accountStates: {
+    /** An account with ETH and ERC20 balance */
+    readonly full: {
```
willclarktech: This is a strange name. Suggest `ethAndErc20`?

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: Thanks for suggestion. I was not happy with the text as well but coder brain did not come up with proper language.

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: It is explicitly documented to be cheap:

```ts
    /**
     * Returns symbol value from options or from chain.
     *
     * On-chain values will be cached internally, i.e. it is cheap to use this getter
     * as long as the class instance is long living.
     */
    name(): Promise<string>;
    /**
     * Returns symbol value from options or from chain.
     *
     * On-chain values will be cached internally, i.e. it is cheap to use this getter
     * as long as the class instance is long living.
     */
    symbol(): Promise<string>;
    /**
     * Returns decimals value from options or from chain.
     *
     * On-chain values will be cached internally, i.e. it is cheap to use this getter
     * as long as the class instance is long living.
     */
    decimals(): Promise<number>;
```

packages/iov-ethereum/src/testconfig.spec.ts
webmaster128: Hmm the account should be used in all kind of tests (e.g. nonce tests) and I don't want to restrict the naming to ETH+ERC20. Thus I came up with `full`. Maybe `default` is better?

PR #832, opened 2019-03-12
[WIP] Upgrade to bnsd v0.12.0

alpe:
**WIP - Do not merge** 
This PR is for further discussions on the proper implementation. 

Breaking changes in Bnsd:
* Blockchain NFT was dropped
* Escrow does not contain an amount anymore. Instead the escrow id would be converted into a weave address (via Condition) to query the balance like any other account. We have this for multisig or distribution implemented in Bnsd already. I had marked this with a TODO in the code here. The backend impl would look like this: https://github.com/iov-one/weave/blob/master/x/escrow/init_test.go#L64

`yarn test` fails with:
```
Error: Cannot find module '@iov/bcp'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:613:15)
    at Function.Module._load (internal/modules/cjs/loader.js:539:25)
    at Module.require (internal/modules/cjs/loader.js:667:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/alex/workspace/iov/iov-core/packages/iov-bns/src/bnscodec.spec.ts:1:1)
    at Module._compile (internal/modules/cjs/loader.js:738:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:749:10)
    at Module.load (internal/modules/cjs/loader.js:630:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:570:12)
    at Function.Module._load (internal/modules/cjs/loader.js:562:3)
error Command failed with exit code 4.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #856, opened 2019-03-26
Upgrade to bnsd 0.13

### LINKED ISSUE #843: Add support for BNS 0.12
IOV-Core 0.13.x is the last branch compatible with BNS 0.11 (Hugnet) and can receive patch releases as needed. Current master can be changed to BNS 0.12 support.

- [ ] Update compatibility table in iov-bns REDME
- [ ] Update iov-bns

### LINKED ISSUE #847: Add support for bns 0.13.0 / escrows use timestamps
Depends on #843 and #855

---

The only external changes between those releases are changing how multisigs are represented (requiring a weight for each member), and using timestamps from block headers to control escrow.

I would change the bcp abstraction layer along with this pull request to expose a timestamp to the client, while possibly writing a block header on the chain. This should improve the api for multi-chain usage, while moving the complexity from the client into iov-core.

ethanfrey:
Closes #843 
Closes #847 

Thanks to @alpe and https://github.com/iov-one/iov-core/pull/832/commits/e16f6ca3576e8ce441789516d4528563bd2bcc01 
I used that as a reference for much of the cleanup, although cherry-picking was causing merge conflicts.

- [x] Update protobuf definitions
- [x] Remove references to deleted BlockchainNft
- [x] Use timestamp not height in bns escrow types
- [x] use new genesis file with fees
- [x] add support for querying fees
- [x] Update escrow queries to get account info separately
- [x] Fix all tests
- [ ] Add support for product fees (or separate PR?)


ethanfrey:
This is not 100% done yet, but pretty close.
I need to add one functionality to get the proper balance on escrow queries and a few tests.

But I'd be happy for a review when you have a chance.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -116,27 +115,31 @@ describe("BnsConnection", () => {
...
     identity: PublicIdentity,
   ): Promise<void> {
-    const sendTx: SendTransaction = {
+    const sendTx = (await connection.withDefaultFee({
       kind: "bcp/send",
       creator: identity,
       recipient: await randomBnsAddress(),
       amount: defaultAmount,
-    };
+    })) as SendTransaction;
```
webmaster128: Hmm, not so happy about all the necessary casting for the default use case.

Can't we use

```ts
const signed = await profile.signTransaction(await connection.withDefaultFee(sendTx), bnsCodec, nonce);
```

instead?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1087,13 +995,16 @@ describe("BnsConnection", () => {
...
 
-      const sendTx: SendTransaction = {
+      const sendTx = (await connection.withDefaultFee({
         kind: "bcp/send",
         creator: brokeIdentity,
         recipient: await randomBnsAddress(),
         memo: "Sending from empty",
         amount: defaultAmount,
-      };
+      })) as SendTransaction;
```
webmaster128: okay here we need `sendTx` with fees for the code below.

What about

```ts
const sendTx: SendTransaction = await connection.withDefaultFee({
  kind: "bcp/send",
  creator: brokeIdentity,
  recipient: await randomBnsAddress(),
  memo: "Sending from empty",
  amount: defaultAmount,
});
```

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: See also other idea below

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -623,18 +606,25 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+      throw new Error("Received transaction of unsupported kind.");
+    }
+    // get gconf fee
+    // TODO: get product fee for this type
+    // take maximum
+    const res = await this.query("/", Encoding.toAscii("gconf:cash:minimal_fee"));
+    if (res.results.length === 0) {
+      throw new Error("No fee defined");
+    } else if (res.results.length > 1) {
+      throw new Error("Received unexpected number of fees");
```
webmaster128: This can be more compact: `res.results.length !=== 1`, which works for 0 as well

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -623,18 +606,25 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
     }
+    const data = Encoding.fromAscii(res.results[0].value);
+    const amount = decodeJsonAmount(data);
+    return { tokens: amount };
+  }
 
-    throw new Error("Received transaction of unsupported kind.");
+  public async withDefaultFee<T extends UnsignedTransaction>(transaction: T): Promise<T> {
```
webmaster128: 👍

packages/iov-bns/src/context.ts
```diff
@@ -72,8 +72,10 @@ export class Context {
         sender: encodeBnsAddress(addressPrefix(this.chainData.chainId), ensure(swap.sender)),
         recipient: encodeBnsAddress(addressPrefix(this.chainData.chainId), ensure(swap.recipient)),
         hash: hash,
-        amounts: ensure(swap.amount).map(coin => decodeAmount(coin)),
-        timeout: { height: asNumber(swap.timeout) },
+        // amounts: ensure(swap.amount).map(coin => decodeAmount(coin)),
+        // TODO: read this is a second query
+        amounts: [],
```
webmaster128: Why was amount removed from Escrow? This seems to be very bad for the client

packages/iov-bns/src/decode.spec.ts
```diff
@@ -216,6 +179,75 @@ describe("Decode", () => {
     });
   });
 
+  describe("decodeJsonAmount", () => {
+    it("can decode strings ", () => {
```
webmaster128: extra space

packages/iov-bns/src/decode.ts
```diff
@@ -135,6 +94,37 @@ export function decodeAmount(coin: codecImpl.x.ICoin): Amount {
   };
 }
 
+export function decodeJsonAmount(json: string): Amount {
+  const data = JSON.parse(json);
+  if (typeof data === "string") {
+    // parse "1.23 IOV", ".001CASH", "12   FOO"
```
webmaster128: The second example is not supported accorting to tests

packages/iov-bns/src/decode.spec.ts
```diff
@@ -216,6 +179,75 @@ describe("Decode", () => {
...
+      expect(() => decodeJsonAmount(`"42.13"`)).toThrowError();
+      // just ticker
+      expect(() => decodeJsonAmount(`"FOO"`)).toThrowError();
+      // wrong order
+      expect(() => decodeJsonAmount(`"ASH 22"`)).toThrowError();
+      // ticker bad size
+      expect(() => decodeJsonAmount(`"0.01 A"`)).toThrowError();
+      // ticker bad size
+      expect(() => decodeJsonAmount(`"0.01 SUPERLONG"`)).toThrowError();
+    });
```
webmaster128: Can we test more than 9 fractional positions? e.g. `0.0123456789` This cannot be represented as a weave amount.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1615,6 +1442,7 @@ describe("BnsConnection", () => {
     expect(faucetEndBalance.quantity).toEqual(
       Long.fromString(faucetStartBalance.quantity)
         .subtract(68_000000000)
+        .subtract(10_000000) // the fee (0.01 CASH)
```
webmaster128: What about `0_010000000` to use _ for the full/fractional separation like above?

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -623,18 +606,25 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
-          tokenTicker: firstToken.tokenTicker,
-        },
-      };
+    if (!isBnsTx(transaction)) {
+      throw new Error("Received transaction of unsupported kind.");
+    }
+    // get gconf fee
+    // TODO: get product fee for this type
+    // take maximum
+    const res = await this.query("/", Encoding.toAscii("gconf:cash:minimal_fee"));
```
webmaster128: Why do we need to support the human readable token format in a machine-machine interface? I thought this was created for genesis editing only.

packages/iov-bns/src/bnsconnection.spec.ts
ethanfrey: That doesn't work (my first try), as the input expands the type from `kind: "bcp/send"` to `kind: string` when passed through the generic withDefaultFee. Thus we need an explicit cast. Explicit casts to other types do not work however, as they have the wrong fields (eg. amount and memo).

Try if you want, but this was the best I could do with the typings

packages/iov-bns/src/bnsconnection.ts
ethanfrey: :+1:

packages/iov-bcp/src/transactions.ts
```diff
@@ -239,6 +240,10 @@ export function isTimestampTimeout(timeout: SwapTimeout): timeout is TimestampTi
   return typeof (timeout as TimestampTimeout).timestamp === "number";
 }
 
+export function timeoutInSeconds(seconds: number): TimestampTimeout {
```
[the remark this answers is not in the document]
ethanfrey: will change. i admin i am not good at naming

packages/iov-bns/src/context.ts
ethanfrey: Because it was not accurate in the general case. Let's discuss this more off-line.

Multisigs, distribution, etc will need two queries, as we do not have triggers in weave to call custom functionality when a contract receives tokens (as in ethereum). We can debate if we need such.

Out of scope of this PR, but let's have some design discussions on this topic

packages/iov-bns/src/decode.spec.ts
ethanfrey: Good point

packages/iov-bns/src/bnsconnection.spec.ts
ethanfrey: Sure

packages/iov-bns/src/bnsconnection.ts
ethanfrey: Yeah, it was.
But somehow gconf doesn't actually parse the genesis file before writing to db (as it should), but rather copies it blindly until we read it.
Didn't realize this gotcha, til I tried parsing with protobuf Coin and it failed.

We should discuss gconf and improvements in a design session.

packages/iov-bns/src/bnsconnection.spec.ts
ethanfrey: I tried again to make types nicer and got this:

```ts
const sendTx = await connection.withDefaultFee<SendTransaction>({
      kind: "bcp/send",
      creator: identity,
      recipient: await randomBnsAddress(),
      amount: defaultAmount,
    });
```

We just enforce the generic type T of the function. Before we explicitly set `const signed: SendTransaction`. This does not force-cast the type, and will complain if the wrong string is used in kind. Does that work better for you? If so, I will update them all.

ethanfrey: I used this several places and I like it

ethanfrey:
Waiting for CI to tell me...
I ran the entire test suite with all blockchains enabled localy (`yarn test`) and it passed. But didn't try in browsers.

Otherwise, I think this is done, except for missing feedback on withDefaultFee usage.

And there are two design discussions that need to be had.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Yeah, cool, it is definitly better!

webmaster128:
Using the generics specialization is definitely a big win over the casting. One reason is that additional fields are disallowed for object literals of a fixed type but can be casted out of the type system.

Broken but no error:

```ts
      const sendTx = (await connection.withDefaultFee({
        kind: "bcp/send",
        creator: brokeIdentity,
        recipient: await randomBnsAddress(),
        memo: "Sending from empty",
        amount: defaultAmount,
        foo: "bar",
      })) as SendTransaction;
```

Error _Object literal may only specify known properties, and 'foo' does not exist in type 'SendTransaction'.ts(2345)_:

```ts
      const sendTx = await connection.withDefaultFee<SendTransaction>({
        kind: "bcp/send",
        creator: brokeIdentity,
        recipient: await randomBnsAddress(),
        memo: "Sending from empty",
        amount: defaultAmount,
        foo: "bar",
      });
```

The other is that the cast allows any string as the kind field.

Broken but no error:

```ts
      const sendTx = (await connection.withDefaultFee({
        kind: "bcp/send123",
        creator: brokeIdentity,
        foo: "bar",
        recipient: await randomBnsAddress(),
        memo: "Sending from empty",
        amount: defaultAmount,
      })) as SendTransaction;
```

Error _Type '"bcp/send123"' is not assignable to type '"bcp/send"'.ts(2322)_:

```ts
      const sendTx = await connection.withDefaultFee<SendTransaction>({
        kind: "bcp/send123",
        creator: brokeIdentity,
        foo: "bar",
        recipient: await randomBnsAddress(),
        memo: "Sending from empty",
        amount: defaultAmount,
      });
```

packages/iov-core/src/signingservercore.spec.ts
```diff
@@ -232,6 +232,8 @@ describe("SigningServerCore", () => {
   describe("signAndPost", () => {
     it("can sign and post", async () => {
       pendingWithoutBnsd();
+      // TODO: fix before merge
+      pending("We need to get some money to these accounts somehow.. not sure how");
```
webmaster128: You can merge https://github.com/iov-one/iov-core/pull/861 into this to get this done

scripts/bnsd/genesis_app_state.json
```diff
@@ -2,46 +2,29 @@
...
     },
     {
       "ticker": "BASH",
       "name": "Another token of this chain",
-      "sig_figs": 6
     },
     {
       "ticker": "ASH",
       "name": "Let the Phoenix arise",
-      "sig_figs": 6
```
webmaster128: Could you run a `yarn prettier --write scripts/bnsd/genesis_app_state.json scripts/bcpd/genesis_app_state.json` to fix the broken JSON (trailing comma in the line above)?

packages/iov-core/src/signingservercore.spec.ts
ethanfrey: Done

ethanfrey:
@webmaster128 This is done when the CI sings.

Open question is naming of this part: https://github.com/iov-one/iov-core/blob/11cd7ed419ebe0cb64cefe1930de576584f4aa86/packages/iov-bns/src/util.ts#L63-L83 which mirrors weave logic of how to calculate the address of an escrow in order to move it.

webmaster128:
@ethanfrey regarding naming: since this is all package internal, we can cleanup easily at any time. Looks good for now

PR #859, opened 2019-03-27
Add atomic swap protocol

### LINKED ISSUE #850: Fully document how IOV atomic swap contract works, what it needs for compatibility with other blockchains.
Is this documented during the ethereum-smart-contracts rewrite?

I think we all understand it know, just please add some docs into iov-core (iov-bcp package?) in a PR.

I will adjust it as part of #847  as to timestamps/block height

We need to discuss timestamp/block height clarification.

```ts
type Timeout = BlockHeightTimeout | TimestampTimeout;

interface BlockHeightTimeout {
  readonly timeoutAtHeight: number;
}

interface TimestampTimeout {
  readonly timeoutAtTime: Date;
}

interface CreateAtomicSwapTransaction extends Timeout {
...
}
```

I would propose that we *try* to support both for each blockchain - one is natively supported on chain, and if possible, we translate the other under the hood. Use the proper one for maximum control.

webmaster128:
Depends on #856
Closes #850

---

This ads a specification of the IOV atomic swap protocol and adapts the client-side implementation accordingly.

- [x] Set authors
- [x] Do TODOs
- [x] Explicitly require that anyone can claim

### DIFF
--- a/docs/atomic-swap-protocol-v1.md
+++ b/docs/atomic-swap-protocol-v1.md
@@ -81,7 +81,23 @@ data with `hash = sha256(preimage)`.
 
 ### Timeouts
 
-TODO
+A timeout is the first point in time, when the HTLC is expired. It is measured
+in one of the two types absolute times or block height.
+
+- **Absolute time** is a second resolution
+  [UNIX time](https://en.wikipedia.org/wiki/Unix_time). It is used for
+  blockchains that have a strong guarantee for correct and fair timestamps with
+  reasonable precision.
+- **Block height** is an integer counting the number of blocks created. It is
+  used when absolute time is not appropriate.
+
+Note: Since the type of timeout is fixed in the HTLC, which belongs to the
+blockchain, a cross-chain atomic swap may need to use different types of
+timeouts for the initial offer and the counter offer.
+
+The counter offer is claimed first. I.e. the time between the counter offer's
+timeout and the initial offer's timeout is the minimal time the actor in short
+position has to claim the initial offer.
 
 ### Swap offer actions
 
@@ -160,7 +176,24 @@ locked tokens until the timeout is reached.
 
 ### Half executed atomic swaps
 
-TODO: timeout issue etc.
+1. **Timeout difference:** The actor in short position needs to ensure to have a
+   sufficient amount of time to claim the initial offer, even if the preimage
+   was revealed by claiming the counter offer at the very last possible point in
+   time. This should cover all risks that can delay the execution of the claim
+   transaction like different time zones, downtimes of the default blockchain
+   nodes, temporary loss of power, internet connection and access to key
+   material.
+2. **Block height based timeout calculation:** The above timeout difference is
+   even harder to calculate when timeouts are measured in block height.
+   Especially with non-constant block times one needs to calculate the expected
+   block height at the desired timeout time and take into account a safety
+   margin.
+3. **Chain stops:** When one chain in a cross-chain atomic swap stops, two
+   things can happen that lead to half executed atomic swaps. (a) A timeout
+   measured in absolute time is reached without the chance to get a claim
+   transaction processed. (b) An initial offer's block height based timeout is
+   pushed unexpectedly far into the future, such that the timeout difference for
+   the actor in short position to claim is 0 or negative.
 
 ## License
 
--- a/docs/atomic-swap-protocol-v1.md
+++ b/docs/atomic-swap-protocol-v1.md
@@ -18,21 +18,21 @@ specification to build compatible components.
 Two actors Alice (A) and Bob (B) agreed on a token exchange off-chain. Both have
 one address on each blockchain X and Y.
 
-- Alice creates a secret key locally (called preimage) and stores it.
+- Alice creates a secret key locally (called a "preimage") and stores it.
 - Alice creates a swap offer on X, the blockchain where her tokens live. This is
-  a HLTC that contains the recipient address of Bob on X (b<sub>X</sub>), the
+  a HTLC that contains the recipient address of Bob on X (b<sub>X</sub>), the
   amount of tokens offered, the hash of the preimage and a timeout. This
   contract locks her tokens and sends them to b<sub>X</sub> if the preimage is
-  releaved on-chain.
+  revealed on-chain.
 - Bob creates a swap offer on Y, the blockchain where his tokens live. This is a
-  HLTC that contains the recipient address of Alice on Y (a<sub>Y</sub>), the
+  HTLC that contains the recipient address of Alice on Y (a<sub>Y</sub>), the
   amount of tokens offered, the hash copied from Alice's offer and a timeout.
   This contract locks his tokens and sends them to a<sub>Y</sub> if the preimage
-  is releaved on-chain.
+  is revealed on-chain.
 - (Variant 1): Alice now uses her preimage to claim Bob's swap offer on Y,
-  causing the token transfer from b<sub>Y</sub> to a<sub>Y</sub>. Bob looks up
-  this now public preimage on Y to claim Alice's swap offer on X, causing the
-  token transfer from a<sub>X</sub> to b<sub>X</sub>.
+  triggering the token transfer from b<sub>Y</sub> to a<sub>Y</sub>. Bob looks
+  up this now public preimage on Y to claim Alice's swap offer on X, triggering
+  the token transfer from a<sub>X</sub> to b<sub>X</sub>.
 - (Variant 2): Alice decides not to reveal her preimage before the timeout of
   Bob's offer is reached. As soon as the timeout is reached, Bob aborts his
   offer and gets back his tokens. Alice also needs to wait for the timeout of
@@ -42,16 +42,16 @@ one address on each blockchain X and Y.
 
 As soon as both offers are created, the actor that created the preimage has the
 one-sided option to execute the atomic swap or to let it expire. Thus this actor
-is called to be _in long position_. The other actor is called to be _in short
-position_.
+is described as being _in long position_. The other actor is described as being
+_in short position_.
 
 Being in long position has two advantages:
 
 1. having the free option to execute the swap or not and
-2. claiming the other actor's swap offer first, which reduces the risk of
-   loosing tokens in the case of a half executed swap.
+2. claiming the other actor's swap offer first, which reduces the risk of losing
+   tokens in the case of a half executed swap.
 
-Being in short position has that advantage to be able to review the other
+Being in short position has the advantage of being able to review the other
 actor's swap offer before creating an offer oneself.
 
 ## The swap offer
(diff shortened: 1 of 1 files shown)


docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+
+TODO
+
+### Swap offer actions
+
+Actions are mutations of the on-chain state triggered by transactions.
+
+- **Create:** By creating a swap offer, the funds are locked in the contract. A
+  timeout and a hash must be specified.
+- **Claim:** By revealing a preimage for the hash before the timeout is reached,
```
willclarktech: In our Ethereum implementations we currently don't require the timeout not to have reached for funds to be claimed. Should we? (It's why we called it `abortable` rather than `expired`.)

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+  the contract transfers the tokens to the recipient. Requires a preimage of
+  exactly 32 bytes. Common synonyms: _release_, _execute_.
+- **Abort:** Withdraws the tokens from the contract to the creator. Requires
+  timeout to be reached.
+
+### Swap offer open state
```
willclarktech: It's confusing that "open" is used for this state category and one of the possible positions.

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+
+Each swap offer individually is either _open_, _claimed_ or _aborted_. The open
+state can only be changed by performing an action via a transaction.
```
willclarktech: suggest: Mention that claimed/aborted swap offers cannot change state again.

docs/atomic-swap-protocol-v1.md
willclarktech: It's also confusing that here "swap offer" covers the counter offer, whereas above it seems to be in opposition to it.

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+
+- **Open:** This is the initial open state.
+- **Claimed:** An _open_ swap becomes _claimed_ by performing the clain action.
+- **Aborted:** An _open_ swap becomes _aborted_ by performing the abort action.
+
+Common synonym: _settled = (claimed or aborted)_
```
willclarktech: suggest: "closed = claimed"

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Swap offer expiry state
+
+Each swap offer individually is either _non-expired_ or _expired_. The state
+automatically changes on-chain as the underlying time (wall time or block
+height) increases. A state can switch from _non-expired_ to _expired_ but not
+vice versa.
+
+- **Non-expired:** In the range `[creation, timeout)`, the swap offer is
```
willclarktech: How well known is this notation for inclusive/exclusive ranges?

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+#### Note on open and expiry state
+
+Both states are independent of each other. A swap offer can be open and expired
+at the same time.
+
+## Risks
+
+The following risks for cross-chain atomic swaps have been identified. Some of
+them are fully avoided by this protocol but still documented for tranparancy.
+Others need to be mitigated by smart choice of parameters.
```
willclarktech: suggest: Make it clear who's responsibility this is: "Others need to be mitigated by the user's choice of parameters."

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Pre-image attacks
+
+The short position copies the hash without knowing the preimage. Now the long
```
willclarktech: suggest: "The actor in short/long position"

willclarktech: suggest: "This means" rather than "Now"

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+position could use a very long preimage that is either expensive or impossible
+to process on one chain but works easily on the other chain. This can make it
+unprofitable or impossible for short to claim their tokens, leading to free
+tokens for long.
+
+This risk is avoided in this protocol by having a fixed preimage length.
+
+### No counter offer created
+
+The short position can pretend to create a counter offer but not do it. In this
```
willclarktech: suggest: "pretend to have the intention to create a counter offer"

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+case long loses access to the locked tokens until the timeout is reached.
+
```
willclarktech: We should make it clear we don't avoid this risk and how to mitigate it (trust where possible, don't lock up more funds than you can afford losing access to for the relevant time period).

docs/atomic-swap-protocol-v1.md
webmaster128: Yes I was also thinking about this issue. Do you have a suggestion for a change? Either for the state name or the state value.

willclarktech: Best I could come up with is just calling this the "swap offer state" and calling the next section "swap offer expiration" (not mentioning state). Otherwise, "swap offer process state"?

webmaster128: I prefer "swap offer process state" since both are states of the swap offer. Process state will do the job

willclarktech: Do you think we should rename the enum in the contract then?

webmaster128: Yes. Not super urgent but I think we should do it.

I will also rename this client-side enum: https://github.com/iov-one/iov-core/blob/atomic-swap-doc/packages/iov-bcp/src/atomicswaptypes.ts#L7

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,156 @@
+# IOV atomic swap protocol version 1
+
+| Status | Authors | Created    | Last updated | License   |
+| ------ | ------- | ---------- | ------------ | --------- |
+| Draft  |         | 2019-03-27 | 2019-03-27   | CC-BY-4.0 |
```
ethanfrey: Add Simon, Ethan, and Isabella?

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+This is a cross-chain atomic swap protocol used by IOV. It is based on hash time
+locked contracts (HTLCs).
+
+IOV provides reference implementations for Ether and ERC20 tokens on Ethereum
+and any token on weave-based blockchains. Any third party is free to use this
+specification to build compatible components.
+
+## Overview
+
+Two actors Alice (A) and Bob (B) agreed on a token exchange off-chain. Both have
```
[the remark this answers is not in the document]
ethanfrey: Good point, we use their names below as well.

docs/atomic-swap-protocol-v1.md
ethanfrey: Yes, we should not allow claiming after the timeout

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,156 @@
...
+
+### No counter offer created
+
+The short position can pretend to have the intention to create a counter offer
+but not do it. In this case long loses access to the locked tokens until the
+timeout is reached.
+
+### Half executed atomic swaps
+
+TODO: timeout issue etc.
```
ethanfrey: Add here text about choice of timeout.

Main point is the short position must expire before the long position.
And with a good margin of time (eg. 24 hours), so it is impossible that the short position is claimed eg. 1 minute before expiry and the counter-party sleeps through their window to claim the other half.

This is on the end users to choose/ensure proper time envelopes, but this should be well explained.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1633,8 +1634,8 @@ describe("BnsConnection", () => {
     const initSwaps = await connection.getSwaps({ recipient: recipientAddr });
     expect(initSwaps.length).toEqual(0);
 
-    const swapOfferPreimage = Encoding.toAscii(`my top secret phrase... ${Math.random()}`);
-    const swapOfferHash = new Sha256(swapOfferPreimage).digest();
+    const swapOfferPreimage = await AtomicSwapHelpers.createPreimage();
```
ethanfrey: Much cleaner here with the helpers.

docs/atomic-swap-protocol-v1.md
webmaster128: okay: https://github.com/iov-one/ethereum-smart-contracts/issues/20

This makes the term "abortable" obsolete since it is equivalent to "expired" (which is what is used in the BNS implementation)

docs/atomic-swap-protocol-v1.md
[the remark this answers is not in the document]
webmaster128: Agreed, not strictly needed

docs/atomic-swap-protocol-v1.md
webmaster128: > A claimed or aborted swap offer process state cannot change again, i.e. the only
possible transitions are _open → claimed_ and _open → aborted_.

?

willclarktech: How about: "A swap offer which has been claimed or aborted cannot change to another process state, i.e. ..."?

webmaster128: Right, reads more natural, thanks

docs/atomic-swap-protocol-v1.md
webmaster128: Yeah might not be clear right away for everybody. II better call it "half-open interval" explicitly in text so everyone who does not know the notation gets a hint on what to search for.

Interesting find there: a "range" is not a synonym for an interval but the interval's size (https://en.wikipedia.org/wiki/Interval_(mathematics)#Terminology).

Glad we taked about that :)

docs/atomic-swap-protocol-v1.md
```diff
@@ -0,0 +1,155 @@
...
+   loosing tokens in the case of a half executed swap.
+
+Being in short position has that advantage to be able to review the other
+actor's swap offer before creating an offer oneself.
+
+## The swap offer
+
+A swap offer is a HTLC with the following properties:
+
+- The amount of tokens is fixed. Implementation may support a group of multiple
```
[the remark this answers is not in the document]
webmaster128: Better?

> The amount of tokens is fixed. Implementations may support including multiple different tokens at once.

webmaster128:
I appreciate your re-reviews and approvals. I'm trying to keep this PR as small as possible to be merged soon. However, I'd like to at least fill the missing TODOs (especially typeout docs) and explicitly require that anyone can claim (the emergency case where I need to ask a friend to claim because I temporarily don't have access to my keys).

Fixing the integration test was pulled out now: https://github.com/iov-one/iov-core/issues/919.

webmaster128:
Gentlemen, I added 3 new commits. Could you please have another look at those additions?

- f840ff0e5f477c5be36fd9e72b7e79e367b7ec68
- da1056637c1f6a1da0df6c79bb2fb3c32195dc96
- 9e17f9482bc198f62a3ddad2aa9c484e5bd8b9d6

I now consider this good enough for merging.

ethanfrey:
Feel free to merge from my side, once you both agree on it.

Issue #829, opened 2019-03-07
Root package.json scripts should also apply to root dir


willclarktech:
- [ ] lint
- [ ] format

webmaster128:
What is there to be linted/formatted in the repo root? Both steps are done for *ts files only

willclarktech:
I guess linting's not an issue, but Prettier can make sure code blocks in `.md` files are formatted nicely. (And less importantly `.json` files).

webmaster128:
Now that `lint-fix` can take care of our source code, we can use `yarn format` for formatting configs. So I propose the following path:

1. Create `lint-fix` scripts in all packages and root (see ponferrada) for manual execution
2. Remove `yarn format`, as this is covered by `yarn lint`/`yarn lint-fix`
3. Create new `format` script that formats configs `.json` / `.yml` and does not run on files that are covered by eslint. (This one I'm not exactly sure about. Is there a good reason to keep jsom/yml/md out of the default prettier+autofixer setup?)

willclarktech:
> Is there a good reason to keep jsom/yml/md out of the default prettier+autofixer setup?

Best I can think of is that the prettier formatters for these kinds of files seem to be a bit more bug-prone than for `.js`/`.ts`, but I don't think that's a good enough reason not to include them by default.

PR #860, opened 2019-03-27
Use prettier to format .md files

webmaster128:
First step for #829

The idea is to format code and text separatly
- since different width is desired,
- to avoid unnecessary build times during development.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "format": "lerna run format",
+    "format-text": "prettier --write --prose-wrap always --print-width 80 \"./*.md\" \"./docs/**/*.md\" && lerna run format-text",
     "lint": "lerna run lint",
     "test": "lerna run test",
     "build": "lerna run build",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,459 +6,518 @@
 
 ## 0.13.0
 
-* @iov/bcp: Add `AtomicSwapMerger`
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "format": "lerna run format",
```
willclarktech: I would expect the `format` script to cover the root as well.

package.json
```diff
@@ -18,6 +18,7 @@
...
+    "format-text": "prettier --write --prose-wrap always --print-width 80 \"./*.md\" \"./docs/**/*.md\" && lerna run format-text",
```
willclarktech: This recurses, no?

package.json
webmaster128: The default "format" implementation is `prettier --write --loglevel warn \"./src/**/*.ts\"`, which does not do anything in the root dir. Thus I did not add it here.

package.json
webmaster128: No:

```
lerna run <script>

Run an npm script in each package that contains that script
```

i.e. `lerna run format-text` runs `yarn format-text` for each folder in `packages/` only.

willclarktech: Ah, right. `lerna run` not `yarn run`.

webmaster128:
> There are a few situations where prettier requires more than one pass to reach a stable point, which is possibly what's happened here.

In this case the change from

```
------------------ Rough ideas not quite ready for backlog ------------------------
```

to 

```
------------------ Rough ideas not quite ready for backlog
------------------------
```

caused the second run to interprete it as a h2. Will fix manually

Issue #818, opened 2019-03-06
Create transactions for simple ETH-ETH atomic swap (one chain)


ethanfrey:
Builds on #817 

Add some custom ethereum transactions and execute an atomic swap.

Use two different copies of the same contract on one ganche chain, so they think they are separate.

As this would be second specific contract call (after erc20 example), this should include refactoring the generic "abi contract call" transaction to work for the more general case.
 
Note: we could also start up to local ganache chains and swap between them if not too much overhead to set up

webmaster128:
To get this done as simple as possible in our existing setup, I suggest the following steps:

1. Deploy Ether atomic swap contract (see a. https://github.com/iov-one/iov-core/tree/master/scripts/ethereum/deployment/contracts b. https://github.com/iov-one/iov-core/blob/master/scripts/ethereum/deployment/contracts.json c. https://github.com/iov-one/iov-core/blob/master/scripts/ethereum/deployment/package.json#L10 d. https://github.com/iov-one/iov-core/blob/master/scripts/ethereum/deployment/src/deploy_contracts.ts#L8 e. https://github.com/iov-one/iov-core/blob/master/scripts/ethereum/deployment/src/deploy_contracts.ts#L41 f. https://github.com/iov-one/iov-core/blob/master/scripts/ethereum/init.sh#L14-L16 g. https://github.com/iov-one/iov-core/tree/master/scripts/ethereum#startstop)
2. Test one-sided swap offer handling similar to https://github.com/iov-one/iov-core/blob/v0.13.0/packages/iov-bns/src/bnsconnection.spec.ts#L1623-L1911
3. Do an ETH-ETH atomic swap in @iov/ethereum using the same contract instance (address printed as part of 1.)

The cross-contract case can easier be tested by integrating Ethereum into `packages/iov-core/src/integrationtests/atomicswap.spec.ts`. So we can go for cross-chain straight.

ethanfrey:
(1) and (2) are good prep, and I think related to other issues.

Updating (3) would solve it, and if you want to go right to ETH-IOV swap in that integration test, you can close two issues in one step :)

Issue #813, opened 2019-03-06
Send ERC20 tokens


ethanfrey:
Builds on #812 

- [ ] Make a unique transaction type to send a given ERC20 token to a new account.
- [ ] Ethereum Codec can encode/decode this transaction properly
- [ ] It is executed on the chain and updates balances as expected.

webmaster128:
>  Make a unique transaction type to send a given ERC20 token to a new account.

Why do you think we cannot encode that token send in a bcp/send transaction?

ethanfrey:
We need a lookup table to translate, as there is info in the ERC20 transfer (such as the token name -> address + sig figs) that is not present in `bcp/send` and must come from somewhere.

I meant this to be a simple step, as you like smaller, clear issues, and then #814 would handle doing that mapping transparently, "under the hood".

PR #863, opened 2019-03-28
Extract and test calculateMethodId

webmaster128:
Needed for #813 and #818

### DIFF
--- a/packages/iov-ethereum/src/abi.spec.ts
+++ b/packages/iov-ethereum/src/abi.spec.ts
@@ -6,6 +6,15 @@ import { Abi } from "./abi";
 const { fromHex } = Encoding;
 
 describe("Abi", () => {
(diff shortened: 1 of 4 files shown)


packages/iov-ethereum/src/abi.spec.ts
```diff
@@ -6,6 +6,15 @@ import { Abi } from "./abi";
 const { fromHex } = Encoding;
 
 describe("Abi", () => {
+  describe("calculateMethodId", () => {
+    it("works", () => {
+      // Examples from https://solidity.readthedocs.io/en/develop/abi-spec.html#examples
```
willclarktech: Is it possible to fix this to a version? E.g. v0.5.7

webmaster128: Good idea, done

PR #865, opened 2019-03-28
Deploy AtomicSwapEther contract

willclarktech:
Part 1 of #818


scripts/ethereum/deployment/src/deploy_contracts.ts
```diff
@@ -40,6 +41,7 @@ export async function main(args: ReadonlyArray<string>): Promise<void> {
   // Order matters to get reproducible contract addresses
   const deploymentJobs: ReadonlyArray<DeploymentJob> = [
     { name: "AshToken", contract: new AshToken(eth) },
+    { name: "AtomicSwapEther", contract: new AtomicSwapEther(eth) },
```
webmaster128: This must come after TrashToken, otherwise the TrashToken is deployed using a different nonce, which changes the contract address

Issue #848, opened 2019-03-25
Parse erc20 transactions from chain as `bcp/send` tx type


ethanfrey:
The reverse of #814

When we find a contract info, we should use the same mapping to convert it back into a generic `bcp/send` tx, such that the same connector could re-encode it to the same state (if that didn't make sense, just that this should be the inverse of #814)

PR #871, opened 2019-04-01
Support sending ERC20 tokens

### LINKED ISSUE #813: Send ERC20 tokens

### LINKED ISSUE #814: Integrate ERC20 tokens with `bcp/send`
Given a map of `token name` to `token info` (address, sig figs, etc) we should be able to auto-convert a `bcp/send` transaction to a `ethereum/send_erc20` transaction and visa versa.

Add option data to the constructor (ethereum codec or connection or both???) to allow this to work transparently.

After this, we should be able to move erc20 tokens with a generic `bcp/send` transaction as well as a targeted send_erc20 transaction

webmaster128:
Closes #813 
Closes #814 

----------

Cannot yet be used in higher level API since `EthereumConnection.searchTransactionsById` > `ethereumCodec.parseBytes` do not yet support ERC20 transfer (see #848)

### DIFF
--- a/packages/iov-bcp/src/transactions.ts
+++ b/packages/iov-bcp/src/transactions.ts
@@ -217,6 +217,11 @@ export interface SendTransaction extends UnsignedTransaction {
   readonly amount: Amount;
   readonly recipient: Address;
   readonly memo?: string;
+  /**
+   * For smart-contract based token transfers, this is the address of the token contract.
+   * For native token transfers, this is undefined.
+   */
+  readonly contractAddress?: Address;
 }
 
 /** A swap offer or a counter offer */
--- a/packages/iov-bcp/src/transactions.ts
+++ b/packages/iov-bcp/src/transactions.ts
@@ -217,11 +217,6 @@ export interface SendTransaction extends UnsignedTransaction {
   readonly amount: Amount;
   readonly recipient: Address;
   readonly memo?: string;
-  /**
-   * For smart-contract based token transfers, this is the address of the token contract.
-   * For native token transfers, this is undefined.
-   */
-  readonly contractAddress?: Address;
 }
 
 /** A swap offer or a counter offer */
--- a/packages/iov-ethereum/src/abi.spec.ts
+++ b/packages/iov-ethereum/src/abi.spec.ts
@@ -54,6 +54,31 @@ describe("Abi", () => {
     });
   });
 
+  describe("encodeUint256", () => {
+    it("works", () => {
+      expect(Abi.encodeUint256("0")).toEqual(
+        fromHex("0000000000000000000000000000000000000000000000000000000000000000"),
+      );
+      expect(Abi.encodeUint256("1")).toEqual(
+        fromHex("0000000000000000000000000000000000000000000000000000000000000001"),
+      );
+      expect(Abi.encodeUint256("2")).toEqual(
+        fromHex("0000000000000000000000000000000000000000000000000000000000000002"),
+      );
+      expect(Abi.encodeUint256("123456789")).toEqual(
+        fromHex("00000000000000000000000000000000000000000000000000000000075bcd15"),
+      );
+    });
+
+    it("throws for invalid input", () => {
+      expect(() => Abi.encodeUint256("")).toThrow();
+      expect(() => Abi.encodeUint256(" 1")).toThrow();
+      expect(() => Abi.encodeUint256("-1")).toThrow();
+      expect(() => Abi.encodeUint256("0x1")).toThrow();
+      expect(() => Abi.encodeUint256("a")).toThrow();
+    });
+  });
+
   describe("decodeHeadTail", () => {
     it("works for single string", () => {
       const data = fromHex(
--- a/packages/iov-ethereum/src/abi.spec.ts
+++ b/packages/iov-ethereum/src/abi.spec.ts
@@ -68,6 +68,14 @@ describe("Abi", () => {
       expect(Abi.encodeUint256("123456789")).toEqual(
         fromHex("00000000000000000000000000000000000000000000000000000000075bcd15"),
       );
+      // 2^255
+      expect(
+        Abi.encodeUint256("57896044618658097711785492504343953926634992332820282019728792003956564819968"),
+      ).toEqual(fromHex("8000000000000000000000000000000000000000000000000000000000000000"));
+      // 2^256-1
+      expect(
+        Abi.encodeUint256("115792089237316195423570985008687907853269984665640564039457584007913129639935"),
+      ).toEqual(fromHex("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"));
     });
 
     it("throws for invalid input", () => {
@@ -76,6 +84,11 @@ describe("Abi", () => {
       expect(() => Abi.encodeUint256("-1")).toThrow();
       expect(() => Abi.encodeUint256("0x1")).toThrow();
       expect(() => Abi.encodeUint256("a")).toThrow();
+
+      // too large (2^256)
+      expect(() =>
+        Abi.encodeUint256("115792089237316195423570985008687907853269984665640564039457584007913129639936"),
+      ).toThrow();
     });
   });
 
--- a/packages/iov-ethereum/src/abi.spec.ts
+++ b/packages/iov-ethereum/src/abi.spec.ts
@@ -79,6 +79,59 @@ describe("Abi", () => {
     });
   });
 
+  describe("decodeAddress", () => {
+    it("works", () => {
+      expect(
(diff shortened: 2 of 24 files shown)


packages/iov-ethereum/src/abi.spec.ts
```diff
@@ -54,6 +54,31 @@ describe("Abi", () => {
...
+        fromHex("00000000000000000000000000000000000000000000000000000000075bcd15"),
+      );
+    });
+
+    it("throws for invalid input", () => {
+      expect(() => Abi.encodeUint256("")).toThrow();
+      expect(() => Abi.encodeUint256(" 1")).toThrow();
+      expect(() => Abi.encodeUint256("-1")).toThrow();
+      expect(() => Abi.encodeUint256("0x1")).toThrow();
+      expect(() => Abi.encodeUint256("a")).toThrow();
```
willclarktech: Number that's too large?

packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -11,14 +11,15 @@ import {
...
-import { ethereumCodec } from "./ethereumcodec";
+import { ethereumCodec, EthereumRpcTransactionResult } from "./ethereumcodec";
 
 const { fromHex } = Encoding;
 
 describe("ethereumCodec", () => {
   describe("parseBytes", () => {
     it("works", () => {
-      const rawGetTransactionByHashResult = {
+      // curl -sS -X POST --data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0x3b87faa3410f33284124a6898fac1001673f0f7c3682d18f55bdff0031cce9ce"],"id":1}' https://rinkeby.infura.io | jq
```
willclarktech: `... | jq .result`

scripts/ethereum/init.sh
```diff
@@ -6,7 +6,7 @@ export GANACHE_PORT="8545"
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-docker pull "alpine"
+# Don't pull before run. We can use whatever version of alpine is available locally.
```
willclarktech: How about a conditional pull if there's no local version?

packages/iov-bcp/src/transactions.ts
```diff
@@ -217,6 +217,11 @@ export interface SendTransaction extends UnsignedTransaction {
   readonly amount: Amount;
   readonly recipient: Address;
   readonly memo?: string;
+  /**
+   * For smart-contract based token transfers, this is the address of the token contract.
+   * For native token transfers, this is undefined.
+   */
+  readonly contractAddress?: Address;
```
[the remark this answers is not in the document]
webmaster128: This field will be removed again in favour of handling the ticker in `amount`

scripts/ethereum/init.sh
webmaster128: That happens automatically in `docker run`

packages/iov-ethereum/src/erc20.ts
```diff
@@ -11,12 +11,12 @@ export interface EthereumRpcClient {
 
 export interface Erc20Options {
   readonly contractAddress: Address;
-  /** Override on-chain symbol. Use this of contract does not define value on-chain */
-  readonly symbol?: string;
-  /** Override on-chain name. Use this of contract does not define value on-chain */
+  /** The token ticker. Overrides the on-chain value. */
+  readonly symbol: string;
+  /** The number of fractional digits. Overrides the on-chain value. */
+  readonly decimals: number;
```
willclarktech: Why is `name` optional but `decimals` and `symbol` not?

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -59,54 +103,76 @@ export const ethereumCodec: TxCodec = {
...
+    let send: SendTransaction;
+    if (shouldBeInterpretedAsErc20Transfer(input, decodeHexQuantityString(json.value))) {
+      const contractAddress = toChecksummedAddress(json.to);
+      const erc20Token = [...this.erc20Tokens.values()].find(
+        options => options.contractAddress.toLowerCase() === contractAddress.toLowerCase(),
+      );
+      if (!erc20Token) {
+        throw new Error(`No token configured for contract address ${contractAddress}`);
+      }
+      const quantity = Abi.decodeUint256(input.slice(4 + 32, 4 + 32 + 32));
```
willclarktech: Would be good to have an explanation of these magic numbers.

packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -30,24 +34,64 @@ import {
...
+  readonly r: string;
+  readonly s: string;
+  readonly to: string;
+  readonly transactionIndex: string;
+  readonly v: string;
+  readonly value: string;
+}
+
+export interface EthereumCodecOptions {
+  /** List of supported ERC20 tokens */
```
willclarktech: "List" is a little confusing. Also what does "supported" mean in this context? (Supported by whom/what?)

packages/iov-ethereum/src/serialization.ts
```diff
@@ -124,4 +201,20 @@ export class Serialization {
       return fromHex(normalizeHex(encodeQuantity(checkedNonce.toNumber())));
     }
   }
+
+  /**
+   * Value 0 must be represented as 0x instead of 0x0 for some strange reason
```
willclarktech: Source?

packages/iov-ethereum/src/utils.ts
```diff
@@ -73,3 +75,25 @@ export function fromBcpChainId(chainId: ChainId): number {
...
   return Uint53.fromString(rest).toNumber();
 }
+
+/**
+ * A function to determine if a transaction is interpreted as ERC20 transfer.
+ * We can not know for sure if it was a ERC20 call without knowledge of the recipient type,
+ * which is not available at the codec level.
+ */
+export function shouldBeInterpretedAsErc20Transfer(input: Uint8Array, ethQuantity: string): boolean {
+  if (ethQuantity !== "0") {
```
willclarktech: Is this definitely required?

packages/iov-ethereum/src/utils.ts
```diff
@@ -73,3 +75,25 @@ export function fromBcpChainId(chainId: ChainId): number {
...
+    return false;
+  }
+
+  if (input.length !== 4 + 32 + 32) {
```
willclarktech: Magic numbers?

scripts/ethereum/deployment/src/deploy_contracts.ts
```diff
@@ -53,18 +57,20 @@ export async function main(args: ReadonlyArray<string>): Promise<void> {
...
-      }
+      mintingJobs.push(
+        { contract: contract, recipient: mainIdentity, quantity: "100000000" /* 100 million atomics */ },
+        { contract: contract, recipient: secondIdentity, quantity: "33445566" },
+        { contract: contract, recipient: noEthAddress, quantity: "38" },
+      );
     }
   }
+
+  // Perform all minting jobs after all deployments to not change delployment nonces when updating minting
```
willclarktech: "Perform minting jobs after all deployments so that deployment nonces stay constant"?

packages/iov-ethereum/src/erc20.ts
webmaster128: Because we need `decimals` and `symbol` in the decoder (for `Amount`).

packages/iov-ethereum/src/serialization.ts
webmaster128: Unfortunatly I have no source for this. Just rial and error. Otherwise the calculated public key is wrong. Value 0 was never tested before. But this is at least consistent with the nonce case above.

packages/iov-ethereum/src/utils.ts
webmaster128: The whole guessing method is pointless since we have ERC20 contract addresses available in the codec now. Will cleanup after merge to get this in quicker

packages/iov-ethereum/src/utils.ts
webmaster128: shouldBeInterpretedAsErc20Transfer will be removed entirely

packages/iov-ethereum/src/ethereumcodec.ts
webmaster128:
```
  /**
   * ERC20 tokens supported by the codec instance.
   *
   * The behaviour of encoding/decoding transactions for other tokens is undefined.
   */
```

PR #872, opened 2019-04-02
Cleanup ERC 20 transfer parsing

### LINKED ISSUE #848: Parse erc20 transactions from chain as `bcp/send` tx type

webmaster128:
Closes #848

### DIFF
--- a/packages/iov-ethereum/src/ethereumcodec.spec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.spec.ts
@@ -168,5 +168,83 @@ describe("ethereumCodec", () => {
         otherSignatures: [],
       });
     });
+
+    it("interprets ERC20 transfer of unknown contract as ETH send", () => {
+      // There is only one transaction type on Ethereum. Smart contract interaction and
+      // ETH sends can only reliably be differentiated when all contract addresses are known.
+      // As a result, we interprete all unknown transactions as ETH send.
+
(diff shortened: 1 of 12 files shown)


packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -168,5 +168,83 @@ describe("ethereumCodec", () => {
...
+              tokenTicker: "ETH" as TokenTicker,
+            },
+          },
+          amount: {
+            quantity: "0",
+            fractionalDigits: 18,
+            tokenTicker: "ETH" as TokenTicker,
+          },
+          recipient: "0xc778417E063141139Fce010982780140Aa0cD5Ab" as Address,
+          // Non-UTF8 automatically represented as hex
```
willclarktech: What does this mean exactly? Is this memo in UTF8 or hex?

webmaster128: The thing is: it is not a memo. BCP uses Ethereum's `input` field to encode a memo as UTF-8. But the other way round we have no idea what kind of binary `input` data comes from Ethereum. So as soon as as receive `input` that is not a valid UTF-8 representation of a string, we represent that data as hex.

PR #874, opened 2019-04-02
Product fee support

### LINKED ISSUE #866: Add support for product fees in iov-bns
One piece that was not completed in #856

This is needed for registering user names in the wallet, but not blocking updating the faucet

ethanfrey:
Closes #866 

getFeeQuote() will also query the product fees for the specified transaction kind and use that instead of the minimum anti-spam fee if present.

* Updated the bnsd CI genesis file to add a 5 CASH product fee for registering username
* Added logic to query these values and override other quote if needed
* Updated tests

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -86,6 +86,12 @@ describe("BnsConnection", () => {
     fractionalDigits: 9,
     tokenTicker: cash,
   };
+  // this is enough money in an accout that registers names... twice the cost of one name registration product fee
+  const registerAmount: Amount = {
+    quantity: "10000000000",
+    fractionalDigits: 9,
+    tokenTicker: cash,
+  };
 
   // The first simple address key (m/4804438'/0') generated from this mnemonic produces the address
   // tiov1k898u78hgs36uqw68dg7va5nfkgstu5z0fhz3f (bech32) / b1ca7e78f74423ae01da3b51e676934d9105f282 (hex).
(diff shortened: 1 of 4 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -613,15 +614,12 @@ export class BnsConnection implements BcpAtomicSwapConnection {
...
+    const product = await this.getProductFee(transaction.kind);
     // take maximum
-    const res = await this.query("/", Encoding.toAscii("gconf:cash:minimal_fee"));
-    if (res.results.length !== 1) {
-      throw new Error("Received unexpected number of fees");
-    }
-    const data = Encoding.fromAscii(res.results[0].value);
-    const amount = decodeJsonAmount(data);
-    return { tokens: amount };
+    const fee = product || min;
```
webmaster128: I can see the elegance of this code, but in this case we should save the `getMinimumFee` call when `getProductFee` returns something.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -86,6 +86,12 @@ describe("BnsConnection", () => {
     fractionalDigits: 9,
     tokenTicker: cash,
   };
+  // this is enough money in an accout that registers names... twice the cost of one name registration product fee
```
webmaster128: accout -> account

packages/iov-bns/src/bnsconnection.ts
webmaster128: This implementation also means that the "min fee" is not a minimum fee but a default fee. In the case procuct fee < min fee, the fee is less than min fee.

ethanfrey:
@webmaster128 I did the cleanup, should be good to merge

Issue #816, opened 2019-03-06
Query transactions for erc20 token


ethanfrey:
Investigate what information is available from ethereum chain.

We want to allow `searchTx`, `listenTx`, and `liveTx` to detect erc20 token transfers when querying with the `sendFromOrTo` flag.

We should be able to use events/logs to query for this information. For now, consider we have access to a node that persists all logs. In the future, we may have to run our own node for ethereum mainnet/testnet.

TODO: sometime try out on mainnet and decide if this is good enough or we need to rely on etherscan there. For CI and local testnets this should work well without scrapper

PR #878, opened 2019-04-03
Various Ethereum improvements

webmaster128:
Pulled out of #816 to reduce diff

### DIFF
--- a/packages/iov-ethereum/src/abi.ts
+++ b/packages/iov-ethereum/src/abi.ts
@@ -14,9 +14,12 @@ export interface HeadTail {
 }
 
 export class Abi {
+  public static calculateMethodHash(signature: string): Uint8Array {
+    return new Keccak256(Encoding.toAscii(signature)).digest();
+  }
+
   public static calculateMethodId(signature: string): Uint8Array {
-    const hash = new Keccak256(Encoding.toAscii(signature)).digest();
-    return hash.slice(0, 4);
+    return Abi.calculateMethodHash(signature).slice(0, 4);
   }
 
   public static encodeAddress(address: Address): Uint8Array {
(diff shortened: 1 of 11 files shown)


packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -86,14 +90,21 @@ export class EthereumConnection implements BcpConnection {
...
-  private readonly erc20Tokens: ReadonlyMap<TokenTicker, Erc20>;
+  private readonly erc20Tokens: ReadonlyMap<TokenTicker, Erc20Options>;
+  private readonly erc20ContractReaders: ReadonlyMap<TokenTicker, Erc20>;
   private readonly codec: EthereumCodec;
 
   constructor(baseUrl: string, chainId: ChainId, options?: EthereumConnectionOptions) {
+    this.blockPollIntervalMs =
+      options && options.blockPollInterval ? options.blockPollInterval * 1000 : 4_000;
+    this.accountPollIntervalMs =
+      options && options.accountPollInterval ? options.accountPollInterval * 1000 : 5_000;
```
willclarktech: Why would these be different?

packages/iov-ethereum/src/abi.ts
```diff
@@ -14,9 +14,12 @@ export interface HeadTail {
 }
 
 export class Abi {
+  public static calculateMethodHash(signature: string): Uint8Array {
```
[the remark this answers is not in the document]
webmaster128: Yes, it is used to crearte a topic from a signature, which I need to search in logs

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: Not strictly required. They just used different values before. But it does not matter much. Should we reduce to one? I don't mind

willclarktech: It doesn't seem important, but since they both have essentially the same granularity (block time) it seems strange to me that they wouldn't have the same default interval.

webmaster128: Jap, let's simplify. With "poll interval" only we save a couple of lines

PR #880, opened 2019-04-04
Support ERC20 tokens in searchTx, listenTx and liveTx

### LINKED ISSUE #816: Query transactions for erc20 token

webmaster128:
Closes #816

### DIFF
--- a/packages/iov-ethereum/src/ethereumcodec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.ts
@@ -36,6 +36,12 @@ import {
   normalizeHex,
 } from "./utils";
 
+const methodCallPrefix = {
+  erc20: {
+    transfer: `0x${Encoding.toHex(Abi.calculateMethodId("transfer(address,uint256)"))}`,
+  },
+};
+
 /**
  * See https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionbyhash
  *
@@ -142,7 +148,7 @@ export class EthereumCodec implements TxCodec {
     );
 
     let send: SendTransaction;
-    if (erc20Token) {
+    if (erc20Token && json.input.startsWith(methodCallPrefix.erc20.transfer)) {
       const positionTransferMethodEnd = 4;
       const positionTransferRecipientBegin = positionTransferMethodEnd;
       const positionTransferRecipientEnd = positionTransferRecipientBegin + 32;
--- a/packages/iov-ethereum/src/ethereumcodec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.ts
@@ -147,7 +147,7 @@ export class EthereumCodec implements TxCodec {
       const positionTransferRecipientBegin = positionTransferMethodEnd;
       const positionTransferRecipientEnd = positionTransferRecipientBegin + 32;
       const positionTransferAmountBegin = positionTransferRecipientEnd;
-      const positionTransferAmountEnd = positionTransferRecipientEnd + 32;
+      const positionTransferAmountEnd = positionTransferAmountBegin + 32;
 
       const quantity = Abi.decodeUint256(input.slice(positionTransferAmountBegin, positionTransferAmountEnd));
       send = {
--- a/packages/iov-ethereum/src/ethereumconnection.spec.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.spec.ts
@@ -854,7 +854,7 @@ describe("EthereumConnection", () => {
       connection.disconnect();
     });
 
-    xit("lists ERC20 transactions when searching by sender", async () => {
+    it("lists ERC20 transactions when searching by sender", async () => {
       pendingWithoutEthereum();
 
       const connection = await EthereumConnection.establish(testConfig.base, {
@@ -910,7 +910,7 @@ describe("EthereumConnection", () => {
       }
 
       connection.disconnect();
-    });
+    }, 30_000);
   });
 
   describe("listenTx", () => {
--- a/packages/iov-ethereum/src/ethereumconnection.spec.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.spec.ts
@@ -1015,7 +1015,7 @@ describe("EthereumConnection", () => {
   });
 
   describe("liveTx", () => {
-    it("can listen to ETH transactions by recipient address (transactions in history and updates)", done => {
+    it("works for ETH transactions by recipient (in history and updates)", done => {
       pendingWithoutEthereum();
       pendingWithoutEthereumScraper();
 
@@ -1123,7 +1123,7 @@ describe("EthereumConnection", () => {
       })().catch(done.fail);
     }, 60_000);
 
-    it("can listen to ERC20 transactions by recipient address (transactions in history and updates)", done => {
+    it("works for ERC20 transactions by recipient (in history and updates)", done => {
       pendingWithoutEthereum();
 
       (async () => {
@@ -1227,7 +1227,7 @@ describe("EthereumConnection", () => {
       })().catch(done.fail);
     }, 60_000);
 
-    it("can listen to transactions by ID (transaction in history)", done => {
+    it("works for transactions by ID (in history)", done => {
       pendingWithoutEthereum();
       pendingWithoutEthereumScraper();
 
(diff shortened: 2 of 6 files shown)


packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -36,6 +36,12 @@ import {
   normalizeHex,
 } from "./utils";
 
+const methodCallPrefix = {
```
willclarktech: 👍

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -795,7 +772,124 @@ describe("EthereumConnection", () => {
...
+      });
+      const codec = new EthereumCodec({ erc20Tokens: testConfig.erc20Tokens });
+
+      const profile = new UserProfile();
+      const wallet = profile.addWallet(Secp256k1HdWallet.fromMnemonic(testConfig.mnemonic));
+      const mainIdentity = await profile.createIdentity(wallet.id, testConfig.chainId, HdPaths.ethereum(0));
+
+      const recipient = await randomAddress();
+
+      for (const sendTest of testConfig.erc20TransferTests) {
```
willclarktech: I find it odd that we're combining configuration options for a test context with test data in this way.

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -829,10 +925,10 @@ describe("EthereumConnection", () => {
             expect(event.primarySignature.pubkey).toEqual(sender.pubkey);
 
             if (events.length === 3) {
-              // This assumes we get two transactions into one block
-              // A == B < C
-              expect(events[0].height).toEqual(events[1].height);
-              expect(events[2].height).toBeGreaterThan(events[1].height);
+              // The order of the events 0, 1, 2 does not necessarily correspons to the send
```
willclarktech: "correspond"

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -871,61 +967,61 @@ describe("EthereumConnection", () => {
...
+        transactionIds.add(postResultB.transactionId);
+        transactionIds.add(postResultC.transactionId);
+
+        // Wait for transaction success
         await postResultA.blockInfo.waitFor(info => !isBlockInfoPending(info));
         await postResultB.blockInfo.waitFor(info => !isBlockInfoPending(info));
-
-        // Post C
-        await connection.postTx(bytesToPostC);
+        await postResultC.blockInfo.waitFor(info => !isBlockInfoPending(info));
```
willclarktech: Suggest using a `Promise.all` here.

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -1017,19 +1113,244 @@ describe("EthereumConnection", () => {
...
+        const sendC: SendTransaction = {
+          kind: "bcp/send",
+          creator: sender,
+          recipient: recipientAddress,
+          fee: {
+            gasPrice: testConfig.gasPrice,
+            gasLimit: testConfig.gasLimit,
+          },
+          ...testConfig.erc20TransferTests[0],
+        };
```
willclarktech: Does this really need to be defined three times?

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -1017,19 +1113,244 @@ describe("EthereumConnection", () => {
...
+            expect(event.transaction.creator).toEqual(sender);
+            expect(event.primarySignature.pubkey).toEqual(sender.pubkey);
+
+            if (events.length === 3) {
+              const receivedIds = new Set(events.map(e => e.transactionId));
+              expect(receivedIds).toEqual(transactionIds);
+
+              subscription.unsubscribe();
+              connection.disconnect();
+              done();
```
willclarktech: Did you consider doing this Promise-style?

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -702,23 +826,106 @@ export class EthereumConnection implements BcpConnection {
...
+      }
     }
+
+    return out;
+  }
+
+  private async searchSendTransactionsByAddressInLogs(
+    address: Address,
+    minHeight: number,
+    maxHeight: number,
```
willclarktech: Could these have default values?

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -702,23 +826,106 @@ export class EthereumConnection implements BcpConnection {
...
+    // console.log(erc20TransferLogsResponse.result);
+    // console.log(
+    //   erc20TransferLogsResponse.result.map((row: any) => {
+    //     return {
+    //       height: decodeHexQuantity(row.blockNumber),
+    //       contractAddress: row.address,
+    //       topics: row.topics,
+    //     };
+    //   }),
+    // );
```
willclarktech: `console.log`s

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -702,23 +826,106 @@ export class EthereumConnection implements BcpConnection {
...
+        },
+      ],
+      id: 7,
+    });
+    if (isJsonRpcErrorResponse(erc20TransferLogsResponse)) {
+      throw new Error(JSON.stringify(erc20TransferLogsResponse.error));
+    }
+
+    if (!Array.isArray(erc20TransferLogsResponse.result)) {
+      throw new Error("Got unuepected type of result");
```
willclarktech: "Expected result to be an array"?

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -702,23 +826,106 @@ export class EthereumConnection implements BcpConnection {
...
+    //     return {
+    //       height: decodeHexQuantity(row.blockNumber),
+    //       contractAddress: row.address,
+    //       topics: row.topics,
+    //     };
+    //   }),
+    // );
+
+    const ids = erc20TransferLogsResponse.result.map(row => Parse.transactionId(row.transactionHash));
+    // query all in paralel
```
willclarktech: "parallel"

packages/iov-ethereum/src/ethereumconnection.spec.ts
webmaster128: Me too. But I do not have a great idea how to change that. I mean we can enforce a list length of 2 and can have a test for the primary and a test for the secondary token. Not sure how that scales.

packages/iov-ethereum/src/ethereumconnection.spec.ts
webmaster128: All 3 are indeed equal in this test. Given that there is no nice native object cloning, I prefer to roll out this code.

packages/iov-ethereum/src/ethereumconnection.spec.ts
webmaster128: Wrapping those event-based tests into promises hides a lot of the underlying detail and removes control, which is important for debugging. Thus all the liveTx/listenTx tests are written this way.

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: Why would we need them? All callers set the values explicitly, which makes this function simpler.

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: Thanks for suggesting an alternative!

packages/iov-ethereum/src/ethereumconnection.spec.ts
webmaster128: But what I find worse the current version is creating tests dynamically (call `describe` or `it` based on runtime data). We do this in iov-tendermint-rpc and it is very hard to follow the test code.

packages/iov-ethereum/src/ethereumconnection.spec.ts
willclarktech: I was assuming they were going to be consumed by a pure function, but maybe that's not a safe assumption to make.

Issue #849, opened 2019-03-25
Query atomic swap transactions


ethanfrey:
Find all transactions that relate to a given `swap id`, `sender` or `recipient`.

This should allow us to make queries similar to the bns implementation of atomic swap

webmaster128:
Covered by https://github.com/iov-one/iov-core/issues/818

Issue #819, opened 2019-03-06
Query atomic swap contact


ethanfrey:
Builds on #818 

Add queries for the state of the specific contract

webmaster128:
Covered by https://github.com/iov-one/iov-core/issues/818

PR #882, opened 2019-04-04
Ether atomic swap support

willclarktech:
Part of #818 
Submitting the PR now as it's somewhat blocked by other issues. E.g.
- https://github.com/iov-one/ethereum-smart-contracts/issues/22
- https://github.com/iov-one/iov-core/issues/819
- https://github.com/iov-one/iov-core/issues/849

### DIFF
--- a/packages/iov-bcp/types/transactions.d.ts
+++ b/packages/iov-bcp/types/transactions.d.ts
@@ -152,6 +152,7 @@ export declare function createTimestampTimeout(secondsFromNow: number): Timestam
 /** A swap offer or a counter offer */
 export interface SwapOfferTransaction extends UnsignedTransaction {
     readonly kind: "bcp/swap_offer";
+    readonly swapId?: SwapIdBytes;
     readonly amounts: ReadonlyArray<Amount>;
     readonly recipient: Address;
     /**
--- a/packages/iov-bcp/types/transactions.d.ts
+++ b/packages/iov-bcp/types/transactions.d.ts
@@ -167,6 +167,7 @@ export interface SwapOfferTransaction extends UnsignedTransaction {
      */
     readonly hash: Uint8Array;
     readonly memo?: string;
+    readonly contractAddress?: Address;
 }
 export interface SwapClaimTransaction extends UnsignedTransaction {
     readonly kind: "bcp/swap_claim";
--- a/packages/iov-bcp/types/transactions.d.ts
+++ b/packages/iov-bcp/types/transactions.d.ts
@@ -152,6 +152,11 @@ export declare function createTimestampTimeout(secondsFromNow: number): Timestam
 /** A swap offer or a counter offer */
 export interface SwapOfferTransaction extends UnsignedTransaction {
     readonly kind: "bcp/swap_offer";
+    /**
+     * The ID of the swap to aid coordination between the two parties.
+     *
+     * This should be generated randomly by the client to avoid collisions.
+     */
     readonly swapId?: SwapIdBytes;
     readonly amounts: ReadonlyArray<Amount>;
     readonly recipient: Address;
@@ -168,7 +173,6 @@ export interface SwapOfferTransaction extends UnsignedTransaction {
      */
     readonly hash: Uint8Array;
     readonly memo?: string;
-    readonly contractAddress?: Address;
 }
 export interface SwapClaimTransaction extends UnsignedTransaction {
     readonly kind: "bcp/swap_claim";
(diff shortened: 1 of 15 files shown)


packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -636,6 +644,79 @@ export class EthereumConnection implements BcpConnection {
...
+      if (isJsonRpcErrorResponse(swapsResponse)) {
+        throw new Error(JSON.stringify(swapsResponse.error));
+      }
+
+      if (swapsResponse.result === null) {
+        return [];
+      }
+
+      const senderBegin = 12;
+      const senderEnd = senderBegin + 20;
```
webmaster128: I think it is easier to understand the code if we slice them field by field. In this case 4 bytes method ID, than 32 bytes sender (zero padded). The address can be extracted from a 32 bytes field using Abi.decodeAddress.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -636,6 +644,79 @@ export class EthereumConnection implements BcpConnection {
...
+      const recipientBegin = senderEnd + 12;
+      const recipientEnd = recipientBegin + 20;
```
webmaster128: same as for sender

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -636,6 +644,79 @@ export class EthereumConnection implements BcpConnection {
...
+      const hashBegin = recipientEnd;
+      const hashEnd = hashBegin + 32;
+      const timeoutBegin = hashEnd;
+      const timeoutEnd = timeoutBegin + 32;
+      const amountBegin = timeoutEnd;
+      const amountEnd = amountBegin + 32;
+
+      const resultArray = Encoding.fromHex(swapsResponse.result.slice(2));
```
webmaster128: `normalizeHex` can be used to convert Ethereum hex -> normal hex

packages/iov-ethereum/src/serialization.ts
```diff
@@ -108,6 +110,59 @@ export class Serialization {
...
+
+      if (!isValidAddress(unsigned.recipient)) {
+        throw new Error("Invalid recipient address");
+      }
+
+      if (unsigned.memo) {
+        throw new Error("Memo cannot be serialized in an atomic swap offer transaction");
+      }
+
+      if (!(unsigned.timeout as BlockHeightTimeout).height) {
```
webmaster128: `!isBlockHeightTimeout(unsigned.timeout)` should make the cast below unnecessary

packages/iov-ethereum/src/serialization.ts
```diff
@@ -184,6 +239,70 @@ export class Serialization {
...
```
webmaster128: `isBlockHeightTimeout`

packages/iov-ethereum/src/serialization.ts
```diff
@@ -108,6 +110,59 @@ export class Serialization {
...
+        throw new Error("Timeout must be specified as a block height");
+      }
+
+      if (unsigned.amounts.length !== 1 || unsigned.amounts[0].tokenTicker !== constants.primaryTokenTicker) {
+        // TODO: Remove when ERC20 swaps are supported
+        throw new Error("Only ETH atomic swap offers are currently supported");
+      }
```
webmaster128: `unsigned.amounts.length !== 1` should have its own check since we have no plan to support multiple amounts in Ethereum any time soon, so it should stay.

packages/iov-ethereum/src/serialization.ts
```diff
@@ -184,6 +239,70 @@ export class Serialization {
...
```
webmaster128: `unsigned.amounts.length !== 1` should have its own check since we have no plan to support multiple amounts in Ethereum any time soon, so it should stay.

packages/iov-ethereum/src/ethereumconnection.ts
willclarktech: Sounds good but there's no method ID in the return value.

webmaster128: Yeah right. The number 12 is so unusual in computer science that I was confused.

packages/iov-bcp/types/transactions.d.ts
```diff
@@ -167,6 +168,7 @@ export interface SwapOfferTransaction extends UnsignedTransaction {
      */
     readonly hash: Uint8Array;
     readonly memo?: string;
+    readonly contractAddress?: Address;
```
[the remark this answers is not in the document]
willclarktech: Yep, much neater.

PR #884, opened 2019-04-08
Add AtomicSwapHelpers.createId()

webmaster128:


packages/iov-bcp/src/atomicswaphelpers.spec.ts
```diff
@@ -10,6 +10,7 @@ describe("AtomicSwapHelpers", () => {
     it("works", async () => {
       const preimage = await AtomicSwapHelpers.createPreimage();
       expect(preimage.length).toEqual(32);
+      expect(new Set(preimage).size).toBeGreaterThanOrEqual(25);
```
willclarktech: How did you pick 25? Also how about a test for more than one result being different?

webmaster128: > How did you pick 25?

It is from trial and error. I got no value less than 27 in all my manual testing. So 25 is stable and prevents us from a horribly broken implementation. This does not even try to test randomness.

> Also how about a test for more than one result being different?

Yeah, why not. Will add.

PR #900, opened 2019-04-10
Create stream operator to dedup transaction events

### LINKED ISSUE #890: @iov/stream: create helper to deduplicate events
In BNS and Ethereum, we have this or similar code

```ts
      const mergedStream = Stream.merge(Stream.create(fromScraperProducer), Stream.create(fromLogsProducer));

      // remove duplicates
      const alreadySent = new Set<TransactionId>();
      const deduplicatedStream = mergedStream
        .filter(ct => !alreadySent.has(ct.transactionId))
        .debug(ct => alreadySent.add(ct.transactionId));

      return deduplicatedStream;
```

This should be extracted into a stream helper.

webmaster128:
Closes #890

### DIFF
--- a/packages/iov-stream/src/dropduplicates.ts
+++ b/packages/iov-stream/src/dropduplicates.ts
@@ -17,7 +17,7 @@ export type SameTypeStreamOperator<T> = (ins: Stream<T>) => Stream<T>;
  * -1-----2-4-3------
  * ```
  *
- * Each value must be uniquely identiified by a string given by
+ * Each value must be uniquely identified by a string given by
  * valueToKey(value).
  *
  * Internally this maintains a set of keys that have been processed already,
--- a/packages/iov-stream/src/dropduplicates.ts
+++ b/packages/iov-stream/src/dropduplicates.ts
@@ -1,0 +1,38 @@
+import { Stream } from "xstream";
+
+/**
+ * The type that fits into Stream.compose() with input stream and output stream
+ * of the same type.
+ */
+export type SameTypeStreamOperator<T> = (ins: Stream<T>) => Stream<T>;
+
+/**
+ * Drops duplicate values in a stream.
+ *
+ * Marble diagram:
+ *
+ * ```text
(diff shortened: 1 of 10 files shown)


packages/iov-stream/src/dropduplicates.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Marble diagram:
+ *
+ * ```text
+ * -1-1-1-2-4-3-3-4--
+ *   dropDuplicates
+ * -1-----2-4-3------
+ * ```
+ *
+ * Each value must be uniquely identiified by a string given by
```
willclarktech: "identified"

packages/iov-stream/src/dropduplicates.ts
```diff
@@ -0,0 +1,38 @@
...
+ * i.e. memory consumption and Set lookup times should be considered when
+ * using this function.
+ */
+export function dropDuplicates<T>(valueToKey: (x: T) => string): SameTypeStreamOperator<T> {
+  const operand: SameTypeStreamOperator<T> = (instream: Stream<T>): Stream<T> => {
+    const emittedKeys = new Set<string>();
+
+    const deduplicatedStream = instream
+      .filter(value => !emittedKeys.has(valueToKey(value)))
+      .debug(value => emittedKeys.add(valueToKey(value)));
```
willclarktech: I find it odd that we use a method called `debug` here, but it doesn't look like there's another method that does quite what we need here.

willclarktech: Bluebird has an analogous method for Promises called `.tap`, which I think is quite a nice name: http://bluebirdjs.com/docs/api/tap.html

webmaster128: Yeah, it is odd. I also looked for alternatives already but it turned out that `debug` is our friend. Good to have it hidden in the @iov/stream library now.

PR #901, opened 2019-04-10
Upgrade TypeScript to 3.3

webmaster128:
Unfortunately we need to wait for typedoc support to go higher than 3.3. But yeah, one step at a time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,8 @@
     "tslint": "^5.12.1",
     "tslint-config-prettier": "^1.13.0",
     "tslint-immutable": "^5.1.0",
-    "typedoc": "^0.14.2",
-    "typescript": "~3.2",
+    "typedoc": "^0.15.0-0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -48,8 +48,8 @@
     "tslint": "^5.12.1",
     "tslint-config-prettier": "^1.13.0",
     "tslint-immutable": "^5.1.0",
-    "typedoc": "^0.14.2",
-    "typescript": "~3.2",
+    "typedoc": "^0.15.0-0",
```
willclarktech: Do we really want to add this prerelease version?

webmaster128: I'd prefer not to but this is the only way we can upgrade TypeScript. Given that `typedoc` or its results are only used by maintainers, not by users of IOV-Core, I think it is okay to use that version as long as `yarn docs` does not fail.

PR #902, opened 2019-04-10
Extend EthereumConnection getSwaps

willclarktech:
Partially addresses #818

Several things to tidy up, but I could use a review at this point to confirm I'm going in the right direction.


packages/iov-ethereum/src/abi.ts
```diff
@@ -12,6 +12,11 @@ export interface HeadTail {
   /** Arguments split by positions as defined by head */
   readonly tail: ReadonlyArray<Uint8Array>;
 }
+export enum SwapContractEvent {
+  Opened = "Opened",
+  Claimed = "Claimed",
+  Aborted = "Aborted",
```
webmaster128: Are the raw values of this enum ever used? If now, we should drop them. If yes, we should document why the need to have a specific value.

packages/iov-ethereum/src/abi.ts
```diff
@@ -112,6 +117,35 @@ export class Abi {
...
+    if (state === undefined) {
+      throw new Error("Invalid swap process state");
+    }
+    return state;
+  }
+
+  public static decodeEventSignature(data: Uint8Array): SwapContractEvent {
+    const key = Encoding.toHex(data);
+    const map: { readonly [key: string]: SwapContractEvent } = {
+      "22f9086560da4f3a67d5fcc1a440655671d27a7e0884f260be3ce12ead52e156": SwapContractEvent.Opened,
```
webmaster128: The key can be replaced by `Encoding.toHex(Abi.calculateMethodHash("Opened(bytes32,address,address,bytes32,uint256,uint256)"))`

packages/iov-ethereum/src/abi.ts
```diff
@@ -112,6 +117,35 @@ export class Abi {
...
+    };
+    const state: SwapProcessState | undefined = map[key];
+
+    if (state === undefined) {
+      throw new Error("Invalid swap process state");
+    }
+    return state;
+  }
+
+  public static decodeEventSignature(data: Uint8Array): SwapContractEvent {
```
webmaster128: I'd add a length check to fail early in case of programming errors

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -33,8 +36,9 @@ import {
 import { Random, Secp256k1 } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
 import { HdPaths, Secp256k1HdWallet, UserProfile, WalletId } from "@iov/keycontrol";
-import { toListPromise } from "@iov/stream";
+import { asArray, toListPromise } from "@iov/stream";
```
webmaster128: unused import: `asArray`

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -133,8 +150,11 @@ export class EthereumConnection implements AtomicSwapConnection {
       this.socket.connect();
     }
 
+    const atomicSwapEtherContractAddress =
+      (options && options.atomicSwapEtherContractAddress) || constants.atomicSwapEtherContractAddress;
```
webmaster128: Can we remove `atomicSwapEtherContractAddress` from constants? It is never a @iov/ethereum-level constant but always network specific. If unset, atomicSwapEtherContractAddress can be `undefined`.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -740,14 +764,18 @@ export class EthereumConnection implements AtomicSwapConnection {
     return { ...transaction, fee: await this.getFeeQuote(transaction) };
   }
 
-  public async getSwaps(query: AtomicSwapQuery): Promise<ReadonlyArray<AtomicSwap>> {
+  public async getSwaps(
+    query: AtomicSwapQuery,
+    minHeight: number = 0,
+    maxHeight: number = Number.MAX_SAFE_INTEGER,
```
webmaster128: interesting compatible addition

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -740,14 +764,18 @@ export class EthereumConnection implements AtomicSwapConnection {
...
+  ): Promise<ReadonlyArray<AtomicSwap>> {
     if (isAtomicSwapIdQuery(query)) {
       const data = Uint8Array.from([...Abi.calculateMethodId("get(bytes32)"), ...query.swapid]);
 
       const params = [
         {
           to: constants.atomicSwapEtherContractAddress,
-          data: "0x" + Encoding.toHex(data),
+          data: `0x${Encoding.toHex(data)}`,
```
webmaster128: We really should have a `toEthereumHex` helper. I saw and wrote this code too often now.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -774,36 +802,160 @@ export class EthereumConnection implements AtomicSwapConnection {
...
-      const recipient = toChecksummedAddress(
-        Abi.decodeAddress(resultArray.slice(recipientBegin, recipientEnd)),
-      );
-      const hash = resultArray.slice(hashBegin, hashEnd) as Hash;
-      const timeoutHeight = new BN(resultArray.slice(timeoutBegin, timeoutEnd)).toNumber();
-      const amount = {
-        quantity: new BN(resultArray.slice(amountBegin, amountEnd)).toString(),
-        fractionalDigits: constants.primaryTokenFractionalDigits,
-        tokenTicker: constants.primaryTokenTicker,
+      const swapData = {
```
webmaster128: `const swapData: SwapData = {` for early type checking?

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -774,36 +802,160 @@ export class EthereumConnection implements AtomicSwapConnection {
...
       };
 
-      return [
+      const kind = Abi.decodeSwapProcessState(resultArray.slice(stateBegin, stateEnd));
+      let swap: AtomicSwap;
+      if (kind === SwapProcessState.Open) {
+        swap = ({
+          kind: kind,
+          data: swapData,
+        } as any) as OpenSwap;
```
webmaster128: why are those cast necessary?

packages/iov-ethereum/src/abi.ts
```diff
@@ -112,6 +117,35 @@ export class Abi {
     return data.slice(32, 32 + length);
   }
 
+  public static decodeSwapProcessState(data: Uint8Array): SwapProcessState {
```
webmaster128: The `Abi` class should not contain helpers for a specific kind of contract. But this we can refactor later.

packages/iov-ethereum/src/abi.ts
```diff
@@ -112,6 +117,35 @@ export class Abi {
...
+      "38d6042dbdae8e73a7f6afbabd3fbe0873f9f5ed3cd71294591c3908c2e65fee": SwapContractEvent.Claimed,
+    };
+    const event: SwapContractEvent | undefined = map[key];
+    if (event === undefined) {
+      throw new Error("Invalid event signature");
+    }
+
+    return event;
+  }
+
```
webmaster128: Test for those two new functions?

packages/iov-ethereum/src/abi.ts
willclarktech: That's how I calculated it in the first place, but I thought there was no need to recalculate it every time. Do you prefer a comment explaining how it's calculated or just do the calculation each time?

packages/iov-ethereum/src/ethereumconnection.ts
willclarktech: Definitely, that's been the plan all along, I just needed somewhere to put it while I got all the other bits sorted out.

packages/iov-ethereum/src/abi.ts
webmaster128: What about storing it a module level or class static constants? Then we have readable code and no significant runtime overhead

packages/iov-ethereum/src/ethereumconnection.ts
willclarktech: Should I update the type? And if so, would it be better to pass it an options object so we could specify a time for relevant blockchains instead?

packages/iov-ethereum/src/ethereumconnection.ts
willclarktech: I can add one, I'm also annoyed with it.

packages/iov-ethereum/src/ethereumconnection.ts
willclarktech: Yeah, this is the bit I'm most unhappy with. I haven't got to the bottom of why yet, and this `as any as X` stuff is definitely a hack that should not stay in.

packages/iov-ethereum/src/ethereumconnection.ts
webmaster128: If this turns out to be useful long term, the better place to put them is inside of `AtomicSwapQuery`, just like in `TransactionQuery`. For now this looks fine to me.

packages/iov-ethereum/src/abi.ts
willclarktech: Why not? The only reason we deal with ABI is to interact with a specific kind of contract.

webmaster128: Yes, but I think it would be good to have explicit dependencies, like:

`Abi`: lowe level ABI helpers for all smart contracts we support
`AtomicSwapAbi` (depends on `Abi`): ABI helpers specific to our atomic swap contracts
`Erc20Abi` (depends on `Abi`): ABI helpers specific all ERC20 contracts

This way you know that something is wrong when you import `AtomicSwapAbi` in ERC20 code. Also you prevent strange dependencies inside of `Abi`.

willclarktech: OK good, then I'll leave it for now and we can refactor.

packages/iov-ethereum/src/utils.ts
```diff
@@ -56,6 +56,15 @@ export function normalizeHex(input: string): string {
   }
 }
 
+/**
+ * Takes a hex representation optionally prefixed with 0x and returns an Ethereum-friendly
+ * representation with a prefix.
+ */
+export function toEthereumHex(input: string): string {
```
webmaster128: I thought this could take Uint8Array directly to avoid unnecessary string operations and shorter caller code in most places.

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -133,8 +155,10 @@ export class EthereumConnection implements AtomicSwapConnection {
       this.socket.connect();
     }
 
-    const erc20Tokens = options.erc20Tokens ? options.erc20Tokens : new Map();
+    const atomicSwapEtherContractAddress = options.atomicSwapEtherContractAddress || ("0x0" as Address);
```
webmaster128: Why this fallback when `this.atomicSwapEtherContractAddress` is optional? I prefer `undefined` over the dummy value.

packages/iov-ethereum/src/serialization.ts
```diff
@@ -57,7 +57,7 @@ export class Serialization {
     unsigned: UnsignedTransaction,
     nonce: Nonce,
     erc20Tokens: ReadonlyMap<TokenTicker, Erc20Options> = new Map(),
-    atomicSwapEtherContractAddress: Address = constants.atomicSwapEtherContractAddress,
```
webmaster128: better use `Address | undefined` and throw if required but unset

packages/iov-ethereum/src/serialization.ts
```diff
@@ -172,7 +198,7 @@ export class Serialization {
   public static serializeSignedTransaction(
     signed: SignedTransaction,
     erc20Tokens: ReadonlyMap<TokenTicker, Erc20Options> = new Map(),
-    atomicSwapEtherContractAddress: Address = constants.atomicSwapEtherContractAddress,
```
webmaster128: same here

PR #909, opened 2019-04-15
Format .json and .md in scripts/

webmaster128:


package.json
```diff
@@ -18,7 +18,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "format": "lerna run format",
-    "format-text": "prettier --write --prose-wrap always --print-width 80 \"./*.md\" \"./docs/**/*.md\" && lerna run format-text",
```
willclarktech: This includes the generated files in `scripts/ethereum/deployment/contracts`. Do we want to format those?

webmaster128: Good point. The mid term strategy is to move `scripts/ethereum/deployment` into a separate repo. Since no file inside of `scripts/ethereum/deployment/contracts` is modified right now, I think we can ignore the issue for some time.

PR #912, opened 2019-04-16
Support swap abort txs in Ethereum

willclarktech:
Partially addresses #818 

Additional Abi code can be refactored out later.

### DIFF
--- a/packages/iov-ethereum/src/ethereumcodec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.ts
@@ -9,6 +9,7 @@ import {
   Nonce,
   PostableBytes,
   PrehashType,
+  Preimage,
   PublicIdentity,
   PublicKeyBytes,
   SendTransaction,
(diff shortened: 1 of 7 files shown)


packages/iov-ethereum/src/ethereumcodec.ts
```diff
@@ -170,43 +173,71 @@ export class EthereumCodec implements TxCodec {
...
-        kind: "bcp/swap_offer",
-        creator: creator,
-        swapId: swapId,
-        fee: fee,
-        amounts: [
-          {
-            quantity: decodeHexQuantityString(json.value),
-            fractionalDigits: constants.primaryTokenFractionalDigits,
-            tokenTicker: constants.primaryTokenTicker,
+      if (method === SwapContractMethod.Open) {
```
webmaster128: I think a switch over `method` is slightly easier to follow, but this is good as well

willclarktech: Yeah I thought about that too. I'll add it in the next PR.

PR #913, opened 2019-04-16
Add ETH-ETH atomic swap integration test using same contract on same blockchain

### LINKED ISSUE #818: Create transactions for simple ETH-ETH atomic swap (one chain)

willclarktech:
Closes #818


packages/iov-ethereum/src/integrationtests/atomicswap.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+    const wallet = profile.addWallet(Secp256k1HdWallet.fromMnemonic(mnemonic));
+    const connection: AtomicSwapConnection = await EthereumConnection.establish(
+      testConfig.base,
+      testConfig.connectionOptions,
+    );
+    const coinType = 60;
+    const account = 0;
+    const change = 0;
+    const path1 = HdPaths.bip44(coinType, account, change, addressIndex1);
+    const path2 = HdPaths.bip44(coinType, account, change, addressIndex2);
```
webmaster128: `const path1 = HdPaths.ethereum(addressIndex1);`

packages/iov-ethereum/src/integrationtests/atomicswap.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+
+  public async claimFromRevealedPreimage(): Promise<void> {
+    // find claim
+    const claimSwaps = await this.connection.getSwaps({ sender: this.sendAddress });
+    const claim1 = claimSwaps[claimSwaps.length - 1];
+    expect(claim1.kind).toEqual(SwapProcessState.Claimed);
+
+    const offerSwaps = await this.connection.getSwaps({ recipient: this.receiveAddress });
+    const offer = offerSwaps[offerSwaps.length - 1];
+    expect(offer.kind).toEqual(SwapProcessState.Open);
```
webmaster128: I'm not super happy about the expect() calls inside of the `Actor` class. I know this works and why and – shame on me – I also did this in the other file. But I think this belongs into the `it` body directly or should be converted into an exception.

packages/iov-ethereum/types/integrationtests/atomicswap.spec.d.ts
```diff
@@ -0,0 +1 @@
+export {};
```
webmaster128: When you change the move-types script in `packages/iov-ethereum/package.json` to (copied from @iov/core)
```
"move-types": "shx rm -r ./types/* && shx mv build/types/* ./types && shx rm ./types/*.spec.d.ts && shx rm ./types/**/*.spec.d.ts",
```

this empty file should disappear. Unfortunatly ** does not work  for no subfolder, so there are two versions of that script.

packages/iov-ethereum/src/integrationtests/atomicswap.spec.ts
willclarktech: I definitely agree with this, I was just following the pattern of `iov-core`.

PR #915, opened 2019-04-17
Create JsonRpcSigningServer

### LINKED ISSUE #801: Create JsonRpcSigningServer wrapper around JsSigningServer
Requires #800

This does the same as JsSigningServer but adds a layer of JSON encoding/decoding

### LINKED ISSUE #914: Export isJsRpcErrorResponse from @iov/core

webmaster128:
Closes #801
Closes #914

---

JsonRpcSigningServer is like JsRpcSigningServer but works over every JSON-based channel. Unfortunately the [website browser extension communication](https://github.com/iov-one/ponferrada/wiki/Communication-channels) does not support the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) (website to WebWorker does), so we need a way to serialize/deserialize Uint8Array as JSON, which is done using "bytes:" prefixed hex strings.

Instead of wrapping JsonRpcSigningServer around JsRpcSigningServer, which adds mostly unnecessary complexity, I changed my mind and copied most of JsRpcSigningServer to prepare the removal of JsRpcSigningServer.

### DIFF
--- a/packages/iov-core/src/jsonrpcsigningserver.spec.ts
+++ b/packages/iov-core/src/jsonrpcsigningserver.spec.ts
@@ -1,13 +1,85 @@
 import {
   Address,
   Algorithm,
+  Amount,
   ChainId,
+  isConfirmedTransaction,
   isPublicIdentity,
+  PublicIdentity,
   PublicKeyBytes,
   SendTransaction,
   TokenTicker,
+  TransactionId,
 } from "@iov/bcp";
+import { bnsCodec, bnsConnector } from "@iov/bns";
+import { Ed25519, Random } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
+import { ethereumConnector } from "@iov/ethereum";
+import { isJsonRpcErrorResponse } from "@iov/jsonrpc";
+import { Ed25519HdWallet, HdPaths, Secp256k1HdWallet, UserProfile } from "@iov/keycontrol";
+import { firstEvent } from "@iov/stream";
+
+import { MultiChainSigner } from "./multichainsigner";
+import { GetIdentitiesAuthorization, SignAndPostAuthorization, SigningServerCore } from "./signingservercore";
+
+const { fromHex } = Encoding;
+
+function pendingWithoutBnsd(): void {
+  if (!process.env.BNSD_ENABLED) {
+    pending("Set BNSD_ENABLED to enable bnsd-based tests");
+  }
+}
+
+function pendingWithoutEthereum(): void {
+  if (!process.env.ETHEREUM_ENABLED) {
+    pending("Set ETHEREUM_ENABLED to enable ethereum-based tests");
+  }
+}
+
+async function randomBnsAddress(): Promise<Address> {
+  const rawKeypair = await Ed25519.makeKeypair(await Random.getBytes(32));
+  const randomIdentity: PublicIdentity = {
+    chainId: "some-testnet" as ChainId,
+    pubkey: {
+      algo: Algorithm.Ed25519,
+      data: rawKeypair.pubkey as PublicKeyBytes,
+    },
+  };
+  return bnsCodec.identityToAddress(randomIdentity);
+}
+
+const bnsdUrl = "ws://localhost:23456";
+const bnsdFaucetMnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
+const ethereumUrl = "http://localhost:8545";
+const ethereumChainId = "ethereum-eip155-5777" as ChainId;
+const ganacheMnemonic = "oxygen fall sure lava energy veteran enroll frown question detail include maximum";
+
+const defaultGetIdentitiesCallback: GetIdentitiesAuthorization = async (_, matching) => matching;
+const defaultSignAndPostCallback: SignAndPostAuthorization = async (_1, _2) => true;
+
+async function makeBnsEthereumSigningServer(): Promise<JsonRpcSigningServer> {
+  const profile = new UserProfile();
+  const ed25519Wallet = profile.addWallet(Ed25519HdWallet.fromMnemonic(bnsdFaucetMnemonic));
+  const secp256k1Wallet = profile.addWallet(Secp256k1HdWallet.fromMnemonic(ganacheMnemonic));
+  const signer = new MultiChainSigner(profile);
+
+  // connect to chains
+  const bnsConnection = (await signer.addChain(bnsConnector(bnsdUrl))).connection;
+  const ethereumConnection = (await signer.addChain(ethereumConnector(ethereumUrl, {}))).connection;
+
+  // faucet identity
+  await profile.createIdentity(ed25519Wallet.id, bnsConnection.chainId(), HdPaths.simpleAddress(0));
+  // ganache second identity
+  await profile.createIdentity(secp256k1Wallet.id, ethereumConnection.chainId(), HdPaths.bip44(60, 0, 0, 1));
+
+  const core = new SigningServerCore(
+    profile,
+    signer,
+    defaultGetIdentitiesCallback,
+    defaultSignAndPostCallback,
+  );
+  return new JsonRpcSigningServer(core);
+}
 
 import { JsonRpcSigningServer } from "./jsonrpcsigningserver";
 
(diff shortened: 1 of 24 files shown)


packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,185 @@
...
+        },
+      };
+      return errorResponse;
+    }
+
+    return this.handleChecked(checkedRequest);
+  }
+
+  /**
+   * Handles a checked JsonRpcRequest
```
willclarktech: What does "checked" mean exactly?

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,185 @@
...
+  public async handleChecked(request: JsonRpcRequest): Promise<JsonRpcResponse> {
+    let call: RpcCall;
+    try {
+      call = parseRpcCall(request);
+    } catch (error) {
+      const errorResponse: JsonRpcErrorResponse = {
+        jsonrpc: "2.0",
+        id: request.id,
+        error: {
+          code: jsonRpcCode.methodNotFound,
```
willclarktech: What if the error is a `ParamsError`?

packages/iov-core/src/signingservice.spec.ts
```diff
@@ -137,13 +139,14 @@ describe("signingservice.worker", () => {
...
-      fromHex("533e376559fa551130e721735af5e7c9fcd8869ddd54519ee779fce5984d7898"),
-    );
+    const result = TransactionEncoder.fromJson(response.result);
+    expect(result).toEqual(jasmine.any(Array));
+    expect((result as ReadonlyArray<any>).length).toEqual(1);
+    expect(result[0].chainId).toEqual(bnsConnection.chainId());
+    expect(result[0].pubkey).toEqual({
+      algo: Algorithm.Ed25519,
+      data: fromHex("533e376559fa551130e721735af5e7c9fcd8869ddd54519ee779fce5984d7898"),
+    });
```
willclarktech: As above re object comparison.

packages/iov-core/src/transactionencoder.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    it("works for dicts", () => {
+      expect(toJson({})).toEqual({});
+      expect(toJson({ foo: 1 })).toEqual({ foo: 1 });
+      expect(toJson({ foo: 1, bar: 2 })).toEqual({ foo: 1, bar: 2 });
+      expect(toJson({ foo: { bar: 1 } })).toEqual({ foo: { bar: 1 } });
+    });
+
+    it("fails for unsupported objects", () => {
+      expect(() => toJson(() => 0)).toThrowError(/Cannot encode type to JSON/i);
```
willclarktech: How about storing the regex in a variable so it's clear they all fail with the same message?

packages/iov-core/src/transactionencoder.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+      expect(() => toJson(/[0-9]+/)).toThrowError(/Cannot encode type to JSON/i);
+      expect(() => toJson(new Uint32Array())).toThrowError(/Cannot encode type to JSON/i);
+      expect(() => toJson(undefined)).toThrowError(/Cannot encode type to JSON/i);
+    });
+
+    it("prefixes strings", () => {
+      expect(toJson("")).toEqual("string:");
+      expect(toJson("abc")).toEqual("string:abc");
+      expect(toJson("\0")).toEqual("string:\0");
```
willclarktech: What about a string which starts `string:`?

packages/iov-core/src/transactionencoder.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+      expect(fromJson("string:abc")).toEqual("abc");
+      expect(fromJson("string:\0")).toEqual("\0");
+
+      // "bytes:" prefix
+      expect(fromJson("bytes:")).toEqual(new Uint8Array([]));
+      expect(fromJson("bytes:12")).toEqual(new Uint8Array([0x12]));
+      expect(fromJson("bytes:aabb")).toEqual(new Uint8Array([0xaa, 0xbb]));
+
+      // other prefixes
+      expect(() => fromJson("")).toThrowError(/Found string with unknown prefix/i);
```
willclarktech: As above re regex.

packages/iov-core/src/transactionencoder.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    });
+  });
+
+  describe("fromJson", () => {
+    it("works for numbers", () => {
+      expect(fromJson(0)).toEqual(0);
+      expect(fromJson(1)).toEqual(1);
+      expect(fromJson(-1)).toEqual(-1);
+      expect(fromJson(Number.POSITIVE_INFINITY)).toEqual(Number.POSITIVE_INFINITY);
+      expect(fromJson(Number.NaN)).toEqual(Number.NaN);
```
willclarktech: Numbers beyond safe range?

packages/iov-core/src/transactionencoder.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    it("prefixes and hex encodes Uint8Array", () => {
+      expect(toJson(new Uint8Array([]))).toEqual("bytes:");
+      expect(toJson(new Uint8Array([0x12]))).toEqual("bytes:12");
+      expect(toJson(new Uint8Array([0x12, 0xab]))).toEqual("bytes:12ab");
+    });
+
+    // Encoding recursive objects is explicitly undefined behavior. Just use this
+    // test to play around.
+    xit("fails for recursive objects", () => {
```
willclarktech: Does this explicitly undefined behaviour also cover recursive arrays?

packages/iov-core/src/transactionencoder.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (Array.isArray(data)) {
+      return data.map(TransactionEncoder.toJson);
+    }
+
+    // Exclude special kind of objects like Array, Date or Uint8Array
+    // Object.prototype.toString() returns a specified value:
+    // http://www.ecma-international.org/ecma-262/7.0/index.html#sec-object.prototype.tostring
+    if (
+      typeof data === "object" &&
+      data !== null &&
```
willclarktech: `null` check not strictly necessary because of above `return`.

packages/iov-core/src/transactionencoder.ts
```diff
@@ -0,0 +1,101 @@
...
+      Object.prototype.toString.call(data) === "[object Object]"
+    ) {
+      const out: any = {};
+      for (const key of Object.keys(data)) {
+        // tslint:disable-next-line: no-object-mutation
+        out[key] = TransactionEncoder.toJson((data as any)[key]);
```
willclarktech: `data as Object`?

packages/iov-core/src/transactionencoder.ts
```diff
@@ -0,0 +1,101 @@
...
+      return data;
+    }
+
+    if (data === null) {
+      return null;
+    }
+
+    if (typeof data === "string") {
+      if (data.startsWith("string:")) {
+        return data.slice(7);
```
willclarktech: Prefer a solution without magic numbers.

packages/iov-core/src/jsonrpcsigningserver.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+      algo: Algorithm.Ed25519,
+      data: fromHex("533e376559fa551130e721735af5e7c9fcd8869ddd54519ee779fce5984d7898"),
+    });
+    expect(result[1]).toEqual(ganacheSecondIdentity);
+
+    server.shutdown();
+    bnsConnection.disconnect();
+  });
+
+  it("send a signing request to service", async () => {
```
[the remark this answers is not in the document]
webmaster128: Oh, even worse. The unit under test we describe here does not _send_ but _handle_.

packages/iov-core/src/jsonrpcsigningserver.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+      params: {
+        reason: "Please sign",
+        transaction: TransactionEncoder.toJson(send),
+      },
+    });
+    if (isJsonRpcErrorResponse(signAndPostResponse)) {
+      throw new Error(`Response must not be an error, but got '${signAndPostResponse.error.message}'`);
+    }
+    const transactionId: TransactionId = TransactionEncoder.fromJson(signAndPostResponse.result);
+    expect(transactionId).toMatch(/^[0-9A-F]+$/);
```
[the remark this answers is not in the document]
webmaster128: Because the BCP implementation of BNS guarantees to use upper case hex for transaction IDs. This property is strictly required to allow string equality on BCP's `TransactionId`s.

packages/iov-core/src/jsonrpcsigningserver.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+    if (isJsonRpcErrorResponse(response)) {
+      throw new Error(`Response must not be an error, but got '${response.error.message}'`);
+    }
+    const result = TransactionEncoder.fromJson(response.result);
+    expect(result).toEqual(jasmine.any(Array));
+    expect((result as ReadonlyArray<any>).length).toEqual(1);
+    expect(result[0]).toEqual(ganacheSecondIdentity);
+
+    server.shutdown();
+  });
```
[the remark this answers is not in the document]
webmaster128: In this test there is no Ethereum testing connection (just the one inside of the server that is disconnected during shutdown). The testing connection in the test above only exists because the chain ID in weave local test chains is generated randomly. In Ethereum tests, we use a static chain ID.

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,185 @@
...
+  jsonRpcCode,
+  JsonRpcErrorResponse,
+  JsonRpcRequest,
+  JsonRpcResponse,
+  JsonRpcSuccessResponse,
+  parseJsonRpcId,
+  parseJsonRpcRequest,
+} from "@iov/jsonrpc";
+
+import { SigningServerCore } from "./signingservercore";
```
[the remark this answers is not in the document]
webmaster128: I agree. There is no specific reason why we ended up with all lower file names.

What about CamelCase file names that match the primary symbol?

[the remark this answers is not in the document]
willclarktech: CamelCase can cause issues, especially when combining git and OSX.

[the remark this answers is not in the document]
willclarktech: See eg https://coderwall.com/p/mgi8ja/case-sensitive-git-in-mac-os-x-like-a-pro

packages/iov-core/src/transactionencoder.spec.ts
webmaster128: Well spotted. This is supposed to be _circular_, not _recursive_.

packages/iov-core/src/transactionencoder.ts
webmaster128: Right. Unfortunately TypeScript's type inference is not great enough to know that. `typeof data === "object"` leads to `object | null` in this block, since `typeof null` is `object`.

packages/iov-core/src/transactionencoder.ts
webmaster128: Gives this error

> Element implicitly has an 'any' type because type 'Object' has no index signature.

packages/iov-core/src/jsonrpcsigningserver.ts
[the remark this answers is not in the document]
webmaster128: The other issue I am seing is that mixed case first letters can be hard to parse or even mess up the file order. Feel free  to open an issue with a better project wide solution.

packages/iov-core/src/transactionencoder.ts
webmaster128: In my experiance, the `object` type is pretty useless as it does not mean _any kind of object I don't know more about_, but basically means `{}`. I don't know the difference between `object` and `Object` but I read once that there are only very few cases where `Object` or `String` should be used as types.

packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -0,0 +1,185 @@
...
+    case "getIdentities": {
+      const { reason, chainIds } = data.params;
+      if (typeof reason !== "string") {
+        throw new ParamsError("1st parameter (reason) must be a string");
+      }
+      if (!Array.isArray(chainIds)) {
+        throw new ParamsError("2nd parameter (chainIds) must be an array");
+      }
+      if (!isArrayOfStrings(chainIds)) {
+        throw new ParamsError("Found non-string element in chainIds array");
```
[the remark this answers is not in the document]
webmaster128: The whole idea of positional indices is oudated as it turned out that JSON-RPC allows us to use names parameters. Will adapt the messages accordingly.

packages/iov-core/src/jsonrpcsigningserver.ts
webmaster128: Then we return jsonRpcCode.methodNotFound – why not! Nice catch

webmaster128:
@willclarktech all done

packages/iov-core/src/jsonrpcsigningserver.ts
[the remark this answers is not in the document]
willclarktech: https://github.com/iov-one/iov-core/issues/927

Issue #929, opened 2019-04-23
Let SigningServerCore.signAndPost return TransactionId | null


webmaster128:
Right now it is returns `TransactionId | undefined` but `undefined` cannot be converted to JSON explicitly.

PR #930, opened 2019-04-23
Test and implement `null` return in JsonRpcSigningServer/signAndPost

webmaster128:


packages/iov-core/src/jsonrpcsigningserver.ts
```diff
@@ -156,10 +156,12 @@ export class JsonRpcSigningServer {
           };
           break;
         case "signAndPost":
+          // Simplify when https://github.com/iov-one/iov-core/issues/929 is done
```
willclarktech: Why not do this directly?

webmaster128: Because #929 is a breaking change of the API (caller code like `result === undefined` would break). In this case nobody would care or notice but I am a bit semver-obsessed.

PR #945, opened 2019-04-25
Add ERC20 Swap Offer support

### LINKED ISSUE #821: Enable ERC20 atomic swaps (with ETH)
Builds on #818, #819 

This contract should expose the same interface as the ethereum contract, as much as possible, possibly needing multiple ethereum calls for one action on the other contract (approve the transferFrom, then create contract) 

Feel free to create a checklist or sub issues as needed to cover this. Simple as "do the same with ERC20" sounds, I understand this is multiple PRs

willclarktech:
Mammoth PR, sorry about that. I can split it into smaller ones if necessary.

Closes #821

Important things to consider:
1. Appropriate location/kind of Erc20ApproveTransaction
1. Appropriate location of SwapIdPrefix
1. How EthereumConnection.getSwaps should work if not querying by ID

### DIFF
--- a/packages/iov-ethereum/src/erc20.ts
+++ b/packages/iov-ethereum/src/erc20.ts
@@ -1,4 +1,4 @@
-import { Address, Amount, UnsignedTransaction } from "@iov/bcp";
+import { Address, Amount, TokenTicker, UnsignedTransaction } from "@iov/bcp";
 
 export interface Erc20ApproveTransaction extends UnsignedTransaction {
   readonly kind: "erc20/approve";
@@ -21,3 +21,5 @@ export interface Erc20Options {
   /** Override on-chain name. Use this if contract does not define value on-chain. */
   readonly name?: string;
 }
+
+export type Erc20TokensMap = ReadonlyMap<TokenTicker, Erc20Options>;
--- a/packages/iov-ethereum/src/erc20.ts
+++ b/packages/iov-ethereum/src/erc20.ts
@@ -1,4 +1,16 @@
-import { Address } from "@iov/bcp";
+import { Address, Amount, UnsignedTransaction } from "@iov/bcp";
+
+export interface Erc20ApproveTransaction extends UnsignedTransaction {
+  readonly kind: "erc20/approve";
+  readonly spender: Address;
+  readonly amount: Amount;
+}
+
+export function isErc20ApproveTransaction(
+  transaction: UnsignedTransaction,
+): transaction is Erc20ApproveTransaction {
+  return (transaction as Erc20ApproveTransaction).kind === "erc20/approve";
+}
 
 export interface Erc20Options {
   readonly contractAddress: Address;
--- a/packages/iov-ethereum/src/ethereumcodec.spec.ts
+++ b/packages/iov-ethereum/src/ethereumcodec.spec.ts
@@ -243,7 +243,6 @@ describe("ethereumCodec", () => {
 
     it("works for Ether atomic swap offer", () => {
       // Retrieved from local instance since we haven't deployed this to a public testnet
-      // curl - sS - X POST--data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0x044870acdb5fdab0f76266eda11cdeee50de880f58ccf6bfb32a6b651914f637"],"id":1}' http://localhost:8545 | jq .result
       const rawGetTransactionByHashResult: EthereumRpcTransactionResult = {
         hash: "0x044870acdb5fdab0f76266eda11cdeee50de880f58ccf6bfb32a6b651914f637",
         nonce: "0x36",
@@ -323,7 +322,6 @@ describe("ethereumCodec", () => {
 
     it("works for Ether atomic swap claim", async () => {
       // Retrieved from local instance since we haven't deployed this to a public testnet
-      // curl -sS -X POST --data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0x85bbdfabbf38e6888488f7e967a5b7784bc8041416773ecdfc2b57e365bc3777"],"id":1}' http://localhost:8545 | jq .result
       const rawGetTransactionByHashResult: EthereumRpcTransactionResult = {
         hash: "0x85bbdfabbf38e6888488f7e967a5b7784bc8041416773ecdfc2b57e365bc3777",
         nonce: "0x7e",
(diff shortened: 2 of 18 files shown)


packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -241,9 +242,85 @@ describe("ethereumCodec", () => {
...
+            0,
+          ).toFixedLength() as SignatureBytes,
+        },
+        otherSignatures: [],
+      });
+    });
+
     it("works for Ether atomic swap offer", () => {
       // Retrieved from local instance since we haven't deployed this to a public testnet
-      // curl - sS - X POST--data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0x044870acdb5fdab0f76266eda11cdeee50de880f58ccf6bfb32a6b651914f637"],"id":1}' http://localhost:8545 | jq .result
```
webmaster128: Why was this comment removed? Even if no reproducible, I find it good for documenting how the data was created and interesting to know what to call when we decide to use a different transaction.

packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -323,7 +401,6 @@ describe("ethereumCodec", () => {
 
     it("works for Ether atomic swap claim", async () => {
       // Retrieved from local instance since we haven't deployed this to a public testnet
-      // curl -sS -X POST --data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0x85bbdfabbf38e6888488f7e967a5b7784bc8041416773ecdfc2b57e365bc3777"],"id":1}' http://localhost:8545 | jq .result
```
webmaster128: same here

packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -390,7 +468,6 @@ describe("ethereumCodec", () => {
 
     it("works for Ether atomic swap abort", async () => {
       // Retrieved from local instance since we haven't deployed this to a public testnet
-      // curl -sS -X POST --data '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0xb8a6bdbcc56f30e385e928fee46374bceec1c6887814c0b4bddb23c8df25d91b"],"id":1}' http://localhost:8545 | jq .result
```
webmaster128: same here

packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -451,5 +529,243 @@ describe("ethereumCodec", () => {
         otherSignatures: [],
       });
     });
+
+    it("works for Erc20 atomic swap offer", () => {
+      // Retrieved from local instance since we haven't deployed this to a public testnet
```
webmaster128: same here

packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -451,5 +529,243 @@ describe("ethereumCodec", () => {
...
+            Encoding.fromHex("02555167f9584575f9740c0487dcdd2e1462a8374dd9dcc9e66cfa98eb1efd87"),
+            1,
+          ).toFixedLength() as SignatureBytes,
+        },
+        otherSignatures: [],
+      });
+    });
+
+    it("works for Erc20 atomic swap claim", async () => {
+      // Retrieved from local instance since we haven't deployed this to a public testnet
```
webmaster128: same here

packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -451,5 +529,243 @@ describe("ethereumCodec", () => {
...
+            Encoding.fromHex("578e59f1cfba37881d4851d0cb96eee287436fbe4a3a9ea4f6407c064f1e4f03"),
+            0,
+          ).toFixedLength() as SignatureBytes,
+        },
+        otherSignatures: [],
+      });
+    });
+
+    it("works for Erc20 atomic swap abort", async () => {
+      // Retrieved from local instance since we haven't deployed this to a public testnet
```
webmaster128: same here

packages/iov-ethereum/src/erc20.ts
```diff
@@ -9,3 +21,5 @@ export interface Erc20Options {
   /** Override on-chain name. Use this if contract does not define value on-chain. */
   readonly name?: string;
 }
+
+export type Erc20TokensMap = ReadonlyMap<TokenTicker, Erc20Options>;
```
webmaster128: Since this type is part of the @iov/ethereum public interface, it should be exported in `index.ts`

packages/iov-ethereum/src/serialization.ts
```diff
@@ -64,67 +68,43 @@ export class Serialization {
...
     const gasLimitHex = Serialization.getGasLimitHex(unsigned);
 
+    const args: [string, string, string, Nonce, Erc20TokensMap, Address?] = [
+      chainIdHex,
+      gasPriceHex,
+      gasLimitHex,
+      nonce,
+      erc20Tokens,
+      atomicSwapContractAddress,
+    ];
```
webmaster128: If `serializeUnsignedSendTransaction`, `serializeUnsignedSwapOfferTransaction`, `serializeUnsignedSwapClaimTransaction`, `serializeUnsignedSwapAbortTransaction`, `serializeUnsignedErc20ApproveTransaction` should share the same interface (which makes sense, looking at this code), I think we should pass an onject as the second paramter instead of creating dummy positional arguments.

packages/iov-ethereum/src/serialization.ts
```diff
@@ -138,73 +118,34 @@ export class Serialization {
       chainId > 0
         ? { forkState: BlknumForkState.Forked, chainId: chainId }
         : { forkState: BlknumForkState.Before };
-    const v = eip155V(chain, sig.recovery);
+    const v = encodeQuantity(eip155V(chain, sig.recovery));
+
+    const args: [string, Uint8Array, Uint8Array, string, string, Nonce, Erc20TokensMap, Address?] = [
```
webmaster128: Also here: make this an object to avoid creating dummy arguments

packages/iov-ethereum/src/serialization.ts
```diff
@@ -25,12 +24,17 @@ import { Abi } from "./abi";
...
 import { constants } from "./constants";
 import { BlknumForkState, Eip155ChainId, eip155V, toRlp } from "./encoding";
-import { Erc20Options } from "./erc20";
+import { Erc20ApproveTransaction, Erc20Options, Erc20TokensMap, isErc20ApproveTransaction } from "./erc20";
 import { encodeQuantity, encodeQuantityString, fromBcpChainId, normalizeHex } from "./utils";
 
 const { fromHex, toUtf8 } = Encoding;
 const ZERO_ETH_QUANTITY = "0";
 
+export enum SwapIdPrefix {
```
webmaster128: I think I'd move this to the ID generation. But it should not matter much.

willclarktech: That would create a circular dependency. We could move the ID generation here? Or to a separate file?

packages/iov-ethereum/src/serialization.ts
willclarktech: I thought about this but wasn't sure if we have a policy of when to use arguments vs objects?

webmaster128: There is no strict policy, but I'd say we use multiple positional arguments by default unless we have a very good reason. In this case we have a very good reason: 5 different function calls should be fed with arguments from a single source.

packages/iov-ethereum/src/serialization.ts
webmaster128: Good point. Let's just leave it here for now.

PR #947, opened 2019-04-29
Refactor AtomicSwapIdQuery and AtomicSwapHashlockQuery

### LINKED ISSUE #942: Capitalize ID in AtomicSwapIdQuery.swapid
Currently this is all lower case for some reason: `readonly swapid: SwapId;`

webmaster128:
Closes #942

### DIFF
--- a/packages/iov-bns/src/bnsconnection.ts
+++ b/packages/iov-bns/src/bnsconnection.ts
@@ -338,8 +338,8 @@ export class BnsConnection implements AtomicSwapConnection {
       } else if (isAtomicSwapRecipientQuery(query)) {
         return this.query("/escrows/recipient", decodeBnsAddress(query.recipient).data);
       } else {
-        // if (isQueryBySwapHash(query))
-        return this.query("/escrows/arbiter", hashIdentifier(query.hashlock));
+        // if (isAtomicSwapHashQuery(query))
+        return this.query("/escrows/arbiter", hashIdentifier(query.hash));
       }
     };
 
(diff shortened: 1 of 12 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -332,14 +332,14 @@ export class BnsConnection implements AtomicSwapConnection {
...
-        return this.query("/escrows", query.swapid.data);
+        return this.query("/escrows", query.id.data);
       } else if (isAtomicSwapSenderQuery(query)) {
         return this.query("/escrows/sender", decodeBnsAddress(query.sender).data);
       } else if (isAtomicSwapRecipientQuery(query)) {
         return this.query("/escrows/recipient", decodeBnsAddress(query.recipient).data);
       } else {
-        // if (isQueryBySwapHash(query))
-        return this.query("/escrows/arbiter", hashIdentifier(query.hashlock));
+        // if (isAtomicSwapHashQuery(query))
```
willclarktech: Why not make this logic explicit?

webmaster128: The type system ensures that in the `else` block, `query` is of type `AtomicSwapHashQuery`. If we eyplicitly add `else if (isAtomicSwapHashQuery(query))`, we create a new dead code branch with `query` of type `never`. Not sure if that would be nicer than the current code.

willclarktech: I think it would be easy to read and also guards against the situation where we develop the type further and don't update this code (we can give a nice error message rather than something about the `hash` property being missing).

webmaster128: ok, https://github.com/iov-one/iov-core/pull/947/commits/b80d111b38395a836ef2b255d0472886bf8d508d

PR #954, opened 2019-04-30
Migrate from jasmine2-custom-message to native withContext()

webmaster128:
Remove hardly used dependency


webmaster128:
> Cool feature

Good to have this natively for the cases where you really need to loop over test data, yeah. However, I don't link how it is integrated into the expect syntax as it breaks the reability of the tests (expect X with contect data number Y to equal Z).

PR #956, opened 2019-05-01
Add ERC20 support to getSwaps

### LINKED ISSUE #950: Fully support getSwaps in Ethereum connection
Currently our Ethereum connection's `getSwaps` implementation supports querying by ID for both ETH and ERC20 token swaps because the prefix will tell the connection which contract to query. For other types of query we don’t have this information. Options here include:
1. A flag to indicate which type of swap is desired, possibly by way of providing an ERC20 contract address to further filter results.
1. Getting both types of swaps and merging them together.
1. Adding additional view functions to the smart contracts themselves.
1. Separate `EthereumConnection` instances for Ether/ERC20 or even each ERC20 token.

willclarktech:
Closes #950

### DIFF
--- a/packages/iov-ethereum/src/ethereumconnection.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.ts
@@ -111,6 +111,10 @@ export interface EthereumLog {
   readonly topics: ReadonlyArray<string>;
 }
 
+interface EthereumLogWithPrefix extends EthereumLog {
+  readonly prefix: SwapIdPrefix;
+}
+
 export interface EthereumConnectionOptions {
   readonly wsUrl?: string;
   /** URL to an Etherscan compatible scraper API */
@@ -151,7 +155,11 @@ export class EthereumConnection implements AtomicSwapConnection {
     return new EthereumConnection(baseUrl, chainId, options);
   }
 
-  private static parseOpenedEventBytes(bytes: Uint8Array): OpenSwap {
+  private static parseOpenedEventBytes(
+    bytes: Uint8Array,
+    prefix: SwapIdPrefix,
+    erc20Tokens: Erc20TokensMap,
+  ): OpenSwap | null {
     const swapIdBegin = 0;
     const swapIdEnd = swapIdBegin + 32;
     const senderBegin = swapIdEnd;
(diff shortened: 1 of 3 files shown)


packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -138,6 +155,130 @@ export class EthereumConnection implements AtomicSwapConnection {
...
+    const swapIdBegin = 0;
+    const swapIdEnd = swapIdBegin + 32;
+
+    return {
+      kind: SwapProcessState.Aborted,
+      swapIdBytes: bytes.slice(swapIdBegin, swapIdEnd) as SwapIdBytes,
+    };
+  }
+
+  private static updateSwapInList(
```
webmaster128: It looks like this does the same as BCP's AtomicSwapMerger. Could you have a look at that?

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -138,6 +155,130 @@ export class EthereumConnection implements AtomicSwapConnection {
...
+    const recipientBegin = senderEnd;
+    const recipientEnd = recipientBegin + 32;
+    const hashBegin = recipientEnd;
+    const hashEnd = hashBegin + 32;
+    const amountBegin = hashEnd;
+    const amountEnd = amountBegin + 32;
+    const timeoutBegin = amountEnd;
+    const timeoutEnd = timeoutBegin + 32;
+    const erc20ContractAddressBegin = timeoutEnd;
+    const erc20ContractAddressEnd = erc20ContractAddressBegin + 32;
```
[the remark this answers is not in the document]
willclarktech: This is the contract where the token is defined. There's no equivalent for ETH. In that case, the returned data is just shorter.

packages/iov-ethereum/src/ethereumconnection.ts
willclarktech: So the difference is that `AtomicSwapMerger` is merging full transactions into a list of swaps, whereas what's happening here is that events from the smart contract are being merged into a list of swaps. Claimed/Aborted events have very minimal information compared to transactions.

webmaster128: Perfect, thanks for clarifying

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -138,6 +155,130 @@ export class EthereumConnection implements AtomicSwapConnection {
...
+
+    return {
+      kind: SwapProcessState.Open,
+      data: {
+        id: {
+          prefix: prefix,
+          data: bytes.slice(swapIdBegin, swapIdEnd) as SwapIdBytes,
+        },
+        sender: toChecksummedAddress(Abi.decodeAddress(bytes.slice(senderBegin, senderEnd))),
+        recipient: toChecksummedAddress(Abi.decodeAddress(bytes.slice(recipientBegin, recipientEnd))),
```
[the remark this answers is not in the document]
webmaster128: Could you rebase and update this now that #959 is in?

PR #958, opened 2019-05-01
Replace TSLint with ESLint

### LINKED ISSUE #876: Migrate to eslint
This should be done on the master branch at some point when there is no heavy development on a different branch anymore to avoid unnecessary merge work.

Prerequirements

- [x] ERC20 support in 0.13 completed
- [x] 0.13.x with ERC20 support integrated into wallet-demo
- [x] ERC20 in wallet-demo tested
- [x] 0.13.x integrated into [ponferrada](https://github.com/iov-one/ponferrada)
- [x] Merge https://github.com/iov-one/iov-core/pull/987 first to reduce diff of this PR
- [x] Merge obsolete `tslint_ci.json` into `tslint.json` and remove `TSLINT_FLAGS`
- [x] Base on branch 0.14, which we may need to support for some more time
- [x] Use configuration format that supports comments ~(yaml (preferred by me) or [js](https://github.com/iov-one/ponferrada/blob/master/.eslintrc.js))~ json
- [x] Start with any ruleset that leaves the current code unchanged
- [x] Ignore `docs/` as commented above
- [x] Turn disabled rules back on and fix errors
- [ ] Remove (or convert) "tslint:" comments

https://github.com/iov-one/ponferrada uses eslint for TypeScript already, so we can get some of the setup from there.

willclarktech:
WIP!
Closes #876

Before this can be merged, we should think about rulesets. In particular, I found no good replacement for tslint-immutable. There’s eslint-plugin-immutable, but it seems unmaintained, slightly buggy (the recommended preset refers incorrectly to redux), and only has a few JS-specific rules, as opposed to the rich TS-friendly rules in our current setup. (Note I've kept in the comments for exceptions to rules so we can see where they are.)


webmaster128:
> Before this can be merged, we should think about rulesets.

Some of the things that I am missing in [ponferrada](https://github.com/iov-one/ponferrada/) with eslint+TS:

- ordered imports
- readonly keyword
- no-floating-promises
- promise-function-async (typescript/promise-function-async seems to be equal)
- prefer-const

willclarktech:
Re `prefer-const`, `eslint-plugin-immutable` has `no-let` (also covers `var`): https://github.com/jhusain/eslint-plugin-immutable#no-let
Ordered imports we can enforce with `eslint-plugin-import`: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md
Re promises, this looks helpful: https://github.com/xjamundx/eslint-plugin-promise

webmaster128:
What about the following approach: migrate to eslint as it will be the future anyway, ensure the current codebase does not change by this (to avoid massive merge conflicts as long as we need to support 0.13) and improve our ruleset along the way. I think we should be careful with pulling in too many external dependencies (like plugins), if they are not maintained in a solid way or add little value.

Regarding ruleset:
* I prefer the native [prefer-const](https://eslint.org/docs/rules/prefer-const) over no-let, since the later treats mutable state unfairly and there is a bunch of very good code with mutable variables (like every for loop). `prefer-const` ensures that every change between mutable/immutable variable needs to be explicit.
* In addition to `prefer-const`, we should have [no-var](https://eslint.org/docs/rules/no-var). Looks like already included.
* In ponferrada we use `"@typescript-eslint/explicit-function-return-type": [ "error", { allowExpressions: true, }]`, which basically means: we soon as a function has a name, all types return type mit be set explicitly. I think we should set this.

.eslintignore
```diff
@@ -0,0 +1,3 @@
+build
+dist
+node_modules
```
webmaster128: Does this file also support the `/` prefix to indicate the names are directories?

`docs/` should also be excluded.

webmaster128:
As said in https://github.com/iov-one/iov-core/pull/958#issuecomment-490749647 let's just try to get this in. Working with eslint in other projects a little more, I a small wishlist before merge:

- [ ] Merge https://github.com/iov-one/iov-core/pull/987 first to reduce diff of this PR
- [ ] Base on branch 0.13, which we may need to sopport for some more time
- [ ] Use configuration format that supports comments ([yaml](https://github.com/webmaster128/amcl-js-module/blob/master/.eslintrc.yml) (preferred by me) or ([js](https://github.com/iov-one/ponferrada/blob/master/.eslintrc.js)))
- [ ] Start with any ruleset that leaves the current code unchanged
- [ ] Remove (or convert) "tslint:" comments
- [ ] Ignore `docs/` as commented above

webmaster128:
Because of the complexity of this migration, I'm moving my checklist into the original issue such that we can split it into multiple PRs

willclarktech:
Re comments in configuration files, eslint supports comments in JSON: https://eslint.org/docs/user-guide/configuring#comments-in-configuration-files

.eslintignore
willclarktech: I believe it does.

webmaster128:
Replaced by #1009

PR #961, opened 2019-05-02
Add LightTransaction types to BCP

### LINKED ISSUE #879: Design concept of LightTransaction
When querying transaction lists from a scraper or logs, not all Information for a full `ConfirmedTransaction<SendTransaction>` are available. Thus we send an extra query for each transaction in the list, which is very heavy.

Information missing in the Etherscan API

* signature and sender pubkey

Information available in an ERC20 transfer log:

```
  { logIndex: '0x0',
    transactionIndex: '0x0',
    transactionHash:
     '0xf27c3803eee1d6d95dc8c43f36f11ec28172ef1d8fc68463792c0274393e7ba9',
    blockHash:
     '0x55d11e7626d0d5ca25e3e98caa6d95c5d0b1cd74559cb62fb5ad3c88b85fcb82',
    blockNumber: '0x6',
    address: '0xf01231195ae56d38fa03f5f2933863a2606a6052',
    data:
     '0x0000000000000000000000000000000000000000000000000000000005f5e100',
    topics:
     [ '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef',
       '0x0000000000000000000000000000000000000000000000000000000000000000',
       '0x00000000000000000000000088f3b5659075d0e06bb1004be7b1a7e66f452284' ],
    type: 'mined' }
```

i.e.

* transaction ID
* height & transaction index, so we have the global ordering
* topics: method call, sender, recipient
* data: amount
* token ticker can be derived from contract adrress and method call

so in BCP speak this is an

* amount
* sender address (no sender pubkey)
* recipient address
* transaction ID
* height/confirmations

willclarktech:
Closes #879 

There’s no concept of hierarchy when extending interfaces, hence the duplicate `kind`s. Otherwise, it seems like all the other information specific to different transaction types would come back from a scraper via a `data` field. Not 100% about that though.

### DIFF
--- a/packages/iov-bcp/src/transactions.ts
+++ b/packages/iov-bcp/src/transactions.ts
@@ -223,11 +223,14 @@ export function isUnsignedTransaction(data: any): data is UnsignedTransaction {
   );
 }
 
-export interface SendTransaction extends UnsignedTransaction {
-  readonly kind: "bcp/send";
+interface SendTransactionProperties {
   readonly amount: Amount;
   readonly recipient: Address;
   readonly memo?: string;
+}
+
+export interface SendTransaction extends UnsignedTransaction, SendTransactionProperties {
+  readonly kind: "bcp/send";
 }
 
 export type SwapTimeout = BlockHeightTimeout | TimestampTimeout;
@@ -254,9 +257,7 @@ export function createTimestampTimeout(secondsFromNow: number): TimestampTimeout
   return { timestamp: Math.floor(ReadonlyDate.now() / 1000) + secondsFromNow };
 }
 
-/** A swap offer or a counter offer */
-export interface SwapOfferTransaction extends UnsignedTransaction {
-  readonly kind: "bcp/swap_offer";
+interface SwapOfferTransactionProperties {
   /**
    * The ID of the swap to aid coordination between the two parties.
    *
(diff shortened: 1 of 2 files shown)


webmaster128:
Hmm, this looks like a lot of repetition for every single transaction type. Can we do the following interface extension instead?

1. Create `LightTransaction` with all the  properties of `UnsignedTransaction` except `creator`
2. Create interface `TransactionWithCreator` that only has the old `creator`
3. Redefine `type UnsignedTransaction = LightTransaction & TransactionWithCreator`
4. Let all the specific transaction types extend `LightTransaction`

Then `SendTransaction` does not include the creator but `SendTransaction & TransactionWithCreator` does.

willclarktech:
That works, it just requires a lot of refactoring for us and probably frontend.

willclarktech:
That works but probably requires a lot of refactoring for us and frontend (adding in those `& TransactionWithCreator` bits).

webmaster128:
> That works but probably requires a lot of refactoring for us and frontend (adding in those & TransactionWithCreator bits).

Yeah, no problem. Having a clean API that allows is to query a list of 1000 transactions without sending 1001 requests is very important.

willclarktech:
Closing in favour of #975

PR #973, opened 2019-05-07
Add connection timeout to SocketWrapper and StreamingSocket 

### LINKED ISSUE #796: Add connection timeout to StreamingSocket
My local BNS blockchain is in some broken state at the moment:

```
{
  "jsonrpc": "2.0",
  "id": "",
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "open /tendermint/data/blockstore.db/000077.sst: no such file or directory"
  }
}
```

When I try to connect to it via WebSocket (iov-faucet), establishing the connection hangs forever. The line that blocks is

```ts
const tm = await TendermintClient.connect(url);
```

in `BnsConnection.establish`. When connecting via HTTP we get a proper HTTP 500 and an expected faucet crash.

So we need a connection timeout in `StreamingSocket`, such that `await this.socket.connected;` throws after some time.

```ts
public async execute(request: JsonRpcRequest): Promise<JsonRpcSuccess> {
    const pendingResponse = this.responseForRequestId(request.id);
    await this.socket.connected;
    const pendingSend = this.socket.send(JSON.stringify(request));

    const response = (await Promise.all([pendingResponse, pendingSend]))[0];
    return throwIfError(response);
}
```

webmaster128:
Closes #796

### DIFF
--- a/scripts/socketserver/echo.py
+++ b/scripts/socketserver/echo.py
@@ -1,12 +1,13 @@
 #!/usr/bin/env python3
 #pylint:disable=missing-docstring,invalid-name
 
+import argparse
 import asyncio
 import websockets
 import sys
+import time
 
 HOST = "localhost"
-PORT = 4444
 
 def log(data):
     print(data, flush=True)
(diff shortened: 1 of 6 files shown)


scripts/socketserver/echo.py
```diff
@@ -26,8 +27,23 @@ def connection_handler(connection, path):
...
+        help="Time in seconds that a connection will be delayed before establishing it",
+        type=int,
+        default=0)
+    args = parser.parse_args()
+
+    def delaying_process_request(path, request_headers):
+        time.sleep(args.delay)
+        return None
+
+    log("Starting server at {}:{}".format(HOST, args.port))
```
willclarktech: Why not make a `--host` option too?

webmaster128: No need to make the script more flexible than necessary

PR #975, opened 2019-05-07
Add LightTransaction type

### LINKED ISSUE #879: Design concept of LightTransaction

willclarktech:
Closes #879

As discussed [here](https://github.com/iov-one/iov-core/pull/961#issuecomment-489673192).

The only thing I'm not sure about with this approach is how to incorporate sender addresses (if at all).

### DIFF
--- a/packages/iov-bns/src/encode.spec.ts
+++ b/packages/iov-bns/src/encode.spec.ts
@@ -10,6 +10,7 @@ import {
   SendTransaction,
   SignatureBytes,
   TokenTicker,
+  WithCreator,
 } from "@iov/bcp";
 import { Ed25519, Ed25519Keypair, Sha512 } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
@@ -159,7 +160,7 @@ describe("Encode", () => {
     };
 
     it("can encode transaction without fees", () => {
-      const transaction: SendTransaction = {
+      const transaction: SendTransaction & WithCreator = {
         kind: "bcp/send",
         creator: defaultCreator,
         amount: defaultAmount,
@@ -176,7 +177,7 @@ describe("Encode", () => {
     });
 
     it("can encode transaction with fees", () => {
-      const transaction: SendTransaction = {
+      const transaction: SendTransaction & WithCreator = {
         kind: "bcp/send",
         creator: defaultCreator,
         amount: defaultAmount,
@@ -212,7 +213,7 @@ describe("Encode", () => {
     };
 
     it("works for SendTransaction", () => {
-      const transaction: SendTransaction = {
+      const transaction: SendTransaction & WithCreator = {
         kind: "bcp/send",
         creator: defaultCreator,
         amount: {
@@ -235,7 +236,7 @@ describe("Encode", () => {
     });
 
     it("works for AddAddressToUsernameTx", () => {
-      const addAddress: AddAddressToUsernameTx = {
+      const addAddress: AddAddressToUsernameTx & WithCreator = {
         kind: "bns/add_address_to_username",
         creator: defaultCreator,
         username: "alice",
@@ -251,7 +252,7 @@ describe("Encode", () => {
     });
 
     it("works for RegisterUsernameTx", () => {
-      const registerUsername: RegisterUsernameTx = {
+      const registerUsername: RegisterUsernameTx & WithCreator = {
         kind: "bns/register_username",
         creator: defaultCreator,
         username: "alice",
@@ -285,7 +286,7 @@ describe("Encode", () => {
     });
 
     it("works for RemoveAddressFromUsernameTx", () => {
-      const removeAddress: RemoveAddressFromUsernameTx = {
+      const removeAddress: RemoveAddressFromUsernameTx & WithCreator = {
         kind: "bns/remove_address_from_username",
         creator: defaultCreator,
         username: "alice",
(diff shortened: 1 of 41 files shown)


webmaster128:
Super nice

> The only thing I'm not sure about with this approach is how to incorporate sender addresses (if at all).

Good question. I think `SendTransaction` should have a sender address (shich we get from Etherscan, e.g. `curl -sS "https://api-rinkeby.etherscan.io/api?module=account&action=txlist&address=0x0c8184c21a51cdb7df9e5dc415a6a54b3a39c991" | jq`). However, this should be the money sender, not the transaction creator in BCP. If we need the signer address in general, I don't know.

willclarktech:
> `SendTransaction` should have a sender address

OK to open a separate issue for this? It might also apply to e.g. `SwapOpenTransaction`s? I'm also a bit concerned about duplicating information in the large number of cases where the transaction creator is the sender.

willclarktech:
CI is failing because of an unrelated issue. PR here: https://github.com/iov-one/iov-core/pull/977

webmaster128:
> OK to open a separate issue for this? It might also apply to e.g. SwapOpenTransactions? I'm also a bit concerned about duplicating information in the large number of cases where the transaction creator is the sender.

I'd prefer to have this question resolved before we merge this, since maybe we do not find a clever solution and may need to dump the idea of LightTransactions entirely. However, I don't have a great answer for this one.

It could be argued that a transaction creator is not necessarily the address that sends money, if you imagine a chain logic where it is leagal for account A to spend tokens of account B. Having creator and sender separately unfortunately increases the code required to create a transaction. But then again there is exactly 2 places in the wallet-demo that create transactions: 1. create a send transaction and 2. create a RegisterUsernameTx, so the large junk of work is on our side.

packages/iov-bns/src/encode.ts
```diff
@@ -142,7 +143,7 @@ function buildAddAddressToUsernameTx(tx: AddAddressToUsernameTx): codecImpl.app.
   };
 }
 
-function buildSendTransaction(tx: SendTransaction): codecImpl.app.ITx {
+function buildSendTransaction(tx: SendTransaction & WithCreator): codecImpl.app.ITx {
```
webmaster128: It would be nice if this function whould check that creator and sender match, preferably at byte level

```
const {prefix1, data1} = decodeBnsAddress(identityToAddress(tx.creator))
vs.
const {prefix2, data2} = decodeBnsAddress(tx.sender)
```

packages/iov-bns/src/encode.spec.ts
```diff
@@ -152,10 +153,11 @@ describe("Encode", () => {
     };
 
     it("can encode transaction without fees", () => {
-      const transaction: SendTransaction = {
+      const transaction: SendTransaction & WithCreator = {
         kind: "bcp/send",
         creator: defaultCreator,
         amount: defaultAmount,
+        sender: "tiov1dcg3fat5zrvw00xezzjk3jgedm7pg70y222af3" as Address,
```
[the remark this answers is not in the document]
willclarktech: This was copied from a helpful comment just above, but I can make it a variable.

packages/iov-bns/src/encode.ts
willclarktech: Do we want this? I thought the point was that the sender might not be the creator.

webmaster128: Right, I'd consider this a "not yet supported" error. We could also choose to support it but then we need to fix the implementation (`src` must be set to the sender, not the creator) and write a test.

willclarktech: Let's go with "not yet supported" for now and talk to Martin about implementation.

willclarktech: Does this also apply to other chains?

webmaster128: Yes. I guess on the other chains it is very unrealistic that sender != creator is possible so the error message can be permanent there.

PR #980, opened 2019-05-08
Fix timeout bug on disconnect() and connection error

webmaster128:


packages/iov-socket/src/socketwrapper.ts
```diff
@@ -142,4 +150,11 @@ export class SocketWrapper {
     }
     this.socket.send(data);
   }
+
+  private clearTimeout(): void {
+    if (!this.timeoutId) {
+      throw new Error("Timeout ID not set. This must not happen and usually means connect() was not called.");
```
willclarktech: "should not happen"

PR #985, opened 2019-05-08
Support multisignature contracts in BNS

### LINKED ISSUE #678: Support Multisig Middleware in BnsConnection
`x/multisig` allows us to create a multisig contract and update the owners per normal transactions. I also allows us to "activate" a multisig by including all needed signatures and then passing the multisig id in a middleware.

We should add support for those transactions and for the usage of multisig. This is an essential feature for launch, and even testnet security

willclarktech:
Closes #678 

Would be good to get some feedback before continuing. In particular, I'm unsure about the following:

1. Naming: to what extent should I be conforming to/diverging from the names in weave? Is `Contract` a term that only applies to these multisignature contracts for example?
1. Nullable values: should these be preserved in @iov/bns or should we enforce non-null values on our end? If we should enforce non-null values, should we provide defaults or throw?
1. Types: Should types from `codecImpl` be used by users of @iov/bns directly or does it make sense to export our own types? (E.g. `Participant` vs `codecImpl.multisig.IParticipant` in this PR.

### DIFF
--- a/packages/iov-bns/src/decode.spec.ts
+++ b/packages/iov-bns/src/decode.spec.ts
@@ -35,9 +35,12 @@ import {
   decodePrivkey,
   decodePubkey,
   isAddAddressToUsernameTx,
+  isCreateMultisignatureTx,
   isRegisterUsernameTx,
   isRemoveAddressFromUsernameTx,
+  isUpdateMultisignatureTx,
   Keyed,
+  Participant,
 } from "./types";
 
 const { fromHex, toUtf8 } = Encoding;
@@ -358,6 +361,72 @@ describe("Decode", () => {
       expect(parsed.username).toEqual("alice");
       expect(parsed.payload.chainId).toEqual("wonderland");
       expect(parsed.payload.address).toEqual("0xAABB001122DD");
+    });
+
+    it("works for CreateMultisignature", () => {
+      // tslint:disable-next-line:readonly-array
+      const participants: codecImpl.multisig.IParticipant[] = [
+        {
+          signature: fromHex("1234567890"),
+          power: 4,
+        },
+        {
+          signature: fromHex("abcdef0123"),
+          power: 1,
+        },
+        {
+          signature: fromHex("9999999999"),
+          power: 1,
+        },
+      ];
+      const transactionMessage: codecImpl.app.ITx = {
+        createContractMsg: {
+          participants: participants,
+          activationThreshold: 2,
+          adminThreshold: 3,
+        },
+      };
+      const parsed = parseMsg(defaultBaseTx, transactionMessage);
+      if (!isCreateMultisignatureTx(parsed)) {
+        throw new Error("unexpected transaction kind");
+      }
+      expect(parsed.participants).toEqual(participants as ReadonlyArray<Participant>);
+      expect(parsed.activationThreshold).toEqual(2);
+      expect(parsed.adminThreshold).toEqual(3);
+    });
+
+    it("works for UpdateMultisignature", () => {
+      // tslint:disable-next-line:readonly-array
+      const participants: codecImpl.multisig.IParticipant[] = [
+        {
+          signature: fromHex("1234567890"),
+          power: 4,
+        },
+        {
+          signature: fromHex("abcdef0123"),
+          power: 1,
+        },
+        {
+          signature: fromHex("9999999999"),
+          power: 1,
+        },
+      ];
+      const transactionMessage: codecImpl.app.ITx = {
+        updateContractMsg: {
+          contractId: fromHex("0123456789"),
+          participants: participants,
+          activationThreshold: 2,
+          adminThreshold: 3,
+        },
+      };
+      const parsed = parseMsg(defaultBaseTx, transactionMessage);
+      if (!isUpdateMultisignatureTx(parsed)) {
+        throw new Error("unexpected transaction kind");
+      }
+      expect(parsed.contractId).toEqual(fromHex("0123456789"));
+      expect(parsed.participants).toEqual(participants as ReadonlyArray<Participant>);
+      expect(parsed.activationThreshold).toEqual(2);
+      expect(parsed.adminThreshold).toEqual(3);
     });
   });
 });
--- a/packages/iov-bns/src/decode.spec.ts
+++ b/packages/iov-bns/src/decode.spec.ts
@@ -365,23 +365,37 @@ describe("Decode", () => {
 
     it("works for CreateMultisignature", () => {
       // tslint:disable-next-line:readonly-array
-      const participants: codecImpl.multisig.IParticipant[] = [
+      const iParticipants: codecImpl.multisig.IParticipant[] = [
         {
-          signature: fromHex("1234567890"),
+          signature: fromHex("1234567890123456789012345678901234567890"),
           power: 4,
         },
         {
-          signature: fromHex("abcdef0123"),
+          signature: fromHex("abcdef0123abcdef0123abcdef0123abcdef0123"),
           power: 1,
         },
         {
-          signature: fromHex("9999999999"),
+          signature: fromHex("9999999999999999999999999999999999999999"),
+          power: 1,
+        },
+      ];
+      const participants: ReadonlyArray<Participant> = [
+        {
+          address: "tiov1zg69v7yszg69v7yszg69v7yszg69v7ysy7xxgy" as Address,
+          power: 4,
+        },
+        {
+          address: "tiov140x77qfr40x77qfr40x77qfr40x77qfrj4zpp5" as Address,
+          power: 1,
+        },
+        {
+          address: "tiov1nxvenxvenxvenxvenxvenxvenxvenxverxe7mm" as Address,
           power: 1,
         },
       ];
       const transactionMessage: codecImpl.app.ITx = {
         createContractMsg: {
-          participants: participants,
+          participants: iParticipants,
           activationThreshold: 2,
           adminThreshold: 3,
         },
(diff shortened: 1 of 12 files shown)


ethanfrey:
> Naming: to what extent should I be conforming to/diverging from the names in weave? Is Contract a term that only applies to these multisignature contracts for example?

Feel free to pick the best name, no need to map them to bnsd names. I think iov-core is much more careful with clear names than weave, so feel free to change, and we may update on our end later as well.

ethanfrey:
> Nullable values: should these be preserved in @iov/bns or should we enforce non-null values on our end? If we should enforce non-null values, should we provide defaults or throw?

Multisigs as authorization should be a `ReadOnlyArray<MultisigId>` or such, for the backend, `null`, `undefined`, and `[]` all map to the same. Don't worry about that (actually for valid sigs, you will want to replace `[]` with `undefined|null`. However, good to enforce that no items in the array are null (only contains valid Ids (which are Uint8Array).

For the various tx to create and update the contract, make them with the same rules as other transactions. Actually better to add those two and the query functions before worrying about using them to authorize eg. send transactions. The first part is straight-forward architecture wise

ethanfrey:
> Types: Should types from codecImpl be used by users of @iov/bns directly or does it make sense to export our own types? (E.g. Participant vs codecImpl.multisig.IParticipant in this PR.

We try to never expose the internal types (protobuf, rlp, json) to the higher level callers, but use our own types for transactions, which make sense. They can be 1:1 mappings (like using Address, or Amounts), but nice to not couple these two tightly.

ethanfrey:
p.s. I have not done a code review, just responded to questions.
I am back from vacation and have an overflowing inbox... prs reviews are subject to triage. Ping me if it is necessary for me to review the code.

packages/iov-bns/src/decode.spec.ts
```diff
@@ -359,5 +362,71 @@ describe("Decode", () => {
       expect(parsed.payload.chainId).toEqual("wonderland");
       expect(parsed.payload.address).toEqual("0xAABB001122DD");
     });
+
+    it("works for CreateMultisignature", () => {
+      // tslint:disable-next-line:readonly-array
```
webmaster128: Why not use `readonly codecImpl.multisig.IParticipant[]`?

packages/iov-bns/src/decode.spec.ts
```diff
@@ -359,5 +362,71 @@ describe("Decode", () => {
...
+          participants: participants,
+          activationThreshold: 2,
+          adminThreshold: 3,
+        },
+      };
+      const parsed = parseMsg(defaultBaseTx, transactionMessage);
+      if (!isCreateMultisignatureTx(parsed)) {
+        throw new Error("unexpected transaction kind");
+      }
+      expect(parsed.participants).toEqual(participants.map(p => p as Participant));
```
webmaster128: No need to map I think: `participants as Participant[]` or `participants as readonly Participant[]`

packages/iov-bns/src/decode.spec.ts
```diff
@@ -359,5 +362,71 @@ describe("Decode", () => {
...
+      expect(parsed.activationThreshold).toEqual(2);
+      expect(parsed.adminThreshold).toEqual(3);
+    });
+
+    it("works for UpdateMultisignature", () => {
+      // tslint:disable-next-line:readonly-array
+      const participants: codecImpl.multisig.IParticipant[] = [
```
webmaster128: same here

packages/iov-bns/src/decode.spec.ts
```diff
@@ -359,5 +362,71 @@ describe("Decode", () => {
...
+          activationThreshold: 2,
+          adminThreshold: 3,
+        },
+      };
+      const parsed = parseMsg(defaultBaseTx, transactionMessage);
+      if (!isUpdateMultisignatureTx(parsed)) {
+        throw new Error("unexpected transaction kind");
+      }
+      expect(parsed.contractId).toEqual(fromHex("0123456789"));
+      expect(parsed.participants).toEqual(participants.map(p => p as Participant));
```
webmaster128: same here

packages/iov-bns/src/decode.ts
```diff
@@ -279,3 +286,42 @@ function parseRemoveAddressFromUsernameTx(
...
+
+function ensureParticipants(
+  maybe: ReadonlyArray<codecImpl.multisig.IParticipant> | null | undefined,
+): ReadonlyArray<Participant> {
+  const participants = ensure(maybe, "participants");
+  return participants.map(participant => {
+    if (participant.signature === null || participant.power === null) {
+      throw new Error("invalid participant in participants");
+    }
+    return participant as Participant;
```
webmaster128: Casts can be dangerous, especially when refactoring code later. Would this also work by making the map's return type explicit and let TypeScript check of `participant` matches the return type? Something like

```ts
  return participants.map((participant): Participant => {
    if (participant.signature === null || participant.power === null) {
      throw new Error("invalid participant in participants");
    }
    return participant;
  }
```

packages/iov-bns/src/encode.spec.ts
```diff
@@ -292,6 +299,64 @@ describe("Encode", () => {
...
+      ];
+      const createMultisignature: CreateMultisignatureTx = {
+        kind: "bns/create_multisignature_contract",
+        creator: defaultCreator,
+        participants: participants,
+        activationThreshold: 2,
+        adminThreshold: 3,
+      };
+      const msg = buildMsg(createMultisignature).createContractMsg!;
+      expect(msg.participants).toEqual(participants.map(p => p as codecImpl.multisig.IParticipant));
```
webmaster128: same here

packages/iov-bns/src/encode.spec.ts
```diff
@@ -292,6 +299,64 @@ describe("Encode", () => {
...
+        kind: "bns/update_multisignature_contract",
+        creator: defaultCreator,
+        contractId: fromHex("abcdef0123"),
+        participants: participants,
+        activationThreshold: 3,
+        adminThreshold: 4,
+      };
+      const msg = buildMsg(updateMultisignature).updateContractMsg!;
+      expect(msg.contractId).toEqual(fromHex("abcdef0123"));
+      expect(msg.participants).toEqual(participants.map(p => p as codecImpl.multisig.IParticipant));
```
webmaster128: same here

packages/iov-bns/src/encode.ts
```diff
@@ -142,6 +149,16 @@ function buildAddAddressToUsernameTx(tx: AddAddressToUsernameTx): codecImpl.app.
   };
 }
 
+function buildCreateMultisignatureTx(tx: CreateMultisignatureTx): codecImpl.app.ITx {
+  return {
+    createContractMsg: {
+      participants: tx.participants.map(participant => participant as codecImpl.multisig.IParticipant),
```
webmaster128: hmm, why this map? Can't `Participant` be assigned to `codecImpl.multisig.IParticipant` by default?

packages/iov-bns/src/encode.ts
```diff
@@ -217,3 +234,14 @@ function buildRemoveAddressFromUsernameTx(tx: RemoveAddressFromUsernameTx): code
     },
   };
 }
+
+function buildUpdateMultisignatureTx(tx: UpdateMultisignatureTx): codecImpl.app.ITx {
+  return {
+    updateContractMsg: {
+      contractId: tx.contractId,
+      participants: tx.participants.map(participant => participant as codecImpl.multisig.IParticipant),
```
webmaster128: same here

packages/iov-bns/src/types.ts
```diff
@@ -214,6 +214,18 @@ export interface AddAddressToUsernameTx extends UnsignedTransaction {
   readonly payload: ChainAddressPair;
 }
 
+export interface Participant {
+  readonly signature: Uint8Array | null;
+  readonly power: number | null;
```
webmaster128: What's the meaning of `signature: null` or `power: null`? I think this should be documented.

Maybe these cases never exist and the nullability is only for the protobuf type. In this case we should assert them away and expose non-nullable properties in IOV-Core.

packages/iov-bns/src/decode.spec.ts
willclarktech: Because `codecImpl.multisig.ICreateContractMsg` expects a non-readonly array: ` Type 'readonly IParticipant[]' is missing the following properties from type 'IParticipant[]': pop, push, reverse, shift, and 6 more.`

packages/iov-bns/src/types.ts
willclarktech: Based on this I think we can make them non-nullable: https://github.com/iov-one/weave/blob/cb5386a6e341eeb26237754a9928000825a72065/x/multisig/codec.proto#L23-L26

packages/iov-bns/src/decode.ts
willclarktech: Sadly not, even if I do the proper check and rule out `undefined`.

packages/iov-bns/src/decode.spec.ts
webmaster128: Interesting, thanks for checking

packages/iov-bns/src/types.ts
```diff
@@ -214,6 +214,18 @@ export interface AddAddressToUsernameTx extends UnsignedTransaction {
   readonly payload: ChainAddressPair;
 }
 
+export interface Participant {
+  readonly signature: Uint8Array;
```
webmaster128: Why the hack is this field called signature in the backend but constans an address?

Can we expose `readonly address: Address` here and hide all the strangeness?

willclarktech: Yep we can, the name comes from weave, it just means more conversion functions and less casting.

packages/iov-bns/src/types.ts
ethanfrey: They are required in the validation logic. Protobuf generation allows null and undefined, which map to missing fields in the binary representation. For higher-level types we can declare validity with the types.

Here are relevant lines that enforce it:

There must be between 1 and 100 participants: https://github.com/iov-one/weave/blob/cb5386a6e341eeb26237754a9928000825a72065/x/multisig/msg.go#L25-L33

Signature and Power must both be valid: https://github.com/iov-one/weave/blob/cb5386a6e341eeb26237754a9928000825a72065/x/multisig/msg.go#L66-L73 (and thresholds must be reasonable numbers)

Power must be between 1 and 255: https://github.com/iov-one/weave/blob/cb5386a6e341eeb26237754a9928000825a72065/x/multisig/model.go#L18-L34 (by the way this is renamed to weight in current master to be compatible with the type)

Signature (Address) must be 20 bytes: https://github.com/iov-one/weave/blob/cb5386a6e341eeb26237754a9928000825a72065/conditions.go#L206-L214

Fit as much as you can in the type system. The rest gladly in run-time checks (or just let the blockchain feed you errors)

packages/iov-bns/src/types.ts
ethanfrey: Please use proper names in the weave types. Then raise issues, so we improve the protobuf naming.

packages/iov-bns/src/types.ts
webmaster128: Thanks for the feedback

PR #1009, opened 2019-05-22
Add eslint support

willclarktech:
Replaces #958 

Open questions:
1. Any rule which caught things I turned *off* (except `no-console`, which I set to *warn* with some exceptions).
1. Currently eslint runs from the root package - if we're only planning to run this on CI, it might make sense to keep it that way, otherwise I'll add package-specific eslint scripts.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,8 @@
+node_modules/
+
+build/
+custom_types/
+dist/
+docs/
+generated/
+types/
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,6 @@
+{
+  "env": {
+    "node": true,
+    "es6": true
+  }
+}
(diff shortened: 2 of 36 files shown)


webmaster128:
> Any rule which caught things I turned off (except no-console, which I set to warn with some exceptions).

Good, since tslint is still running, this already makes things strictly better than before.

> Currently eslint runs from the root package - if we're only planning to run this on CI, it might make sense to keep it that way, otherwise I'll add package-specific eslint scripts.

I think `yarn lint` should run 1. eslint and 2. tslint consistently in all packages. This is not strictly needed in the development workflow since we have good linter support in the IDE, but let's you lint the code in the package you are working on manually if you have the desire to do so. So instead of `lint:eslint`, I'd update the "lint" implementation to

```
"lint": "eslint --max-warnings 1 \"**/*.{js,ts}\" && tslint -t verbose --project **."
```

PR #1013, opened 2019-05-23
Eslint rule: no-console

willclarktech:
Partially addresses #876

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,7 +18,7 @@
   ],
   "rules": {
     "curly": ["error", "multi-line", "consistent"],
-    "no-console": ["off", { "allow": ["error", "info", "warn"] }],
+    "no-console": ["warn", { "allow": ["error", "info", "warn"] }],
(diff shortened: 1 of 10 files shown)


.eslintrc.json
```diff
@@ -18,7 +18,7 @@
   ],
   "rules": {
     "curly": ["error", "multi-line", "consistent"],
-    "no-console": ["off", { "allow": ["error", "info", "warn"] }],
+    "no-console": ["error", { "allow": ["error", "info", "warn"] }],
```
webmaster128: can we make this a "warn", to get nice green lines instead of red ones when using console.log during development?

willclarktech: Sure

webmaster128: with `--max-warnings 0`, we don't really need the differentiation between warning and error at all. Unfortunately, it looks like you cannot just tell the elint plugin to highlight all lint issues in green (in contrast to the tslint plugin).

PR #1015, opened 2019-05-23
Eslint rule: spaced-comment

willclarktech:
Partially addresses #876 

I deleted some comments which seemed unnecessary.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,7 @@
     "no-console": ["warn", { "allow": ["error", "info", "warn"] }],
     "no-shadow": "error",
     "prefer-const": "error",
-    "spaced-comment": ["off", "always", { "exceptions": ["*"] }],
+    "spaced-comment": ["warn", "always", { "line": { "markers": ["/ <reference"] } }],
     "@typescript-eslint/array-type": "off",
     "@typescript-eslint/camelcase": "off",
     "@typescript-eslint/explicit-function-return-type": ["off", { "allowExpressions": true }],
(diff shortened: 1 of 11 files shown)


packages/iov-tendermint-rpc/src/responses.ts
```diff
@@ -259,7 +259,7 @@ export interface NodeInfo {
   readonly channels: string; // ???
   readonly moniker: string;
   readonly other: Map<string, string>;
-  /** Optional because this does not exist in Tendermint 0.25.x */
+  // Optional because this does not exist in Tendermint 0.25.x
```
webmaster128: Those diffs are a problem, since they convert tsdoc comments into regular comments

packages/iov-tendermint-rpc/src/responses.ts
```diff
@@ -151,7 +151,7 @@ export const getTxEventHeight = (event: TxEvent): number => event.height;
 export const getHeaderEventHeight = (event: NewBlockHeaderEvent): number => event.height;
 export const getBlockEventHeight = (event: NewBlockEvent): number => event.header.height;
 
-/**** Helper items used above ******/
+// Helper items used above
```
webmaster128: 👍

.eslintrc.json
```diff
@@ -21,7 +21,7 @@
     "no-console": ["warn", { "allow": ["error", "info", "warn"] }],
     "no-shadow": "error",
     "prefer-const": "error",
-    "spaced-comment": ["off", "always", { "exceptions": ["*"] }],
+    "spaced-comment": ["error", "always", { "line": { "markers": ["/ <reference"] } }],
```
webmaster128: warn

packages/iov-tendermint-rpc/src/responses.ts
willclarktech: Ah, didn't realise they were doing something. Will adjust.

PR #1016, opened 2019-05-23
Eslint rules: no-unused-vars and prefer-interface

willclarktech:
Partially addresses #876

I disabled rather than fixing for functions that throw a "not implemented" error.

### DIFF
--- a/packages/iov-ethereum/src/ethereumconnection.ts
+++ b/packages/iov-ethereum/src/ethereumconnection.ts
@@ -31,7 +31,6 @@ import {
   isSwapProcessStateClaimed,
   isSwapProcessStateOpen,
   Nonce,
-  OpenSwap,
   PostableBytes,
   PostTxResponse,
   Preimage,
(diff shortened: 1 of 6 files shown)


packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -974,6 +973,7 @@ export class EthereumConnection implements AtomicSwapConnection {
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
webmaster128: hmm, `_` or `_*` is the way to to express explicity that a variable is unused in TypeScript any many other languages. I think it would be better to adapt the linter https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md#argsignorepattern

packages/iov-core/src/signingservercore.spec.ts
```diff
@@ -356,7 +349,7 @@ describe("SigningServerCore", () => {
         await sendTokensFromFaucet(connection, identity, minimalFee);
       }
 
-      async function throwingCallback(_1: string, _2: UnsignedTransaction): Promise<boolean> {
+      async function throwingCallback(): Promise<boolean> {
```
webmaster128: To be honest, this those missing arguments, I find it much harder to see why the callbacks actually work for the interface. Also see the other option below

PR #1019, opened 2019-05-23
Eslint rules: curly, array-type, no-var-requires

willclarktech:
~Merge #1017 first~
Partially addresses #876

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,7 +17,6 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
-    "curly": ["error", "multi-line", "consistent"],
     "no-console": ["warn", { "allow": ["error", "info", "warn"] }],
     "no-shadow": "error",
     "prefer-const": "error",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,7 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
+    "curly": ["error", "all"],
     "no-console": ["warn", { "allow": ["error", "info", "warn"] }],
     "no-shadow": "error",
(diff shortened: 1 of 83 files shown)


.eslintrc.json
```diff
@@ -17,12 +17,10 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
-    "curly": ["error", "multi-line", "consistent"],
```
webmaster128: In the project I created this for, curly was not enabled by default. So this might make things less-strict. Could you verify this?

willclarktech: It will make things less strict. The problem is with the `multi-line` option: Prettier might split a line due to length, and then what was a `curly` pass is now a `curly` fail. How do you feel about the `all` option?

webmaster128: Is this really a problem?

Initial code

```
if (condition) long_expression()
```

After first prettier run

```
if (condition)
  long_expression()
```

After curly auto-fixer

```
if (condition) {
  long_expression()
}
```

So it should stabilize after 2 file saves.

-------

curly `all` is basically what we have in tslint right now. Sometimes I find it unnecessary that very short statemets are forced to take 3 lines instead of 1. But I got used to it.

The only scenario we definitely need to avoid is

```
if (condition)
  a()
```

which gets easily extended to

```
if (condition)
  a()
  b()
```

The problem is not so big with auto-formatting, but still.

willclarktech: I'll set it to `all` so we don't need to worry about saving twice/order effects, and if we decide the extra lines are annoying we can change it later.

webmaster128: Perfect, feel free to merge when done and green

PR #1020, opened 2019-05-23
Eslint rules camelcase/no-empty-interface

willclarktech:
Merge #1019 first
Partially addresses #876

We should consider whether we really need some of these empty interfaces.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,7 +17,6 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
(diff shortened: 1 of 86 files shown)


webmaster128:
> We should consider whether we really need some of these empty interfaces.

I don't think there is anything wrong with empty interfaces (which is why the rule is disabled in tslint). If we really change this setting, we should adapt `"no-empty-interface": false,` in tslint as well.

willclarktech:
@webmaster128 Removed

PR #1022, opened 2019-05-27
Remove local BCPd chain

webmaster128:


packages/iov-core/src/integrationtests/atomicswapbnsbns.spec.ts
```diff
@@ -216,47 +182,42 @@ class Actor {
   }
 }
 
-describe("Full atomic swap between bns and bcp", () => {
+describe("Full atomic swap between BNS and BNS", () => {
```
willclarktech: This sounds like it might be two chains. Suggest "between two different tokens on BNS"?

packages/iov-core/src/integrationtests/atomicswapbnsbns.spec.ts
```diff
@@ -27,7 +27,7 @@ import { Ed25519HdWallet, HdPaths, UserProfile } from "@iov/keycontrol";
 import { MultiChainSigner } from "../multichainsigner";
```
willclarktech: Maybe this file should be in the BNS package rather than Core? We have an analogous integration test in the Ethereum package.

webmaster128: Very good idea. It just tried that. Unfortunately, the test is based on the MultiChainSigner, which means a lot of refactoring is required to use lower-level equivalents (explicit signing and posting). So this is out of scope for this PR but definitely worth doing.  https://github.com/iov-one/iov-core/issues/1035

packages/iov-core/src/integrationtests/atomicswapbnsbns.spec.ts
webmaster128: Done in the latest commit

PR #1026, opened 2019-05-27
Add UserProfile.deriveEncryptionKey

### LINKED ISSUE #898: Improve UserProfile API for many store operations
When a caller needs to store the UserProfile very often (i.e. after every change), we should not run Argon2 on the password every single time.

webmaster128:
Closes #898 

~Based on #1024~


packages/iov-keycontrol/src/userprofile.spec.ts
```diff
@@ -37,6 +42,275 @@ describe("UserProfile", () => {
...
+      const wallet1 = Ed25519HdWallet.fromMnemonic(defaultMnemonic3);
+
+      const original = new UserProfile();
+      original.addWallet(wallet1);
+      original.setWalletLabel(wallet1.id, "My secret 😛");
+
+      await original.storeIn(db, defaultEncryptionPassword);
+      const restored = await UserProfile.loadFrom(db, defaultEncryptionPassword);
+
+      expect(restored.wallets.value).toEqual(original.wallets.value);
```
willclarktech: It's a bit confusing that different aspects of the profile are being checked, but the titles talk about eg special chars.

packages/iov-keycontrol/src/userprofile.spec.ts
```diff
@@ -37,6 +42,275 @@ describe("UserProfile", () => {
...
+
+      const createdAt = new ReadonlyDate("1985-04-12T23:20:50.521Z");
+      const keyring = new Keyring();
+      const original = new UserProfile({ createdAt: createdAt, keyring: keyring });
+
+      await original.storeIn(db, defaultEncryptionPassword);
+
+      const otherEncryptionPassword = "something wrong";
+      await UserProfile.loadFrom(db, otherEncryptionPassword)
+        .then(() => fail("loading must not succeed"))
```
willclarktech: What about passing two handlers to the `then` instead of `.then().catch()`? It represents the flow slightly more faithfully I think.

packages/iov-keycontrol/src/userprofile.spec.ts
```diff
@@ -37,6 +42,275 @@ describe("UserProfile", () => {
...
+
+    it("can load format version 1 profile with password", async () => {
+      const db = levelup(MemDownConstructor<string, string>());
+
+      // Storage data created with IOV-Core 0.14 cli
+      await db.put("format_version", userprofileData.serializations.version1.format_version);
+      await db.put("created_at", userprofileData.serializations.version1.created_at);
+      await db.put("keyring", userprofileData.serializations.version1.keyring);
+
+      const loaded = await UserProfile.loadFrom(db, "secret passwd");
```
willclarktech: Would be nice to have this stored in a more obvious place. Perhaps even in the JSON file.

packages/iov-keycontrol/src/userprofile.spec.ts
```diff
@@ -355,115 +629,95 @@ describe("UserProfile", () => {
...
-  });
+      const profile = new UserProfile();
+      await profile.storeIn(db, defaultEncryptionPassword);
 
-  it("throws when loading a profile with unsupported format version", async () => {
-    const db = levelup(MemDownConstructor<string, string>());
-    await db.put("format_version", "123");
+      await db
+        .get("foo")
+        .then(() => fail("get 'foo' promise must not reslve"))
```
willclarktech: "resolve"

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -27,16 +27,51 @@ const storageKeyCreatedAt = "created_at";
...
 };
+
+// not great but can be used on the main thread
+const passwordHashingOptionsVersion2: Argon2idOptions = {
+  outputLength: 32,
+  opsLimit: 11,
+  memLimitKib: 8 * 1024,
+};
+
 // A fixed salt is choosen to archive a deterministic password to key derivation.
```
willclarktech: "chosen"

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -27,16 +27,51 @@ const storageKeyCreatedAt = "created_at";
...
+  readonly formatVersion: number;
+  readonly data: Uint8Array;
+}
+
+/**
+ * An error class that allows handling an unexpected format version.
+ * It contains all the data needed to derive the encryption key in a different
+ * format version using UserProfile.deriveEncryptionKey.
+ */
+export class UserProfileEncryptionKeyUnexpectedFormatVersion extends Error {
```
willclarktech: It would be clearer if this name ended with `Error`

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -50,22 +85,64 @@ export interface UserProfileOptions {
...
+   * Use this function to cache the encryption key in memory.
+   *
+   * @param formatVersion Set this if you got a UserProfileEncryptionKeyUnexpectedFormatVersion error. This
+   * error usually means a profile was encrypted with an older format version.
+   */
+  public static async deriveEncryptionKey(
+    password: string,
+    formatVersion?: number,
+  ): Promise<UserProfileEncryptionKey> {
+    const version = formatVersion !== undefined ? formatVersion : latestFormatVersion;
```
willclarktech: How about using a default value instead?

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -104,20 +181,40 @@ export class UserProfile {
     this.wallets = new ValueAndUpdates(this.walletsProducer);
   }
 
-  // this will clear everything in the database and store the user profile
-  public async storeIn(db: LevelUp<AbstractLevelDOWN<string, string>>, password: string): Promise<void> {
+  /**
+   * Store this profile in database.
+   *
+   * This will clear everything in the database and store the user profile.
```
willclarktech: These two sentences can be combined.

packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -104,20 +181,40 @@ export class UserProfile {
...
+        ? await UserProfile.deriveEncryptionKey(encryptionSecret)
+        : encryptionSecret;
+    if (key.formatVersion !== storageFormatVersion) {
+      throw new UserProfileEncryptionKeyUnexpectedFormatVersion(storageFormatVersion, key.formatVersion);
+    }
+
+    // encrypt
+    const encryptedKeyring = await KeyringEncryptor.encrypt(this.keyring.serialize(), key.data);
 
     // create storage values (raw strings)
```
willclarktech: Are these comments helpful?

packages/iov-keycontrol/src/userprofile.spec.ts
webmaster128: Indeed. Basically what this test tries to do (and other tests as well) is: perform an equality check on UserProfile (`restored` equals `original`). So I see basically two options to fix this confusion:
1. create a userProfileEquals testing helper that checks all the state that can be checked using public interfaces.
2. only check the very specific data in question (`restored.wallets.value[0].label`)

packages/iov-keycontrol/src/userprofile.spec.ts
webmaster128: I did not know that .then can take more than one argument and this is how we wrote it in many many places in this project.

But this one is also fine:

```ts
      await UserProfile.loadFrom(db, otherEncryptionPassword).then(
        () => fail("loading must not succeed"),
        error => expect(error).toMatch(/invalid usage/),
      );
```

I tend to agree that your suggestion makes sense. But maybe we better do this as a bigger refactoring through the whole project.

packages/iov-keycontrol/src/userprofile.ts
webmaster128: I am very much surprised that I can use the module runtime constant `latestFormatVersion` as part of the function signature. But for some reason it works, great.

packages/iov-keycontrol/src/userprofile.ts
webmaster128: Nope, thanks for pointing that out. Will remove

webmaster128:
Updated, everything from "Fix typo in reslve" was added.

packages/iov-keycontrol/src/userprofile.spec.ts
willclarktech: I'm in favour of the helper option.

webmaster128: Check, implemented

packages/iov-keycontrol/src/userprofile.spec.ts
```diff
@@ -31,6 +31,27 @@ import { Ed25519HdWallet, Ed25519Wallet, Secp256k1HdWallet } from "./wallets";
 
 const { fromHex } = Encoding;
 
+/**
+ * A possibly incomplete equality checker that tests as good as possible if two profiles contain
```
willclarktech: "as well as possible"

PR #1030, opened 2019-05-28
Add Chain IDs spec

### LINKED ISSUE #1014: Write specification about chain IDs

webmaster128:
Closes #1014

### DIFF
--- a/docs/chain-ids.md
+++ b/docs/chain-ids.md
@@ -1,0 +1,69 @@
+# Chain IDs
+
+| Status | Author                    | Created    | Last updated | License   |
+| ------ | ------------------------- | ---------- | ------------ | --------- |
+| Draft  | The document contributors | 2019-05-23 | 2019-05-28   | CC-BY-4.0 |
+
+## Abstract
+
+In the IOV ecosystem, many tools can be used with different blockchains.
+Client-side, this is every application built on top of IOV-Core. Backend-side,
+this is the Blockchain Name System (BNS), a lookup service that allows assigning
+native blockchain addresses to a human readable name.
+
+All this tooling needs to be able to uniquely identify a blockchain, which is
+what chain IDs do. The chain ID is interpreted client-side only and all client
+software that wants to be compatible needs to agree on the same set of chain
+IDs. The BNS just stores chain IDs without interpreting them.
+
+## Format
+
+A chain ID is a case-sensitive string in the format `^[a-zA-Z0-9_.-]{4,128}$`.
+When choosing a new chain ID, the following guildelines should be followed:
+
+1. Choose a chain ID that is long enough to be unique within the blockchain
+   ecosystem.
+2. Choose a chain ID as short as possible to avoid unnecessary data.
+3. Choose a chain ID that is to some degree human readable and helps for basic
+   debugging.
+
+Some bad chain IDs are `ethereum` (violates 1.),
+`da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba` (violates 2.
+and 3.). Better alternatives are `ethereum-eip155-3` and `lisk-da3ed6a454`.
+
+### Reserved
+
+Chain IDs starting with `local-` are reserved for local chains. Please note the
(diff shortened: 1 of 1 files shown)


docs/chain-ids.md
```diff
@@ -0,0 +1,69 @@
...
+
+| Status | Author                    | Created    | Last updated | License   |
+| ------ | ------------------------- | ---------- | ------------ | --------- |
+| Draft  | The document contributors | 2019-05-23 | 2019-05-28   | CC-BY-4.0 |
+
+## Abstract
+
+In the IOV ecosystem, many tools can be used with different blockchains.
+Client-side, this is every application built on top of IOV-Core. Backend-side,
+this is the Blockchain Name System (BNS), a lookup service that allows assigning
```
willclarktech: "which enables the assignment of native blockchain addresses to human readable names"

docs/chain-ids.md
```diff
@@ -0,0 +1,69 @@
...
+trailing `-`, i.e. `localichain` could be registered at some point.
+
+## Registry
+
+The following chain IDs are registered. Please add new chain IDs via a pull
+request.
+
+| Chain ID             | Description                                                                                                                             | Example                                                                 |
+| -------------------- | --------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
+| `bns-%s`             | A BNS chain by IOV with the netwok name in `%s`                                                                                         | `bns-hugnet` for the Hugnet testnet                                     |
```
willclarktech: "network"

docs/chain-ids.md
```diff
@@ -0,0 +1,69 @@
...
+A chain ID is a case-sensitive string in the format `^[a-zA-Z0-9_.-]{4,128}$`.
+When choosing a new chain ID, the following guildelines should be followed:
+
+1. Choose a chain ID that is long enough to be unique within the blockchain
+   ecosystem.
+2. Choose a chain ID as short as possible to avoid unnecessary data.
+3. Choose a chain ID that is to some degree human readable and helps for basic
+   debugging.
+
+Some bad chain IDs are `ethereum` (violates 1.),
```
ethanfrey: :+1:

docs/chain-ids.md
```diff
@@ -0,0 +1,69 @@
...
+`da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba` (violates 2.
+and 3.). Better alternatives are `ethereum-eip155-3` and `lisk-da3ed6a454`.
+
+### Reserved
+
+Chain IDs starting with `local-` are reserved for local chains. Please note the
+trailing `-`, i.e. `localichain` could be registered at some point.
+
+## Registry
```
ethanfrey: :+1:

webmaster128:
> This will require a bit of update to iov-core code, right?

Right

> No changes to weave, right?

The only thing we could adjust is the size of the chain ID. Given the current spec, 128 chars/bytes are clearly not needed and we can reduce to reduce the risk of storage abuse.

> Or should we make changes to the genesis configuration of future test/dev nets to match some pattern?

I was planning to set local devnet chain IDs explicitly instead (similar to https://github.com/iov-one/ponferrada/commit/e35199dbab276d7c5b01e07d722f0e9cb69e788a but with the "local-"prefix). This has the great advantage that in many tests we can replace additional observer connections that do nothing else than getting the random chain ID with constants.

ethanfrey:
> I was planning to set local devnet chain IDs explicitly instead (similar to iov-one/ponferrada@e35199d but with the "local-"prefix). This has the great advantage that in many tests we can replace additional observer connections that do nothing else than getting the random chain ID with constants.

That one looks great. No more auto-generated chain id nonsense spilling into tests

webmaster128:
@ethanfrey do you have any opinion on

> The only thing we could adjust is the size of the chain ID. Given the current spec, 128 chars/bytes are clearly not needed and we can reduce to reduce the risk of storage abuse.

? I'd like to use a length value in the spec that we are happy with for a long time. Implementations can follow later.

PR #1031, opened 2019-05-28
Atomic swap spec improvements

webmaster128:


docs/atomic-swap-protocol-v1.md
```diff
@@ -196,6 +196,16 @@ locked tokens until the timeout is reached.
    the actor in short position to claim is 0 or negative (i.e. impossible to
    claim).
 
+## Contributors
+
+Substantial contributions by the following people in the form of text, review
+and ideas made this document possible:
+
+- Simon Warta
```
willclarktech: Links to GitHub profiles could be nice

webmaster128: Good idea, added in the last commit

PR #1037, opened 2019-05-28
Eslint: no-explicit-any

willclarktech:
Partially addresses #876


willclarktech:
One option would be to enable it on a per-file basis in any situation where we might be dealing with unknown data (e.g. parsers etc)?

webmaster128:
This makes all
- type guards
- error callbacks
- parsers

"bad" in the sense of linter judgement and pollutes great valid code with exceptions. I'd very much prefer to not introduce a rule that does not add a good amount of value.

willclarktech:
Makes sense.

webmaster128:
Thanks

PR #1038, opened 2019-05-28
Eslint: no-inferrable-types

willclarktech:
Partially addresses #876


willclarktech:
@webmaster128 Updated to include that commit only

PR #1042, opened 2019-05-28
Eslint: no-use-before-define

willclarktech:
Partially addresses #876

Functions are hoisted, so I left them as-is (we even have circular definitions in places which require breaking the rule). Classes and `const`s aren't, so I fixed those.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -26,7 +26,7 @@
     "@typescript-eslint/no-empty-interface": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }],
-    "@typescript-eslint/no-use-before-define": "off"
+    "@typescript-eslint/no-use-before-define": ["error", { "functions": false }]
   },
   "overrides": [
     {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,16 +17,16 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
-    "curly": ["error", "all"],
+    "curly": ["warn", "all"],
     "no-console": ["warn", { "allow": ["error", "info", "warn"] }],
-    "no-shadow": "error",
-    "prefer-const": "error",
+    "no-shadow": "warn",
+    "prefer-const": "warn",
     "spaced-comment": ["warn", "always", { "line": { "markers": ["/ <reference"] } }],
-    "@typescript-eslint/explicit-function-return-type": ["error", { "allowExpressions": true }],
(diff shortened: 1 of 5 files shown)


.eslintrc.json
```diff
@@ -29,7 +29,7 @@
     "@typescript-eslint/no-non-null-assertion": "off",
     "@typescript-eslint/no-object-literal-type-assertion": "off",
     "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }],
-    "@typescript-eslint/no-use-before-define": "off"
+    "@typescript-eslint/no-use-before-define": ["error", { "functions": false }]
```
webmaster128: Can we default all the rules ro warn instead of error? During active development it is good to be able to run non-linting code and I find it very annoying if the IDE marks this stuff in red.

willclarktech: Is it just a styling issue? It feels wrong to change lint settings because of the appearance in a particular IDE. How about using these settings to make the red less distracting?
```
"workbench.colorCustomizations": {
  "editorError.foreground": "#22ff22", // squiggly line
  "editorError.border": "#ff2222" // additional border under squiggly line
}
```

webmaster128: The issue is not being able to distinguish between compile errors and lint rules breaks, since eslint errors are marked the same way as typescript compiler errors. So I'd say it is more than styling but related.

The tslint plugin has a very nice setting to avoid this:

> tslint.alwaysShowRuleFailuresAsWarnings - Always show rule failures as warnings, ignoring the severity configuration in the tslint.json configuration.

See also
- https://github.com/eslint/eslint/issues/10438
- https://stackoverflow.com/questions/50705508/is-it-possible-to-show-warnings-instead-of-errors-on-all-of-eslint-rules

The `workbench.colorCustomizations` seems to affect both compiler errors and errors reported by the linter.

willclarktech: This seems like an interesting option: https://github.com/eslint/eslint/issues/10438#issuecomment-422465222

We could set a default severity level conditional upon the environment?

webmaster128: > This seems like an interesting option: eslint/eslint#10438 (comment)

This is only a slight deduplication of "warn" -> `defaultSeverity` and you still need to set it for every rule. If we used a JavaScript config, this would be fine. In JSON, I'd just use "warn" for all the rules that make the IDE red as long as there is no native alternative to https://github.com/bfanger/eslint-plugin-only-warn.

willclarktech: One problem here is that we'd need to override all the rules inherited from other plugins/configs to be consistent.

webmaster128: Indeed, that is a problem. Maybe we just change them to warn at the point they annoy us.

Issue #853, opened 2019-03-26
Cleanup Ethereum RPC client


webmaster128:
Right now, the EthereumConnection uses two different connections

* HTTP for all basic RPCs,
* Websocket for subscriptions.

It would be nice to have a generic EthereumRpcClient that uses a different implementation depending on the base URL (very similar to [the Tendermint-RPC Client](https://github.com/iov-one/iov-core/blob/v0.13.0/packages/iov-tendermint-rpc/src/client.ts#L11-L12)). This EthereumRpcClient does not have subscription functionality when used with a http implementation.

This can be split into two steps:

1. Update implementation respecting `baseUrl` and `EthereumConnectionOptions.wsUrl` in an intelligent way
2. Remove field `EthereumConnectionOptions.wsUrl` (API breaking)

willclarktech:
@webmaster128 Step 2 should go into v0.15?

webmaster128:
@willclarktech yes, since breaking.

Step 1:

We have up to 2 sources for node URLs, `baseUrl` and `EthereumConnectionOptions.wsUrl` and should find a clever algorithm that always uses the better one. I.e. as soon as there is a WS URL, we ignore the HTTP URL.

Step 2:

Cleanup by removing `EthereumConnectionOptions.wsUrl`. The connection type is now only determined by the URL schema in `baseUrl`.

willclarktech:
@webmaster128 Is WS definitely always preferable to HTTP? Eg could someone provide both because they want to use HTTP where possible but they still want to use the streaming methods that rely on WS?

webmaster128:
@willclarktech I was never facing a case where a WS connection was possible but HTTP was preferred. In some cases the stateless HTTP is better/easier, but as soon as a WS URL is provided and WS is used, it is safe to assume everything should go via WS.

willclarktech:
Closed in #1050

PR #1045, opened 2019-05-29
Allow WS baseUrl in EthereumConnection

willclarktech:
Opening now to get feedback on the direction before I tidy up/complete tests etc.

Step 1 of #853


packages/iov-ethereum/src/wsjsonrpcclient.ts
```diff
@@ -0,0 +1,30 @@
...
+  public constructor(socket: StreamingSocket) {
+    this.socket = socket;
+  }
+
+  public async run(request: JsonRpcRequest): Promise<JsonRpcResponse> {
+    await this.socket.connected;
+
+    const promise = new Promise<JsonRpcResponse>((resolve, reject) => {
+      this.socket.events
+        .map(event => parseJsonRpcResponse2(JSON.parse(event.data)))
```
webmaster128: Parsing exceptions in here should be ignored. I suggest converting exception to `null` in the map body and then use a null filter like this one: https://github.com/iov-one/ponferrada/blob/95ec319/packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts#L12-L14

packages/iov-ethereum/src/wsjsonrpcclient.ts
```diff
@@ -0,0 +1,30 @@
...
+        .filter(response => response.id === request.id)
+        .take(1)
+        .addListener({
+          next: resolve,
+          error: reject,
+        });
```
webmaster128: Nice streaming flow

packages/iov-ethereum/src/wsjsonrpcclient.ts
```diff
@@ -0,0 +1,30 @@
...
+  private readonly socket: StreamingSocket;
+
+  public constructor(socket: StreamingSocket) {
+    this.socket = socket;
+  }
+
+  public async run(request: JsonRpcRequest): Promise<JsonRpcResponse> {
+    await this.socket.connected;
+
+    const promise = new Promise<JsonRpcResponse>((resolve, reject) => {
```
webmaster128: `const pendingResponse =`?

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -119,7 +120,7 @@ export class EthereumConnection implements AtomicSwapConnection {
   }
 
   private readonly pollIntervalMs: number;
-  private readonly rpcClient: HttpJsonRpcClient;
+  private readonly rpcClient: HttpJsonRpcClient | WsJsonRpcClient;
```
webmaster128: Can we have a shared interface for those two?

willclarktech: Definitely, this is the main thing I want to refactor in the current approach.

willclarktech:
I've added an extra test for the ws client in many tests for `EthereumConnection`, but I left them out in some of the more involved cases. I'm not entirely happy about this pattern; a better approach is probably to add unit tests for the various JSON RCP clients and then it shouldn't matter which one we use in the `EthereumConnection` tests. That might be beyond the scope of this ticket however.

packages/iov-ethereum/src/wsjsonrpcclient.ts
willclarktech: ❤️ xstream

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -119,18 +120,24 @@ export class EthereumConnection implements AtomicSwapConnection {
...
   private readonly erc20Tokens: ReadonlyMap<TokenTicker, Erc20Options>;
   private readonly erc20ContractReaders: ReadonlyMap<TokenTicker, Erc20Reader>;
   private readonly codec: EthereumCodec;
 
   public constructor(baseUrl: string, chainId: ChainId, options: EthereumConnectionOptions) {
+    const baseUrlIsHttp = ["http://", "https://"].some(prefix => baseUrl.startsWith(prefix));
+    const baseUrlIsWs = ["ws://", "wss://"].some(prefix => baseUrl.startsWith(prefix));
+    const wsUrl = baseUrlIsWs ? baseUrl : options.wsUrl;
+    if (!wsUrl && !baseUrlIsHttp) {
+      throw new Error("Unsupported protocol for baseUrl: must be one of http, https, ws or wss");
```
webmaster128: I think this error messge should be based on `baseUrlIsWs`/`baseUrlIsHttp` only, so `wsUrl` can be set below this block.

packages/iov-ethereum/src/jsonrpcclient.ts
```diff
@@ -0,0 +1,11 @@
+import { Stream } from "xstream";
+
+import { JsonRpcRequest, JsonRpcResponse } from "@iov/jsonrpc";
+import { SocketWrapperMessageEvent } from "@iov/socket";
+
+export interface JsonRpcClient {
```
webmaster128: Can we call this `EthereumRpcClient`? We have JsonRpcClient already in `@iov/jsonrpc`

webmaster128:
The CI is most likely because of the `async function loadChainId(baseUrl: string): Promise<ChainId> {` implementation, which uses a HTTP client on a WS URL

willclarktech:
Hm, weird that it passes in Safari though?

webmaster128:
> Hm, weird that it passes in Safari though?

axios, which is used in the HTTP client, is only a wrapper around native APIs. Those APIs can behave differently. However, I did not look into the details.

PR #1047, opened 2019-05-29
Upgrade to weave 0.15

### LINKED ISSUE #1027: Upgrade to weave 0.15

webmaster128:
Closes #1027 

- [x] Codec and code
- [x] Local blockchains
- [x] Use new aswap module for atomic swaps
- [x] `Participant.power` -> `.weight`
- [x] Docs, CHANGELOG


packages/iov-bns/scripts/collect_testdata.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
ethanfrey: Awesome script!

Would have saved me hours in the past

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -413,22 +412,22 @@ export class BnsConnection implements AtomicSwapConnection {
   public async getSwapsFromState(query: AtomicSwapQuery): Promise<readonly AtomicSwap[]> {
     const doQuery = async (): Promise<QueryResponse> => {
       if (isAtomicSwapIdQuery(query)) {
-        return this.query("/escrows", query.id.data);
+        return this.query("/aswaps", query.id.data);
```
ethanfrey: Yup, proper indexes

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -648,24 +647,23 @@ export class BnsConnection implements AtomicSwapConnection {
...
-    if (res.results.length !== 1) {
-      throw new Error("Received unexpected number of fees");
+    const { results } = await this.query("/", Encoding.toAscii("_c:cash"));
+    if (results.length !== 1) {
+      throw new Error(`Unexpected number of results for minimal fee. Expected: 1 Got: ${results.length}`);
     }
-    const data = Encoding.fromAscii(res.results[0].value);
-    const amount = decodeJsonAmount(data);
-    return amount;
+    const { minimalFee } = decodeCashConfiguration(codecImpl.cash.Configuration.decode(results[0].value));
```
ethanfrey: Yes. This is correct.
We aim to provide proper protobuf objects for all configuration items in a predictable location.
I hoe this works for you, at least better than the previous one, and previously parsed and normalized from the genesis file input/

packages/iov-bns/src/decode.spec.ts
```diff
@@ -182,96 +174,17 @@ describe("Decode", () => {
     });
   });
 
-  describe("decodeJsonAmount", () => {
```
ethanfrey: Nice to kill this code.

packages/iov-bns/src/decode.ts
```diff
@@ -104,37 +100,11 @@ export function decodeAmount(coin: codecImpl.coin.ICoin): Amount {
...
-      fractional: fracString ? Int53.fromString(rightPadZeros(fracString.slice(1), 9)).toNumber() : undefined,
-      ticker: ticker as TokenTicker,
-    };
-    return decodeAmount(coin);
-  } else if (typeof data === "object" && data !== null) {
-    // parse a json coin representation
-    return decodeAmount(data as codecImpl.coin.ICoin);
-  }
-  throw new Error("Impossible type for amount json");
+export function decodeCashConfiguration(config: codecImpl.cash.Configuration): CashConfiguration {
```
ethanfrey: Glad to see decodeJsonAmount gone... this is simpler, no?

scripts/bnsd/genesis_app_state.json
```diff
@@ -31,14 +31,85 @@
...
-  "gconf": {
-    "cash:minimal_fee": "0.01 CASH",
-    "cash:collector_address": "0000000000000000000000000000000000000000"
+  "conf": {
+    "cash": {
+      "minimal_fee": "0.01 CASH",
+      "collector_address": "0000000000000000000000000000000000000000"
+    },
+    "migration": {
+      "admin": "cond:multisig/usage/0000000000000001"
```
ethanfrey: Future note, in 0.16 you can simplify this with
`seq:multisig/usage/1`
where `seq:` is like `cond:`, except it takes a normal int and encodes it as 8-byte bigendian, rather than you manually writing out the hex definition... which gets uglier when there are many.

scripts/bnsd/genesis_app_state.json
```diff
@@ -31,14 +31,85 @@
...
+    }
   },
   "msgfee": [
     {
       "msg_path": "nft/username/issue",
       "fee": "5 CASH"
     }
+  ],
+  "initialize_schema": [
```
ethanfrey: :+1:

packages/iov-bns/src/bnsconnection.ts
webmaster128: Yes the new protobuf config is way better than what we had before

packages/iov-bns/src/decode.ts
webmaster128: yes simpler, and most of all: consistent. Everything I get from the chain is provable (thus binary) encoded as protobuf.

Issue #1008, opened 2019-05-22
Use LightTransaction


webmaster128:
With https://github.com/iov-one/iov-core/pull/975 merged, we should do the following `BlockchainConnection` changes:

- [x] Create
`getTransaction: (id: TransactionId) => Promise<ConfirmedTransaction | FailedTransaction>;`.
This can later be used when clicking on transaction details from a transactions list. With the result of that, it should be possible to verify the signature locally.
- [x] Change interface to
`searchTx(query: TransactionQuery): Promise<ReadonlyArray<LightTransaction | FailedTransaction>>`
- [x] Change interface to
`listenTx(query: TransactionQuery): Stream<LightTransaction | FailedTransaction>`
- [x] Change interface to
`liveTx(query: TransactionQuery): Stream<LightTransaction | FailedTransaction>`

willclarktech:
@webmaster128 Definitely `LightTransaction` in these cases rather than `ConfirmedTransaction<LightTransaction>`? What should be done with all the (now superfluous) information?

webmaster128:
> Definitely LightTransaction in these cases rather than ConfirmedTransaction<LightTransaction>? 

Nope, my fault. `ConfirmedTransaction<LightTransaction>` is most likely the way to go. Feel free to do a plausibility check and error correction on the description above. It was more a sketch of the direction than a careful API design.

willclarktech:
Does the scope of this issue also cover changing `TxCodec.parseBytes`? I.e. should it return a signed `LightTransaction` without a `creator`? (And therefore not require the `chainId` parameter.)

webmaster128:
> Does the scope of this issue also cover changing TxCodec.parseBytes? I.e. should it return a signed LightTransaction without a creator? (And therefore not require the chainId parameter.)

Looking at the description

```
  /** parseBytes will recover bytes from the blockchain into a format we can use */
  readonly parseBytes: (bytes: PostableBytes, chainId: ChainId) => SignedTransaction;
```

I tend to say that `parseBytes` should be unchanged and be used as-is in the new `getTransaction`. However, if we need something similar for `LightTransaction`, it should be created at some proper place.

willclarktech:
Does this also apply to depencies of these methods such as `watchSwaps`? (Relies on `liveTx` in at lease one case.)

webmaster128:
> watchSwaps

[...]

PR #1051, opened 2019-06-03
Use LightTransaction in BlockchainConnection methods

willclarktech:
Part 1 of #1008

Definitely fighting the type system rather than working with it here to some degree. Opening this draft PR to get feedback on the approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@
   functions.
 - @iov/bcp: Add `LightTransaction` and `WithCreator` types, and
   `isLightTransaction` helper function.
+- @iov/bcp: `ConfirmedTransaction` and `SignedTransaction` now take a type
+  argument that extends `LightTransaction` (`UnsignedTransaction` is still the
+  default).
+- @iov/bcp: Helper methods `isConfirmedTransaction` and `isFailedTransaction`
+  now take a type argument that extends`LightTransaction`(`UnsignedTransaction`
+  is still the default).
 - @iov/bns: Add support for multisignature transactions.
 - @iov/bns: Add `CreateMultisignatureTx` and `UpdateMultisignatureTx` types
   along with `Participant`.
@@ -40,10 +46,15 @@ Breaking changes
   `UnsignedTransaction` (i.e. they have no `creator` field): `SendTransaction`,
   `SwapOfferTransaction`, `SwapClaimTransaction`, and `SwapAbortTransaction`.
 - @iov/bcp: Add `sender` field to `SendTransaction`.
+- @iov/bcp: `BlockchainConnection` methods `listenTx`, `liveTx` and `searchTx`
+  now all resolve to confirmed `LightTransaction`s instead of
+  `UnsignedTransaction`s.
 - @iov/bns: Remove obsolete types `BnsBlockchainNft`,
   `BnsBlockchainsByChainIdQuery`, `BnsBlockchainsQuery`.
 - @iov/bns: Switch to new `SwapId` type instead of `SwapIdBytes` where
   appropriate.
+- @iov/bns: `BnsConnection` methods `listenTx`, `liveTx` and `searchTx` now all
+  resolve to confirmed `LightTransaction`s instead of `UnsignedTransaction`s.
 - @iov/core: Remove `JsRpcSigningServer` and all related JS-RPC types.
 - @iov/core: Change return type of `SigningServerCore.signAndPost` to
   `TransactionId | null`.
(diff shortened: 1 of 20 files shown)


webmaster128:
This looks good as far as I can see.

> Definitely fighting the type system rather than working with it here to some degree. Opening this draft PR to get feedback on the approach.

hmm, can you say what exactly is causing the fighting? The only potential improvement I see in the diff is getting rid of the default generic type of `ConfirmedTransaction` and friends to make clear, this always wraps some kind of transaction that needs to ne specified.

willclarktech:
The main issue is here: https://github.com/iov-one/iov-core/pull/1051/files#diff-3bca4f284dd667ca2096454a4f73d657R79

With some blockchains we use a scraper/logs to get light transaction data out of the blockchain, but in other cases (eg BNS) we use a different approach and get the full transaction out (including creator). I didn't find a way to express that with types which didn't massively overcomplicate things. So the current approach has this workaround for BNS where we know that everything's ok, but the type system doesn't.

webmaster128:
For arrays, we can use TSs type inference mechanism and use

```
public async searchTx(
    query: TransactionQuery,
  ): Promise<readonly (ConfirmedTransaction<UnsignedTransaction> | FailedTransaction)[]> {
```

to implement

```
readonly searchTx: (query: TransactionQuery) => Promise<readonly (ConfirmedTransaction<LightTransaction> | FailedTransaction)[]>;
```

i.e. the BnsConnection provides strictly more than the interface asks for in a compatible manner. However, this approach does not work for Streams.

In this case I'd pull out a private `searchTxImpl` and friends that provide the full functionality and can be used internaly.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@
   functions.
 - @iov/bcp: Add `LightTransaction` and `WithCreator` types, and
   `isLightTransaction` helper function.
+- @iov/bcp: `ConfirmedTransaction` and `SignedTransaction` now take a type
+  argument that extends `LightTransaction` (`UnsignedTransaction` is still the
+  default).
+- @iov/bcp: Helper methods `isConfirmedTransaction` and
+  `isFailedTransaction now take a type argument that extends`LightTransaction`(`UnsignedTransaction`
```
webmaster128: Missing ` here

packages/iov-bcp/src/connection.ts
```diff
@@ -148,14 +148,14 @@ export interface FailedTransaction {
   readonly message?: string;
 }
 
-export function isConfirmedTransaction(
-  transaction: ConfirmedTransaction | FailedTransaction,
-): transaction is ConfirmedTransaction {
+export function isConfirmedTransaction<T extends LightTransaction = UnsignedTransaction>(
```
webmaster128: If the default value needed? Isn't this inferred by from the argument?

packages/iov-bcp/src/connection.ts
```diff
@@ -148,14 +148,14 @@ export interface FailedTransaction {
...
+  transaction: ConfirmedTransaction<T> | FailedTransaction,
+): transaction is ConfirmedTransaction<T> {
   return typeof (transaction as any).transaction !== "undefined";
 }
 
-export function isFailedTransaction(
-  transaction: ConfirmedTransaction | FailedTransaction,
+export function isFailedTransaction<T extends LightTransaction = UnsignedTransaction>(
```
webmaster128: If the default value needed? Isn't this inferred by from the argument?

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -566,7 +573,7 @@ export class BnsConnection implements AtomicSwapConnection {
    * It returns a stream starting the array of all existing transactions
    * and then continuing with live feeds
    */
-  public liveTx(query: TransactionQuery): Stream<ConfirmedTransaction | FailedTransaction> {
+  public liveTx(query: TransactionQuery): Stream<ConfirmedTransaction<LightTransaction> | FailedTransaction> {
```
webmaster128: Now you can deduplictae the implementation by using `this.liveTxUnsigned`; same for `listenTx` and `searchTx`

packages/iov-ethereum/src/ethereumconnection.ts
```diff
@@ -712,7 +713,9 @@ export class EthereumConnection implements AtomicSwapConnection {
     }
   }
 
-  public listenTx(query: TransactionQuery): Stream<ConfirmedTransaction | FailedTransaction> {
+  public listenTx(
+    query: TransactionQuery,
+  ): Stream<ConfirmedTransaction<LightTransaction> | FailedTransaction> {
```
webmaster128: Great! When you now change the return type of `searchSendTransactionsByAddressOnScraper`, we get rid of this extra network query for each transaction: `const transaction = (await this.searchTransactionsById(Parse.transactionId(tx.hash)))[0];`

packages/iov-bns/src/bnsconnection.ts
willclarktech: We still have this problem with streams though, don't we?

webmaster128: We cannot use the result 1:1, but a simple map simipar to

```ts
this.liveTxUnsigned.map((original): ConfirmedTransaction<LightTransaction> | FailedTransaction) => {
   return original;
});```

should do the job

PR #1059, opened 2019-06-04
Implement BlockchainConnection.getTx

### LINKED ISSUE #1008: Use LightTransaction

willclarktech:
Closes #1008

Probably the tests verifying signatures are unnecessary but I added them as a sanity check. Happy to remove them if they’re just clutter.

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -35,11 +35,12 @@ import {
   SwapProcessState,
   SwapTimeout,
   TokenTicker,
+  TransactionId,
   TransactionState,
   UnsignedTransaction,
   WithCreator,
 } from "@iov/bcp";
-import { Random } from "@iov/crypto";
+import { Ed25519, Random, Sha512 } from "@iov/crypto";
 import { Encoding, Uint64 } from "@iov/encoding";
 import { Ed25519HdWallet, HdPaths, UserProfile, WalletId } from "@iov/keycontrol";
 import { asArray, firstEvent, lastValue, toListPromise } from "@iov/stream";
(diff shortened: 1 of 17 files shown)


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -938,6 +939,101 @@ describe("BnsConnection", () => {
...
+      {
+        const nonExistentId = "abcd" as TransactionId;
+        await connection
+          .getTx(nonExistentId)
+          .then(fail.bind(null, "should not resolve"), error =>
+            expect(error).toMatch(/transaction does not exist/i),
+          );
+      }
+
+      // by existing ID (from lisk/init.sh)
```
webmaster128: lisk/init.sh in BNS tests?

packages/iov-ethereum/src/ethereumconnection.spec.ts
```diff
@@ -691,17 +691,122 @@ describe("EthereumConnection", () => {
...
+      {
+        const nonExistentId = "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef" as TransactionId;
+        await connection
+          .getTx(nonExistentId)
+          .then(fail.bind(null, "should not resolve"), error =>
+            expect(error).toMatch(/transaction does not exist/i),
+          );
+      }
+
+      // by existing ID (from lisk/init.sh)
```
webmaster128: lisk/init.sh?

webmaster128:
> What about moving them in the implementation of getTx and document getTx to be signature verifying? Seems to be a strong feature.

hmm, maybe a bit out of scope and when the signature is already verified, there is no real reason to return it.

packages/iov-bns/src/bnsconnection.spec.ts
willclarktech: Whoops, leftover from copy/pasting the test structure.

willclarktech:
Nice idea, but I think it makes more sense to have separate functions for retrieval and verification. Do we have an issue already to provide a nice wrapper for this kind of verification?

webmaster128:
> Do we have an issue already to provide a nice wrapper for this kind of verification?

Nope, signature verification has never been on the TODO

willclarktech:
Then I guess we wait until after MVP?

Issue #229, opened 2018-08-06
Implement retries into WS and HTTP RPC


Isabello:
Right now, when we connect to WS(s) or HTTP(s) RPC, it only tries once. We should allow for retries before throwing completely.

Additionally, being able to retry the connection if we lose it would be helpful as well.

- [x] Implement a queueing wrapper around `StreamingSocket` with a non-blocking send function
- [x] Implement reconnects that send messages from the queue via the new connection as soon as established
- [x] Expose connected state (connecting, connected, disconnected) as a `ValueAndUpdates` field (Because we can. It is a fair API, but not strictly required for the product)
- [x] Make it possible for higher level applications to react to reconnects via a simple `reconnected` callback
- [x] Automatically reconnect on error/completion
- [x] Automatically re-subscribe in tendermint-rpc
- [x] Automatically re-subscribe in ethereum

ethanfrey:
This happens in the web-wallet when there is connectivity break and afaik, it cannot recover without reloading the app. This "reload" will be a bit harder with an extension.

Reconnecting on network failure is important, so the next call may succeed.

However, we must also consider subscriptions and how to handle that when it break?
Do we re-subscribe? What about items lost in the meantime? There needs to be some higher-level logic at work.

eg. I use `liveTx` to search and subscribe for all tx that match my account. I lose connectivity and reconnect 2 minutes later, missing a tx event. If I auto-resubscribe I will never get that one (but will get future ones). To handle it properly, we must notifiy so the higher level code (eg liveTx can do a search since last receive and then subscribe again).

Properly handling subscriptions can be done later. This should allow:
* minimally a reconnect, so a query will work later
* optionally reconnect the subscriptions as well (even if there is a gap they will miss)

willclarktech:
@webmaster128 What's the scope of this issue? `tendermint-rpc` only or others as well?

webmaster128:
> What's the scope of this issue?

Preferribly lower level to cover `tendermint-rpc` as well as `ethereum`. I guess an auto-queueing and auto-reconnecting socket wrapper in @iov/sockets makes sense.

ethanfrey:
Yes, it would be good to auto-reconnect for long-term stability.

The one issue is that this information of a reconnect must be passed to the higher-level handler.
If a tendermint connection opens a number of subscriptions for account updates, then the lower-level websocket reconnects, we will also want to re-create those subscriptions.

[...]

PR #1060, opened 2019-06-05
Add QueueingStreamingSocket wrapper around StreamingSocket

willclarktech:
Part 1 of #229

### DIFF
--- a/packages/iov-socket/src/queueingstreamingsocket.ts
+++ b/packages/iov-socket/src/queueingstreamingsocket.ts
@@ -1,32 +1,73 @@
-// tslint:disable:readonly-array readonly-keyword
+// tslint:disable:readonly-array readonly-keyword no-object-mutation
+import { Listener, Stream } from "xstream";
+
+import { SocketWrapperMessageEvent } from "./socketwrapper";
 import { StreamingSocket } from "./streamingsocket";
 
 /**
- * A StreamingSocket that can queue requests.
+ * A wrapper around StreamingSocket that can queue requests.
  */
-export class QueueingStreamingSocket extends StreamingSocket {
-  private queue: string[] = [];
+export class QueueingStreamingSocket {
+  public readonly connected: Promise<void>;
+  public readonly events: Stream<SocketWrapperMessageEvent>;
+
+  private errorProducerListener: Listener<Error> | undefined;
+  private readonly socket: StreamingSocket;
+  private readonly queue: string[] = [];
+  private isProcessingQueue = false;
+  private timeoutIndex = 0;
+  private processQueueTimeout: NodeJS.Timeout | null = null;
+
+  public constructor(url: string, timeout: number = 10_000) {
+    this.socket = new StreamingSocket(url, timeout);
+    this.connected = this.socket.connected;
+    this.events = this.socket.events;
+  }
+
+  public connect(): void {
+    this.socket.connect();
+  }
+
+  public disconnect(): void {
+    this.socket.disconnect();
+  }
 
   public getQueueLength(): number {
     return this.queue.length;
   }
 
-  public async queueRequest(data: string): Promise<void> {
-    this.queue.push(data);
-    return this.processQueue();
+  public queueRequest(request: string): void {
+    this.queue.push(request);
+    // We don’t need to wait for the queue to be processed.
+    // tslint:disable-next-line:no-floating-promises
+    this.processQueue();
   }
 
-  public async processQueue(): Promise<void> {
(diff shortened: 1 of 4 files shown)


packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  public async queueRequest(data: string): Promise<void> {
+    this.queue.push(data);
+    return this.processQueue();
+  }
+
+  public async processQueue(): Promise<void> {
+    let data: string | undefined = this.queue.shift();
+    while (data) {
```
webmaster128: With `while ((data = this.queue.shift()) !== undefined)`, you can deduplicate the shifting

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -0,0 +1,32 @@
...
+      try {
+        await this.send(data);
+      } catch (error) {
+        // Assume that the connection is down; stop processing queue for now.
+        this.queue.unshift(data);
```
webmaster128: nice one

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -0,0 +1,32 @@
...
+ * A StreamingSocket that can queue requests.
+ */
+export class QueueingStreamingSocket extends StreamingSocket {
+  private queue: string[] = [];
+
+  public getQueueLength(): number {
+    return this.queue.length;
+  }
+
+  public async queueRequest(data: string): Promise<void> {
```
webmaster128: why should `queueRequest` bother about the processing? What is processes here can very well be unrelated to the enqueued request. I think this should just trigger `this.processQueue` without waiting for the result.

We might need other ways of error reporting later on, but right now I don't see how processQueue could throw/reject

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -0,0 +1,32 @@
+// tslint:disable:readonly-array readonly-keyword
+import { StreamingSocket } from "./streamingsocket";
+
+/**
+ * A StreamingSocket that can queue requests.
+ */
+export class QueueingStreamingSocket extends StreamingSocket {
```
webmaster128: I like the extension design in this case

webmaster128: Hmm, on the other hand: this extension allows you to bypass the queue, which is something the API should disallow. So I tend to prefer composition over extension here.

packages/iov-socket/src/queueingstreamingsocket.ts
willclarktech: True, but we have a lint rule against that because it's confusing to assign things in conditional statements: https://eslint.org/docs/rules/no-cond-assign

packages/iov-socket/src/queueingstreamingsocket.ts
willclarktech: I also had this thought, I'm just scared of leaving promises unhandled in something changes later on and we don't update this. Another option would be to leave out the processing altogether and rely on the caller to do both? (Then the caller can leave the promise unhandled if they really want to.)

packages/iov-socket/src/queueingstreamingsocket.ts
webmaster128: I really don't like when tools prevent us from writing smart code. I checked the docs again and it turned out that eslint by default allows the version I suggested and just prevents = vs. == typos. The linter rule complaining here was tslint. See https://github.com/iov-one/iov-core/pull/1064

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -0,0 +1,71 @@
...
+      try {
+        await this.socket.send(request);
+        this.timeoutIndex = 0;
+        this.isProcessingQueue = false;
+      } catch (error) {
+        // Probably the connection is down; try again in a bit.
+        this.queue.unshift(request);
+        this.isProcessingQueue = false;
+        // Starts with a 1 second timeout; doubles every time
+        this.processQueueTimeout = setTimeout(() => this.processQueue(), 2 ** this.timeoutIndex++ * 1000);
```
webmaster128: Nice logic. I think 1000 ms for the first retry is very long (but probably no problem). But this also means when my computer was offline for ~1 hour, the sleep will be ~1 hour. So this should have an upper limit

PR #1061, opened 2019-06-05
Add reconnect method to QueueingStreamingSocket

willclarktech:
~WIP: Creating draft PR now to check direction.~
Part 2 of #229
~Depends on #1060~

~Main things I’m uncertain about:~
~1. Relationships between classes (also affects #1060)~
~2. Changes of accessibility/readonly status~

Note unconventional use of `.imitate` to switch the `events` stream to the new socket. From [the docs](https://github.com/staltz/xstream#-imitatetarget):

> This method exists to allow one thing: **circular dependency of streams**

which is definitely not what it's being used for here.

### DIFF
--- a/packages/iov-socket/src/queueingstreamingsocket.spec.ts
+++ b/packages/iov-socket/src/queueingstreamingsocket.spec.ts
@@ -74,7 +74,6 @@ describe("QueueingStreamingSocket", () => {
 
       requests.forEach(request => socket.queueRequest(request));
 
-      socket.reconnect();
       socket.events.subscribe({
         next: event => {
           expect(event.data).toEqual(requests[eventsSeen++]);
@@ -85,6 +84,7 @@ describe("QueueingStreamingSocket", () => {
           }
         },
       });
+      socket.reconnect();
     });
   });
 });
--- a/packages/iov-socket/src/queueingstreamingsocket.spec.ts
+++ b/packages/iov-socket/src/queueingstreamingsocket.spec.ts
@@ -18,7 +18,7 @@ describe("QueueingStreamingSocket", () => {
     it("can queue and process requests with a connection", done => {
       pendingWithoutSocketServer();
       const socket = new QueueingStreamingSocket(socketServerUrl);
-      const requests: readonly string[] = ["request 1", "request 2", "request 3"];
+      const requests = ["request 1", "request 2", "request 3"] as const;
       let eventsSeen = 0;
       socket.events.subscribe({
         next: event => {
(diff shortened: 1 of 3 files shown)


packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -8,20 +8,23 @@ import { StreamingSocket } from "./streamingsocket";
  * A wrapper around StreamingSocket that can queue requests.
  */
 export class QueueingStreamingSocket {
-  public readonly connected: Promise<void>;
+  public connected: Promise<void>;
```
webmaster128: I don't think this API is very useful, as it only tells you that _some_ socket connected, which says very little about the connection status. I think we can just drop it.

In the one place I found where it is used, we probably don't need it:

```ts
      socket.connect();
      // tslint:disable-next-line:no-floating-promises
      socket.connected.then(() => {
        requests.forEach(request => socket.queueRequest(request));
      });
});
```

-> 

```ts
      socket.connect();
      requests.forEach(request => socket.queueRequest(request));
});
```

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -32,6 +35,13 @@ export class QueueingStreamingSocket {
     this.socket.disconnect();
   }
 
+  public reconnect(): void {
+    this.socket = new StreamingSocket(this.url, this.timeout);
+    this.events.imitate(this.socket.events);
```
webmaster128: Nice find with `imitate`.

However, I guess we need some moderation here. For example when there is a connection failure, we get stream error events that should not be passed to the caller. So I suggest creating a stream event producer for `this.events` and use changing source streams to produce events.

packages/iov-socket/src/queueingstreamingsocket.spec.ts
```diff
@@ -56,8 +56,35 @@ describe("QueueingStreamingSocket", () => {
...
         socket.connect();
       }, 5_000);
     });
   });
+
+  describe("reconnect", () => {
+    it("can reconnect and process remaining queue", done => {
+      pendingWithoutSocketServer();
+      const socket = new QueueingStreamingSocket(socketServerUrl);
+      const requests: readonly string[] = ["request 1", "request 2", "request 3"];
```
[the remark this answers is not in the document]
willclarktech: Thanks, didn't know about this. Syntax looks weird to me but definitely useful.

packages/iov-socket/src/queueingstreamingsocket.ts
willclarktech: Should we just filter out all the errors? Are there any errors we do want to propagate? If not, we should be able to use the following (with a corresponding line in the constructor):
```ts
this.events.imitate(this.socket.events.replaceError(Stream.never));
```

webmaster128: I'm not entirely sure how this will look at the end of the day. A custom producer os most flexible, but a bit more code. `this.events.imitate(this.socket.events.replaceError(Stream.never));` is probably sufficient for now

webmaster128: What I am also concerned about is completion of the source stream

willclarktech: Good point, we can do a similar move there though.

PR #1062, opened 2019-06-05
WIP: Add nefarious test for JSON TransactionEncoder

willclarktech:


packages/iov-core/src/transactionencoder.spec.ts
```diff
@@ -163,5 +163,12 @@ describe("TransactionEncoder", () => {
       expect(restored).toEqual(original);
       expect(isIdentity(restored.creator)).toEqual(true);
     });
+
+    fit("throws an error if too deeply nested", () => {
+      const a: any = {};
+      // tslint:disable-next-line no-object-mutation
+      a.a = a;
+      expect(() => fromJson(a)).toThrowError(/maximum call stack size exceeded/i);
+    });
```
webmaster128: What this does is trying to decode a circular structure. We have a test like this for the JS->JSON direction: https://github.com/iov-one/iov-core/blob/master/packages/iov-core/src/transactionencoder.spec.ts#L70-L80

In the JSON->JS direction, this can never happen as there exists no circularity in JSON data. So the better question is: can we guarantee that the input of this function is really JSON?

willclarktech: That's definitely a good question to ask, but I was using this as a shortcut to creating something very deeply nested. The same problem ought to apply. (Although to be clear I don't think this is much of a problem since the user would have to have bad configuration for it to arise.)

PR #1069, opened 2019-06-06
Use names "pubkey"/"privkey"

### LINKED ISSUE #1054: Rename PrivateKey* -> Privkey*
Some months ago we decided that privkey and pubkey are atomic words in our space. Those are the remaining old terms.

- [ ] PrivateKeyBundle
- [ ] PrivateKeyBytes
- [ ] PublicKeyBundle
- [ ] PublicKeyBytes

webmaster128:
Closes #1054

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@ Breaking changes
 - @iov/bcp: `BlockchainConnection` methods `listenTx`, `liveTx` and `searchTx`
   now all resolve to confirmed `LightTransaction`s instead of
   `UnsignedTransaction`s.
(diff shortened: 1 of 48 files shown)


webmaster128:
"pubkey" and "privkey" give us shorter but still very meaningful symbol names and less word boundaries when composing the terms. But at the end of the day this is some tradeoff between best possible language and compactness.

We use "pubkey" and "privkey" almost in the entire codebase and I think the alternative ways to ensure consistency would be to rename all ~1400 "pubkey"s to "publicKey"s.

In the BNS codec, we don't have consistency right now. However, there are more "pubkey"s than "publickKey"s. To rename the rest as well, we just need to create tickets in weave.

PR #1070, opened 2019-06-11
Upgrade weave codec to  v0.16

### LINKED ISSUE #1053: Upgrade to weave 0.16
not released yet

webmaster128:
Closes #1053


webmaster128:
> Is it normal to put a generated .d.ts file in src and then end up with a copy of it in types and commit both?

Interesting question, I just checked again: In contrast to the rest of our code, there is no .ts file for the codec. So if we want to keep the build flow from `./src/` into `./build/types/` and `./types/`, we need to keep this duplication. Right now, I don't see a great potential for improvement.

PR #1071, opened 2019-06-11
Cleanup types.ts in iov-bns

### LINKED ISSUE #1055: Remove remaining decoder function from iov-bns/src/types.ts
Stuff like `decodeFullSig` should go to decode.ts

webmaster128:
~Based on #1069~
Closes #1055


webmaster128:
> Nice, does the CHANGELOG need updating?

Thanks. I don't think so, since this is all package internal and does not change externally visible bevaviour.

PR #1073, opened 2019-06-11
Add BNS governance functionality (part 1)

webmaster128:


packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,89 @@ export interface CashConfiguration {
...
+}
+
+export enum ProposalStatus {
+  Submitted,
+  Closed,
+  Withdrawn,
+}
+
+/** Union type for possible options */
+export type ProposalOption = string;
```
willclarktech: Union type?

packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,89 @@ export interface CashConfiguration {
...
+
+export interface Proposal {
+  readonly title: string;
+  /** The transaction to be executed when the proposal is accepted */
```
willclarktech: How should this be encoded?

packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,89 @@ export interface CashConfiguration {
...
+  readonly option: ProposalOption;
+  readonly description: string;
+  readonly electionRule: VersionedId;
+  readonly electorate: VersionedId;
```
willclarktech: Do we need the version here? Are we going to have forking electorates on a single chain?

packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,89 @@ export interface CashConfiguration {
...
+  /** Time when the voting on this proposal starts (Unix timestamp) */
+  readonly votingStartTime: number;
+  /** Time when the voting on this proposal starts (Unix timestamp) */
+  readonly votingEndTime: number;
+  /** Time of the block where the proposal was added to the chain (Unix timestamp) */
+  readonly submissionTime: number;
```
willclarktech: Do we need to use this type to e.g. preview proposals before submission? If so we might want to make this optional.

packages/iov-bns/src/types.ts
```diff
@@ -188,6 +271,25 @@ export function isUpdateEscrowPartiesTx(tx: LightTransaction): tx is UpdateEscro
...
+  readonly title: string;
+  /** The transaction to be executed when the proposal is accepted */
+  readonly option: ProposalOption;
+  readonly description: string;
+  readonly electionRuleId: Uint8Array;
+  /** Unix timestamp when the proposal starts */
+  readonly startTime: number;
+  /** The author of the proposal must be included in the list of transaction signers. */
+  readonly author: Address;
+}
```
willclarktech: The electorate is inferred from the election rule? Is the ID of an election rule guaranteed to be unique?

packages/iov-bns/src/decode.ts
```diff
@@ -192,6 +215,132 @@ export function decodeParticipants(
...
+      return ProposalStatus.Closed;
+    case codecImpl.gov.Proposal.Status.PROPOSAL_STATUS_WITHDRAWN:
+      return ProposalStatus.Withdrawn;
+    default:
+      throw new Error("Received unknown value for proposal status");
+  }
+}
+
+function decodeRawProposalOption(rawOption: Uint8Array): ProposalOption {
+  const po = codecImpl.app.ProposalOptions.decode(rawOption);
```
willclarktech: Prefer verbose names

webmaster128: Right. Before I extracted this into a separate method, all other obvious names were taken. But now we can adjust easily.

packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,89 @@ export interface CashConfiguration {
...
+  readonly weight: number;
+}
+
+export interface Electorate {
+  readonly id: number;
+  readonly version: number;
+  readonly admin: Address;
+  readonly title: string;
+  /** An unordered map from address to elector */
+  readonly electors: { readonly [index: string]: Elector };
```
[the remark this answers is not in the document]
webmaster128: Right. But since `Electorate` data only flows from blockchain -> client (with no way to create a new `Electorate` from the client), we only get valid data in here (via decodeElectorate).

Expressing `electors` as a list as it is done in the weave codec indicates that it has a fixed order and addresses can occur multiple times, so this duplication is probably the smaller issue.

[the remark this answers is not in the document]
willclarktech: One option would be to make this a map from address to weight.

packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,89 @@ export interface CashConfiguration {
...
+export interface Fraction {
+  readonly numerator: number;
+  readonly denominator: number;
+}
+
+export interface ElectionRule {
+  readonly id: Uint8Array;
+  readonly version: number;
+  readonly admin: Address;
+  /** The electorate used by this (ID without version, as changing electorate changes the rule) */
```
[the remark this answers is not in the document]
webmaster128: Haha, yes. I copied this from the weave docs and don't understand it myself anymore.

packages/iov-bns/src/types.ts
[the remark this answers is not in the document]
webmaster128: Yes, since we don't need to mirror the `Elector` type from weave (which I did), we can just use

```ts
interface ElectorProperties {
   readonly weight: number;
}
type Electors = { readonly [index: string]: ElectorProperties }
```

packages/iov-bns/src/types.ts
webmaster128: `string | never` ;) This is just the beginning of a type that will later on be able to express all the different actions that can be executed when a proposal is accepted.

packages/iov-bns/src/types.ts
webmaster128: Assume you have an electorate "tech committee". People can leave and join this commitee in a slow and trusted fashion. Whenever this committee changes, the `version` from the `(id, version)` pair is bumped, but the `id` is not. This allows you to reference the latest version of the tech comittee when creating a proposal. As soon as the proposal is created on chain (i.e. the create transaction is executed), the version is fixed. This implicitely fixes all members and their weights.

packages/iov-bns/src/types.ts
webmaster128: The blockchain is the only place where `Proposal` is created. An instruction to create a proposal is the `CreateProposalTx`, which should be the only source of data for any kind of preview.

packages/iov-bns/src/types.ts
webmaster128: 1. Yes
2. Yes, the `electionRuleId` must be pulled from the blockchain and is guaranteed to be unique there

packages/iov-bns/src/types.ts
willclarktech: Right, so as soon as the committee changes we need new versions of all the rules we want to continue, right? And is there going to be some mechanism on the backend preventing proposals referencing older versions of an electorate?

webmaster128: 1. A proposal is created without the need to know the latest version (which is considered a feature). In [CreateProposalMsg]( https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#gov.CreateProposalMsg), we only set the `id` of `(id, version)` for both ElectionRule and implicitly Electorate.
2. Due to 1., at proposal creation time the blockchain picks the latest version. As soon as a poposal is created, the version is fixed.

packages/iov-bns/src/types.ts
```diff
@@ -25,19 +25,23 @@ export interface CashConfiguration {
 
 // Governance
 
-export interface Elector {
-  readonly address: Address;
+/** Like Elector from the backend but without the addess field */
```
willclarktech: "address"

packages/iov-bns/src/types.ts
```diff
@@ -23,6 +23,104 @@ export interface CashConfiguration {
...
+
+export interface ElectionRule {
+  readonly id: Uint8Array;
+  readonly version: number;
+  readonly admin: Address;
+  /**
+   * The eligible voters in this rule.
+   *
+   * This is an unversioned ID (see `id` field in weave's VersionedIDRef), meaning the
+   * electorate can change over time without changing this ID. When a poposal with this
```
willclarktech: "proposal"

PR #1076, opened 2019-06-11
Add QueueingStreamingSocket.connectionStatus stream

willclarktech:
Based on #1061
Part 3 of #229

### DIFF
--- a/packages/iov-socket/src/queueingstreamingsocket.spec.ts
+++ b/packages/iov-socket/src/queueingstreamingsocket.spec.ts
@@ -1,4 +1,4 @@
-import { QueueingStreamingSocket } from "./queueingstreamingsocket";
+import { ConnectionStatus, QueueingStreamingSocket } from "./queueingstreamingsocket";
 
 function pendingWithoutSocketServer(): void {
   if (!process.env.SOCKETSERVER_ENABLED) {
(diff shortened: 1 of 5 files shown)


packages/iov-socket/src/queueingstreamingsocket.spec.ts
```diff
@@ -87,4 +89,51 @@ describe("QueueingStreamingSocket", () => {
...
+            case 2:
+            case 6:
+              expect(status).toEqual(WebSocket.CLOSING);
+              break;
+            case 3:
+            case 7:
+              expect(status).toEqual(WebSocket.CLOSED);
+              break;
+            default:
+              fail("Got too many status changes");
```
webmaster128: when there is a `done` callback, you need to use `done.fail`. The simple `fail` is unreliable in this case.

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -18,27 +21,36 @@ export class QueueingStreamingSocket {
...
 
   public connect(): void {
+    this.connectionStatusProducer.update(WebSocket.CONNECTING);
     this.socket.connect();
   }
 
   public disconnect(): void {
+    this.connectionStatusProducer.update(WebSocket.CLOSING);
     this.socket.disconnect();
+    this.connectionStatusProducer.update(WebSocket.CLOSED);
```
webmaster128: At this point, the websocket is in the CLOSING state since WebSocket.close() does not block.

I suggest using our own, higher level connection status codes, which behave slightly different than a reaw WebSocket. Something like:

- `ConnectionStatus.Unconnected` (default at construct time)
- `ConnectionStatus.Connecting`
- `ConnectionStatus.Connected`
- `ConnectionStatus.Disconnected` (as soon as disconnect() was called, no matter what the implementation does)

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -3,17 +3,27 @@ import { Stream } from "xstream";
...
+import { DefaultValueProducer, ValueAndUpdates } from "@iov/stream";
+
 import { SocketWrapperMessageEvent } from "./socketwrapper";
 import { StreamingSocket } from "./streamingsocket";
 
 function wrapStream<T>(stream: Stream<T>): Stream<T> {
   return concat(stream.replaceError(Stream.never), Stream.never());
 }
 
+export enum ConnectionStatus {
```
webmaster128: this new type (and QueueingStreamingSocket as well) should be exported in index.ts

packages/iov-socket/src/queueingstreamingsocket.ts
```diff
@@ -3,17 +3,27 @@ import { Stream } from "xstream";
...
+  Connecting,
+  Connected,
+  Disconnected,
+}
+
 /**
  * A wrapper around StreamingSocket that can queue requests.
  */
 export class QueueingStreamingSocket {
+  public connectionStatus: ValueAndUpdates<ConnectionStatus>;
```
webmaster128: `public readonly`

PR #1080, opened 2019-06-12
Add stop script for socketserver

### LINKED ISSUE #1077: Add stop script for socketserver
Currently we have a start script but have to kill the processes manually.

willclarktech:
Closes #1077

### DIFF
--- a/scripts/socketserver/stop.sh
+++ b/scripts/socketserver/stop.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+command -v shellcheck > /dev/null && shellcheck "$0"
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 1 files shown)


scripts/socketserver/stop.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+command -v shellcheck > /dev/null && shellcheck "$0"
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+pkill -f "$SCRIPT_DIR"/echo.py --port 444
```
webmaster128: Can we put the search pattern into a single string, to make clear what belongs where?

```
pkill -f "$SCRIPT_DIR/echo.py --port 444"
```

should be the same. 444 is used as a common prefix for both processes, right?

willclarktech: Yep, we use 4444 and 4445.

PR #1084, opened 2019-06-13
Add ReconnectingSocket

willclarktech:
Part 5 of #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   using an optional constructor argument.
 - @iov/socket: Add 10 second timeout to `StreamingSocket`. Value can be
   configured using an optional constructor argument.
-- @iov/socket: Add `QueueingStreamingSocket` class.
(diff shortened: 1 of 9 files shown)


willclarktech:
Did some more investigation and found that there some edge cases where the queue won't work as intended. Specifically, if the base Websocket instance's `send` method is called just when a disconnection occurs, it might fail silently, with the result that a request never makes it to the server, but leaves the queue permanently. I've reduced the scope of this for non-browser environments by adjusting `SocketWrapper` (although it still seems possible - see `TODO` comment in the `ReconnectingSocket` test). In browsers I didn't find a way to reduce the scope though, as this method doesn't take a callback.

Having made this change, the tests for `SigningService` are now failing in Chrome in CI, and fail for both Chrome and Safari locally for me. Right now I can't see why they should be affected (tests pass fine on `master`), but I might return to this issue later in the week in case some inspiration hits.

PR #1090, opened 2019-06-19
Implement BNS escrow

### LINKED ISSUE #1058: Add implementation for BNS escrow module

willclarktech:
Closes #1058

I'm not 100% sure I've got the logic right here (especially with `buildUpdateEscrowPartiesTx` and amounts), but did what made most sense given the types and error messages from the node during testing.

A few naming issues surfaced:
- `CreateEscrowTx` has an `amount` field which looks singular but takes an array of amounts
- The `ITx` type from the codec has an `updateEscrowMsg` but elsewhere the same concept is referred to as `UpdateEscrowParties` rather than just `UpdateEscrow`. I opted for the more verbose option where relevant.
- I wasn't sure what the relationship between `src` when creating an escrow and `sender` when updating was (or what `src` is if not the same thing as `sender`). The documentation says "If sender is not defined, it defaults to the first signer" for `CreateEscrowMsg`, but there is not field for `sender`.

### DIFF
--- a/packages/iov-bns/src/decode.ts
+++ b/packages/iov-bns/src/decode.ts
@@ -32,13 +32,17 @@ import {
   BnsUsernameNft,
   CashConfiguration,
   ChainAddressPair,
+  CreateEscrowTx,
   CreateMultisignatureTx,
   Keyed,
   Participant,
   PrivkeyBundle,
   PrivkeyBytes,
   RegisterUsernameTx,
+  ReleaseEscrowTx,
   RemoveAddressFromUsernameTx,
+  ReturnEscrowTx,
+  UpdateEscrowPartiesTx,
   UpdateMultisignatureTx,
 } from "./types";
 import { addressPrefix, encodeBnsAddress, identityToAddress } from "./util";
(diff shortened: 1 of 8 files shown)


packages/iov-bns/src/decode.ts
```diff
@@ -188,6 +192,13 @@ export function decodeParticipants(
   }));
 }
 
+export function decodeTimeout(timeout: number | Long): number {
```
webmaster128: There is `asIntegerNumber` already, see usage for timeout in parseSwapOfferTx

packages/iov-bns/src/types.ts
```diff
@@ -155,7 +201,11 @@ export type BnsTx =
   | RemoveAddressFromUsernameTx
   // BNS: Multisignature contracts
   | CreateMultisignatureTx
-  | UpdateMultisignatureTx;
+  | UpdateMultisignatureTx
```
webmaster128: Group separator comment missing

packages/iov-bns/src/types.ts
```diff
@@ -140,6 +141,51 @@ export function isUpdateMultisignatureTx(tx: LightTransaction): tx is UpdateMult
   return tx.kind === "bns/update_multisignature_contract";
 }
 
+export interface CreateEscrowTx extends LightTransaction {
```
webmaster128: Could you add a group separator comment above (`// Transactions: Escrows`)

packages/iov-bns/src/decode.ts
```diff
@@ -221,7 +225,7 @@ function parseSwapOfferTx(
     kind: "bcp/swap_offer",
     hash: hash as Hash,
     recipient: encodeBnsAddress(prefix, ensure(msg.recipient, "recipient")),
-    timeout: { timestamp: asIntegerNumber(ensure(msg.timeout)) },
+    timeout: { timestamp: asIntegerNumber(ensure(msg.timeout, "timeout")) },
```
webmaster128: 👍 thanks

PR #1091, opened 2019-06-25
Prepare Governance

webmaster128:
Small BNS refactorings that should be included independent of governance


webmaster128:
> Why is the integration test being deleted?

It is not. Just the types of the spec file are removed (like everywhere else)

PR #1096, opened 2019-06-27
Add voting and tallying

### LINKED ISSUE #960: Expose BNS governance functionality
Depends on https://github.com/iov-one/iov-core/issues/1053

webmaster128:
Closes #960

------

- [x] CHANGELOG entry
- [x] Create `interface TextProposal { text: string }` in `ProposalOption` to better match the backend structure


webmaster128:
If I am not missing something, this closes #960

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1220,6 +1222,132 @@ describe("BnsConnection", () => {
...
+        const nonce = await connection.getNonce({ pubkey: author.pubkey });
+        const signed = await profile.signTransaction(voteForProposal, bnsCodec, nonce);
+        const response = await connection.postTx(bnsCodec.bytesToPost(signed));
+        await response.blockInfo.waitFor(info => !isBlockInfoPending(info));
+      }
+
+      await sleep(5_000);
+
+      {
+        // Election voting period ended but not yet tallyed
```
willclarktech: "tallied"

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1220,6 +1222,132 @@ describe("BnsConnection", () => {
...
+        if (!blockInfo.result) {
+          throw new Error("Transaction result missing");
+        }
+        proposalId = toHex(blockInfo.result).toUpperCase();
+      }
+
+      {
+        // Election submitted, voting period not yet started
+        const proposal = (await connection.getProposals()).find(p => p.id === proposalId)!;
+        expect(proposal.votingStartTime).toBeGreaterThan(Date.now() / 1000);
```
willclarktech: The timing conditions are less about how the code works and more about how the test setup works. If the transaction fails we throw an error rather than failing an expectation. Should we also do that for these timing-related requirements?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1220,6 +1222,132 @@ describe("BnsConnection", () => {
...
+        expect(proposal.state.totalYes).toEqual(1);
+        expect(proposal.state.totalNo).toEqual(0);
+        expect(proposal.state.totalAbstain).toEqual(0);
+        expect(proposal.status).toEqual(ProposalStatus.Closed);
+        expect(proposal.result).toEqual(ProposalResult.Accepted);
+        expect(proposal.executorResult).toEqual(ProposalExecutorResult.Succeeded);
+      }
+
+      connection.disconnect();
+    }, 30_000);
```
willclarktech: Is this enough to pass CI reliably?

packages/iov-bns/src/decode.spec.ts
```diff
@@ -310,6 +321,12 @@ describe("Decode", () => {
         votingEndTime: 42424243,
         submissionTime: 3003,
         author: "tiov1qqgjyv6y24n80zyeqqgjyv6y24n80zyed9d6mt" as Address,
+        state: {
```
willclarktech: Why use a different key?

packages/iov-bns/src/types.ts
```diff
@@ -93,10 +93,25 @@ export enum ProposalStatus {
...
+  Yes,
+  No,
+  Abstain,
+}
+
+export interface TallyResult {
+  readonly totalYes: number;
+  readonly totalNo: number;
+  readonly totalAbstain: number;
+  readonly totalElectorateWeight: number;
```
willclarktech: Is this just a sum of the other numbers? Do we need it at all?

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: I had no issues with that so far. Locally the test runs in about 12 seconds if I remember correctly.

packages/iov-bns/src/decode.spec.ts
webmaster128: Different than `.voteState`? Because I was overwhelmed by all those terms like election, vote, voting, tally, result that seem to be mixed without any real concept. How can a "state" have a type called "result", which is not a result until the voting period is over? Why it it `votingEndTime` but `voteState`? Then the noun "vote" has like [8 distict meanings](https://www.dict.cc/?s=vote), so term adds basically nothing in this context.

I'm open for better ideas but don't want to blindly copy what is coming from weave.

willclarktech: Fair enough. It would be nice to have an explicit domain model actually.

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: A proper [testing framework](https://stackoverflow.com/a/2565309/2013738) comes with [built-in expectations and assetions](https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#natural-expressions), where assertions stop test execution and expectations don't. Unfortunately, everything in the JS world does not have that, which causes a bit of strangenes in code like:

```js
expect(encoded.sendMsg).toBeDefined(); // test should fail and end here
expect(encoded.sendMsg!.memo).toEqual("free transaction"); // but it only ends when accessing `undefined.memo`, creating a reporting that is misleading
```

What we have is not great but does the job:

* `expect` is the most readable option in most cases, especially before we allowed `if`s without curly brackets
* `throw` stops execution, which is nice to have. Most importantly, it is used by the TypeScript type analyzer to infer correct types in the non-error case.

So the main reason for using exception for transaction fails is that otherwise `blockInfo.result` would not be available below. My rule of thumb so far is: use `expect` by default for readability and `throw` when TS needs it.

willclarktech: How would you feel about using [Node assert](https://nodejs.org/api/assert.html#assert_assert) for these assertions which should stop execution if they fail?

packages/iov-bns/src/types.ts
webmaster128: No, it is the total weight of eligable voters, so

```
voterParticipation = (totalYes+totalNo+totalAbstain) / totalElectorateWeight
```

The number is indeed redundant (backend stores it twice or even 3 times on purpose), but with data that is indirect (pseudo-code):

```
getElectorate(Proposal.electorateVersionedID).totalWeight
```

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: Would be good to have something like that. Unfortunately, Node assert does not play well with TypeScript, as TS does not get the information that for some cases an exception is thrown. So

```
assert(encoded.sendMsg);
expect(encoded.sendMsg.memo).toEqual("free transaction"); // encoded.sendMsg is still cash.ISendMsg | null | undefined at compile time
```

As far as I know there is no way to express _function will throw type X if runtime condition y fails_ using TypeScript declarations. If there was, we could build a small assertion library on our own.

webmaster128:
+3 commits, starting with "Fix typo: tallyed -> tallied"

packages/iov-bns/src/types.ts
```diff
@@ -93,10 +93,33 @@ export enum ProposalStatus {
...
+}
+
+export interface CreateTextResolution {
```
willclarktech: I find the word `Create` a bit confusing here.

webmaster128: I think the naming issue is in "option". This field holds an action to be performed when the proposal is accepted. All of those action [got verb-prefixed recently](https://github.com/iov-one/weave/issues/738). My personal intuition was to use "publish text resolution", but then I used the [new name from weave 0.17](https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.17.0/docs/proto/index.html#gov.CreateTextResolutionMsg).

Thoughts?

packages/iov-bns/src/types.ts
```diff
@@ -93,10 +93,33 @@ export enum ProposalStatus {
...
+  readonly resolution: string;
+}
+
+export function isCreateTextResolution(option: ProposalOption): option is CreateTextResolution {
+  return typeof option.resolution === "string";
+}
+
+/** Union type for possible options. Any option is ony action that will be executed when the proposal is accepted */
```
willclarktech: "A proposal option is the action to be executed when the proposal is accepted"?

packages/iov-bns/src/types.ts
willclarktech: Yeah, "action" would be much clearer. Could we make it `CreateTextResolutionAction`?

webmaster128: In https://github.com/iov-one/iov-core/pull/1096/commits/e7fffb539e95e800c67a1b57585dca05fd2dba4f I renamed the "option"s to "action"s. I think we don't need it as a suffic for every single type in `ProposalAction` then.

PR #1109, opened 2019-07-03
Add @iov/bns-governance package

### LINKED ISSUE #1102: Create @iov/bns-governance
A packages built on top of @iov/bns, that makes the most common governance tasks as easy as possible to use from the wallet project.

It should also be easy to use from the @iov/cli

willclarktech:
Closes #1102

Still to do:
- [x] Add tests
- [ ] Confirm API
- [ ] Confirm transaction creator structure (in particular `title` property)

For later issues:
~- Support other proposal types~ (https://github.com/iov-one/iov-core/issues/1112)
~- Specify committee constants~ (https://github.com/iov-one/iov-core/issues/1113)
~- Add an integration test for posting and getting proposals~ (https://github.com/iov-one/iov-core/issues/1114)


packages/iov-bns-governance/src/governor.ts
```diff
@@ -0,0 +1,111 @@
...
+  public constructor({ connection, identity }: GovernorOptions) {
+    this.connection = connection;
+    this.identity = identity;
+    this.address = bnsCodec.identityToAddress(this.identity);
+  }
+
+  public async getElectorates(): Promise<readonly Electorate[]> {
+    const electorates = await this.connection.getElectorates();
+    return electorates.filter(({ electors }) =>
+      Object.keys(electors).some(key => key.toUpperCase() === this.address.toUpperCase()),
```
webmaster128: BNS addresses have a deterministic casing and as soon as an `Address` instance is created, we can assume the casing is correct. User input needs to be normalized in the place where the input happens.

packages/iov-bns-governance/src/governor.ts
```diff
@@ -0,0 +1,111 @@
...
+  public async getProposals(): Promise<readonly Proposal[]> {
+    const electorates = await this.getElectorates();
+    const proposals = await this.connection.getProposals();
+    return proposals.filter(({ electorate }) => {
+      const electorateId = new BN(electorate.id).toNumber();
+      return electorates.some(({ id }) => id === electorateId);
+    });
+  }
+
+  public async createProposalTx(options: ProposalOptions): Promise<BnsTx> {
```
webmaster128: Should return `BnsTx & WithCreator`

packages/iov-bns/src/index.ts
```diff
@@ -22,6 +22,7 @@ export {
   UpdateMultisignatureTx,
   isUpdateMultisignatureTx,
   // Governance
+  CreateProposalTx,
```
webmaster128: Fixed on master

packages/iov-bns-governance/src/governor.ts
willclarktech: Perfect, wasn't sure about this.

packages/iov-bns-governance/src/governor.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    if (filteredRules.length === 0) {
+      throw new Error("No election rule found for electorate");
+    }
+
+    const mostRecentRule = filteredRules
+      .slice(1)
+      .reduce((rule1, rule2) => (rule1.version > rule2.version ? rule1 : rule2), filteredRules[0]);
+
+    return Encoding.toHex(mostRecentRule.id);
```
webmaster128: Let's use `return new BN(mostRecentRule.id).toNumber()` here instead to hide the lack of beauty of @iov/bns

See https://github.com/iov-one/iov-core/issues/1110

packages/iov-bns-governance/src/governor.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.identity = identity;
+    this.address = bnsCodec.identityToAddress(this.identity);
+  }
+
+  public async getElectorates(): Promise<readonly Electorate[]> {
+    const electorates = await this.connection.getElectorates();
+    return electorates.filter(({ electors }) => Object.keys(electors).some(key => key === this.address));
+  }
+
+  public async getElectionRuleId(electorateId: number): Promise<string> {
```
webmaster128: This assumes there is a 1:1 mapping between an electorate and an election rule. Is this a valid assumption? If yes, it should be documented in a small comment or something.

willclarktech: It doesn't assume a 1:1 mapping, it just assumes that for most purposes you're only interested in the most recent rule for a given electorate. Maybe it should instead find the rule with the version which matches the current electorate version?

webmaster128: But when there are rule `(id: 1, version: 1)`, `(id: 1, version: 2)`, `(id: 2, version: 1)`, only the latest version of the first rule is returned.

> Maybe it should instead find the rule with the version which matches the current electorate version?

I don't think the versions of the rule and the electorate are necessarily in sync

willclarktech: OK, clarified things with Alex. The electorates and election rules are established in the genesis block and currently there is no way to modify/add these after genesis. In principle one electorate can have more than one election rule, and when creating a proposal, the proposer is free to choose which of these to use. In practice though, we're likely to have only one election rule per electorate. I've updated the method to return an array of election rules so that the caller can select which one to use.

packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+
+// The first IOV key (m/44'/234'/0') generated from this mnemonic produces the address
+// tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea (bech32) / a4f97447e7df55b6ef0d6209ebef2a7b22625376 (hex).
+// This account has money in the genesis file (see scripts/bnsd/README.md).
+const faucetMnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
+const faucetPath = HdPaths.iov(0);
+const bnsdUrl = "ws://localhost:23456";
+
+async function getConnectionAndIdentity(
+  url = bnsdUrl,
```
webmaster128: do we need a configurable `url` argument here?

packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      const governor = new Governor(options);
+
+      const electorateId = 100;
+      return governor
+        .getElectionRules(electorateId)
+        .then(
+          () => fail("Expected promise to be rejected"),
+          err => expect(err.message).toMatch(/no election rule found/i),
+        )
+        .then(() => options.connection.disconnect());
```
webmaster128: This flow is a bit hard to follow. What about

```ts
      const electorateId = 100;
      await governor
        .getElectionRules(electorateId)
        .then(
          () => fail("Expected promise to be rejected"),
          err => expect(err.message).toMatch(/no election rule found/i),
        );
      options.connection.disconnect();
```

packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+          startTime: 1562161125,
+          electionRuleId: 5,
+          committee: 8 as CommitteeId,
+          address: "abcd" as Address,
+        })
+        .then(
+          () => fail("Expected promise to be rejected"),
+          err => expect(err.message).toMatch(/proposal type not yet supported/i),
+        )
+        .then(() => options.connection.disconnect());
```
webmaster128: Flow as above

packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  describe("createProposalTx", () => {
+    it("throws an error for unsupported proposal types", async () => {
+      const options = await getConnectionAndIdentity();
+      const governor = new Governor(options);
+
+      return governor
+        .createProposalTx({
+          type: ProposalType.AddCommitteeMember,
+          description: "Change something",
+          startTime: 1562161125,
```
webmaster128: What about using `ReadonlyDate` instead of timestamps in the bns-governance interface? Seems to feel a bit less lowlevel.

packages/iov-bns-governance/src/governor.ts
webmaster128: > In practice though, we're likely to have only one election rule per electorate.

Okay, we can make this assumption in the caller code, where it is easy to generalize

packages/iov-bns-governance/src/governor.spec.ts
willclarktech: No, this was premature optimisation.

packages/iov-bns-governance/src/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+      }
+      return grouped;
+    },
+    // tslint:disable-next-line:readonly-array readonly-keyword
+    [] as { readonly key: K; values: T[] }[],
+  );
+}
+
+export function maxWithComparatorCallback<T>(values: readonly T[], callback: (a: T, b: T) => T): T {
+  return values.slice(1).reduce(callback, values[0]);
```
webmaster128: Pretty close to what I has in mind. Ideas:
1. A typical comparator function (across many different ecosystems) returns a negative value, 0, a positive value (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description), allowing implementations like `(rule1, rule2) => rule1.version - rule2.version`
2. `maxWithComparatorCallback` can be implemented using array.sort and using the last element
3. There should be some kind of defined behavior for `values` being empts. A hard throw is fine I guess.

packages/iov-bns-governance/src/governor.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  public async getElectionRules(electorateId: number): Promise<readonly ElectionRule[]> {
+    const electionRules = await this.connection.getElectionRules();
+    const filteredRules = electionRules.filter(rule => rule.electorateId === electorateId);
+
+    if (filteredRules.length === 0) {
+      throw new Error("No election rule found for electorate");
+    }
+
+    const groupedRules = groupByCallback(filteredRules, rule => rule.id);
```
webmaster128: sweet

packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+  };
+}
+
+describe("Governor", () => {
+  // TODO: Keys for COMMITTEE_IDS to be finalised
+  it("exports COMMITTEE_IDS constants", async () => {
+    expect(typeof COMMITTEE_IDS.testnet.GOVERNING_BOARD).toBe("number");
+    expect(typeof COMMITTEE_IDS.testnet.ECONOMIC_COMMITTEE).toBe("number");
+    expect(typeof COMMITTEE_IDS.testnet.VALIDATOR_COMMITTEE).toBe("number");
+    expect(typeof COMMITTEE_IDS.testnet.TECH_COMMITTEE).toBe("number");
```
webmaster128: Can we use `toEqual` consistently when we want data equality and reserve `toBe` for object identity? I think it better expresses the intent and it is to easy to change an argument from string/number to an object, making the test wrong.

packages/iov-bns-governance/src/utils.ts
willclarktech: Re 2, I ended up with a solution which was pretty simple even without using sort. Probably we won't have enough proposals anytime soon to make the complexity difference significant, but just in case we'll be ready.

webmaster128: Sometimes it is about the beauty of code :)

packages/iov-bns-governance/types/index.d.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./constants";
+export * from "./governor";
+export * from "./proposals";
```
webmaster128: Can we make the exported types explicit? It is important that the caller projects that's have more types available than they should use.

packages/iov-bns-governance/src/constants.ts
```diff
@@ -0,0 +1,18 @@
+import { CommitteeId } from "./proposals";
+
+export interface CommitteeIds {
+  readonly GOVERNING_BOARD: CommitteeId;
```
webmaster128: Can we use `lowerCamelCase` for consistently all values and properties? So far, we could avoid using SCREAM_CASE, which is a good thing.

packages/iov-bns-governance/src/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+  readonly ECONOMIC_COMMITTEE: CommitteeId;
+  readonly VALIDATOR_COMMITTEE: CommitteeId;
+  readonly TECH_COMMITTEE: CommitteeId;
+}
+
+export const COMMITTEE_IDS: { readonly [chainId: string]: CommitteeIds } = {
```
webmaster128: no SCREAM_CASE please

packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+      const tx = await governor.createProposalTx({
+        type: ProposalType.AmendProtocol,
+        text: "Switch to Proof-of-Work",
+        title: "Custom title",
+        description: "Proposal to change consensus algorithm to POW",
+        startTime: new ReadonlyDate(1562164525898),
+        electionRuleId: 1,
+      });
+      expect((tx as any).title).toEqual("Custom title");
```
webmaster128: `tx` should be of type `CreateProposalTx & WithCreator`, which would avoid the cast

packages/iov-bns-governance/src/governor.ts
```diff
@@ -0,0 +1,100 @@
...
+  public async getProposals(): Promise<readonly Proposal[]> {
+    const electorates = await this.getElectorates();
+    const proposals = await this.connection.getProposals();
+    return proposals.filter(({ electorate }) => {
+      const electorateId = new BN(electorate.id).toNumber();
+      return electorates.some(({ id }) => id === electorateId);
+    });
+  }
+
+  public async createProposalTx(options: ProposalOptions): Promise<BnsTx & WithCreator> {
```
webmaster128: * Return type is always `CreateProposalTx & WithCreator`, just with different `action` properties
* Since this creates a `CreateProposalTx`, the name should be `createCreateProposalTx`. Alternatively, we can prefix all 3 methods with `make` or `build`

packages/iov-bns-governance/src/proposals.ts
```diff
@@ -0,0 +1,90 @@
...
+  AddValidator,
+  RemoveValidator,
+  AccessGuaranteeFund,
+  DistributeFunds,
+  AmendProtocol,
+}
+
+export interface CommonProposalOptions {
+  readonly type: ProposalType;
+  readonly title?: string;
```
webmaster128: Every proposal needs a title and a description (weave says "Human readable title. Must match `^[a-zA-Z0-9 _.-]{4,128}$`"), so this cannot be optional.

packages/iov-bns-governance/src/governor.ts
willclarktech: I like `build`.

webmaster128: [I like turtles](https://www.youtube.com/watch?v=CMNry4PE93Y).

packages/iov-bns-governance/src/proposals.ts
willclarktech: So currently a default is provided in the transaction builder. I was pretty unclear on the exact roles of `title` vs `description` vs `text`, so I thought we might want a fairly generic/uniform title by default (with the option to override) to avoid a mess?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## 0.16.0
 
+- @iov/bns-governance: Add new package to help with common governance tasks.
+  Includes `Governor` class (with `GovernorOptions`); `ProposalType` along with
+  the union type `ProposalOptions` and individual options types for each
+  proposal type; `CommitteeId` and `CommitteeIds` types along with a constant
+  `COMMITTEE_IDS` to be populated with known committee IDs on a per-chain basis.
```
webmaster128: COMMITTEE_IDS -> committeeIds

Issue #1112, opened 2019-07-04
Support all BNS governance proposal types


willclarktech:
As of https://github.com/iov-one/iov-core/pull/1109 we only support `AmendProtocol` but we should have structures and functions to be able to support all the others listed here: https://github.com/iov-one/iov-core/blob/721c761836904af0f3160fa4ba67b3f927e6efb2/packages/iov-bns-governance/src/proposals.ts#L9-L19

- [x] `AddCommitteeMember`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#gov.UpdateElectorateMsg
- [x] `RemoveCommitteeMember`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#gov.UpdateElectorateMsg
- [x] `AmendElectionRuleThreshold`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#gov.UpdateElectionRuleMsg
- [x] `AmendElectionRuleQuorum`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#gov.UpdateElectionRuleMsg (~~blocked by https://github.com/iov-one/weave/issues/869~~)
- [x] `AddValidator`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#validators.SetValidatorsMsg
- [x] `RemoveValidator`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.16.0/docs/proto/index.html#validators.SetValidatorsMsg
- [x] `ReleaseGuaranteeFunds`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/master/docs/proto/index.html#escrow.ReleaseMsg [CLARIFICATION REQUIRED]
- [x] `DistributeFunds`: https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/master/docs/proto/index.html#bnsd.ExecuteProposalBatchMsg and https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/master/docs/proto/index.html#cash.SendMsg [CLARIFICATION REQUIRED]
- [x] `AmendProtocol`

webmaster128:
[...]

PR #1117, opened 2019-07-04
Support AddCommitteeMember and RemoveCommittee member proposals

willclarktech:
Part 1 of #1112

~Might want an enum instead of a string literal to individuate action types?~ ✅


packages/iov-bns-governance/src/governor.ts
```diff
@@ -61,19 +61,46 @@ export class Governor {
   }
 
   public async buildCreateProposalTx(options: ProposalOptions): Promise<CreateProposalTx & WithCreator> {
+    const kind: "bns/create_proposal" = "bns/create_proposal";
+    const commonProperties = {
+      kind: kind,
```
webmaster128: This can also be inlined: `kind: "bns/create_proposal" as const,`

packages/iov-bns/src/encode.ts
```diff
@@ -14,6 +14,7 @@ import {
   SwapOfferTransaction,
   UnsignedTransaction,
   WithCreator,
+  Address,
```
webmaster128: linting error

packages/iov-bns-governance/src/governor.ts
willclarktech: Had to separate it out to avoid this really weird cast: `kind: "bns/create_proposal" as "bns/create_proposal";` but much nicer with `as const`.

packages/iov-bns/src/types.ts
```diff
@@ -99,6 +99,11 @@ export enum VoteOption {
   Abstain,
 }
 
+export enum ActionKind {
```
webmaster128: Please add explicit string raw values to thse cases, since all transaction data needs to go over JSON interfaces and source and destination may use different versions of IOV-Core, so this must be extensible in a way that the same raw value never translates to a different enum case.

PR #1123, opened 2019-07-08
Improve import tooling

webmaster128:
With this merged, imports are sorted automatically as part of the eslint autofix step. For me this saves a lot of time that is usually needed to manually sort imports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "format": "lerna run format",
     "format-text": "prettier --write --prose-wrap always --print-width 80 \"./*.md\" \"./docs/**/*.md\" \"./scripts/**/*.{json,md}\" && lerna run format-text",
     "lint": "lerna run lint",
+    "lint-fix": "lerna exec 'eslint \"**/*.{js,ts}\" --fix'",
     "test": "lerna run test",
     "build": "lerna run build",
     "docs": "lerna run docs",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "eslint": "^5.16.0",
     "eslint-config-prettier": "^4.3.0",
     "eslint-plugin-prettier": "^3.1.0",
+    "eslint-plugin-simple-import-sort": "^4.0.0",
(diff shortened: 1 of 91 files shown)


packages/iov-cli/src/tsrepl.ts
```diff
@@ -1,7 +1,7 @@
 import { diffLines } from "diff";
 import { join } from "path";
 import { Recoverable, REPLServer, start } from "repl";
-import { register, Register, TSError } from "ts-node";
+import { Register, register, TSError } from "ts-node";
```
willclarktech: 🤔

packages/iov-crypto/src/libsodium.ts
```diff
@@ -5,7 +5,7 @@
 
 // use require instead of import because of this bug
 // https://github.com/jedisct1/libsodium.js/issues/148
```
willclarktech: Probably time to remove this comment?

package.json
```diff
@@ -34,6 +35,7 @@
     "eslint": "^5.16.0",
     "eslint-config-prettier": "^4.3.0",
     "eslint-plugin-prettier": "^3.1.0",
+    "eslint-plugin-simple-import-sort": "^4.0.0",
```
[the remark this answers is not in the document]
webmaster128: There has never been a good reason for the @iov/* block other than some sense of beauty. I think we can live without that at no cost.

packages/iov-cli/src/tsrepl.ts
webmaster128: That's the JS default sort order:

```
> ["register", "Register"].sort()
[ 'Register', 'register' ]
```

No idea why tslint wanted something different before. Maybe tslint allowed both?

webmaster128:
> How come we can switch away from import x = require('x')? Was there a change in TS?

If I understand correctly, this is what the TS compiler setting `esModuleInterop: true` allows us to do

webmaster128:
Updated

PR #1125, opened 2019-07-09
Add support for AddValidator/RemoveValidator

webmaster128:
Belongs to https://github.com/iov-one/iov-core/issues/1112


packages/iov-bns/src/types.ts
```diff
@@ -25,6 +26,24 @@ export interface CashConfiguration {
 
 // Governance
 
+export interface ValidatorProperties {
+  readonly power: number;
+}
+
+export interface Validator extends ValidatorProperties {
+  readonly pubkey: PubkeyBundle;
+  readonly power: number;
```
willclarktech: Do you need to specify this here?

webmaster128: In `encodeValidators` we need to set `power` when creating a [ValidatorUpdate](https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.18.0/docs/proto/index.html#weave.ValidatorUpdate). I assumed not every validator gets the same power (whatever that means in Tendermint land), so I made it configurable.

willclarktech: I just mean because `Validator` already extends `ValidatorProperties`.

webmaster128: Ah great catch. That's a bug

PR #1132, opened 2019-07-16
Support AmendElectionRuleThreshold/Quorum proposals

willclarktech:
Part of #1112

### DIFF
--- a/packages/iov-bns-governance/src/governor.spec.ts
+++ b/packages/iov-bns-governance/src/governor.spec.ts
@@ -238,6 +238,96 @@ describe("Governor", () => {
       options.connection.disconnect();
     });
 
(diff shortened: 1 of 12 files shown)


webmaster128:
Could you run this through the linter once #1123 is merged? Looks like no change needed but just in case

webmaster128:
Wait a second – this will crash in `packages/iov-bns/src/encode.ts` at

```
  } else {
    throw new Error("Got unsupported type of ProposalOption");
  }
```

right? I'd prefer to merge this when the whole flow works. Otherwise we need an extra ticket to ensure we don't forget about that.

willclarktech:
Ah yes, sorry, missed that step.

willclarktech:
We need to wait for weave 0.19 for this.

PR #1133, opened 2019-07-16
Upgrade prettier to 1.18.2 and adapt code

webmaster128:


webmaster128:
indeed. I hoped you would not notice ;)

PR #1134, opened 2019-07-17
Support ReleaseGuaranteeFunds governance proposal

willclarktech:
Part of #1112

### DIFF
--- a/packages/iov-bns/src/decode.ts
+++ b/packages/iov-bns/src/decode.ts
@@ -55,10 +55,10 @@ import {
   UpdateEscrowPartiesTx,
   UpdateMultisignatureTx,
   UpdateTargetsOfUsernameTx,
+  Validators,
   VersionedId,
   VoteOption,
   VoteTx,
-  Validators,
 } from "./types";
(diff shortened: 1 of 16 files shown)


packages/iov-bns/src/decode.ts
```diff
@@ -328,13 +329,41 @@ function decodeProposalStatus(status: codecImpl.gov.Proposal.Status): ProposalSt
...
   const option = codecImpl.bnsd.ProposalOptions.decode(rawOption);
   if (option.govCreateTextResolutionMsg) {
     return {
       kind: ActionKind.CreateTextResolution,
       resolution: decodeString(option.govCreateTextResolutionMsg.resolution),
     };
+  } else if (option.escrowReleaseMsg) {
+    return {
+      kind: ActionKind.ReleaseGuaranteeFunds,
+      // TODO: codecImpl.bnsd.ProposalOptions.decode currently returns a Buffer
```
webmaster128: Right, this is because of the JS implementation of the protobuf encoder/decoder. The copy into Uint8Array is perfect

PR #1135, opened 2019-07-18
Support DistributeFunds proposal type

willclarktech:
Part of #1112

Some things I'm unclear about:
1. How the funds are actually authorised to be released from the account - is it a special case handled by the backend governance implementation?
1. Once the funds have been divided according to the weightings, what should happen to the remainder? The current approach leaves them in the account.
1. Whether the fee quote can be assumed to be correct when passing an empty array of messages for the batch proposal execution and filling in the messages later. Or in general how the fee works: is it one fee for the whole batch or one per message? Is there even a fee for this particular action?

### DIFF
--- a/packages/iov-bns/src/encode.ts
+++ b/packages/iov-bns/src/encode.ts
@@ -25,7 +25,9 @@ import {
   CreateProposalTx,
   isBnsTx,
   isCreateTextResolution,
+  isExecuteProposalBatch,
   isReleaseEscrow,
+  isSend,
   isSetValidators,
   isUpdateElectionRule,
   isUpdateElectorate,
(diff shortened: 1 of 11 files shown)


packages/iov-bns/src/encode.ts
```diff
@@ -318,7 +320,25 @@ function buildCreateProposalTx(tx: CreateProposalTx): codecImpl.bnsd.ITx {
...
+        messages: action.messages.map(message => {
+          if (!isSend(message)) {
+            throw new Error("Only send actions are currently supported in proposal batch");
+          }
+          return {
+            sendMsg: {
+              source: decodeBnsAddress(message.sender).data,
+              destination: decodeBnsAddress(message.recipient).data,
+              amount: encodeAmount(message.amount),
+              memo: message.memo || null,
```
webmaster128: We have `encodeString` for that

willclarktech:
@webmaster128 Made the `guaranteeFundEscrowId` and `rewardFundAddress` `Governor` options optional because some governors won't be concerned with them.

webmaster128:
Regarding 3.: now that the tally is executed by the cron job, the tally as well as the transaction executed if the proposal succeeds are free

PR #1139, opened 2019-07-23
Upgrade TypeScript to 3.5

webmaster128:


packages/iov-ethereum/src/ethereumcodec.spec.ts
```diff
@@ -48,7 +58,8 @@ describe("ethereumCodec", () => {
       ) as PubkeyBytes;
 
       const postableBytes = Encoding.toUtf8(JSON.stringify(rawGetTransactionByHashResult)) as PostableBytes;
-      expect(ethereumCodec.parseBytes(postableBytes, "ethereum-eip155-4" as ChainId)).toEqual({
+      const parsed = ethereumCodec.parseBytes(postableBytes, "ethereum-eip155-4" as ChainId);
+      expect((parsed as unknown) as SignedTransaction<SendTransaction & WithCreator>).toEqual({
```
willclarktech: What changed to make these additions necessary?

webmaster128: `parsed` is of type `SignedTransaction<UnsignedTransaction>`. Before, TS accepted additional properties such as `amount` in the equality check (via Jasmine) at compile time. Now TS complains when trying to check with additional properties.

The more egegant way would be to use a type checker like `isSendTransaction`. However, I don't see a way to use them for checking `SignedTransaction<SendTransaction>` without introducing a new checker for every transaction type.

webmaster128:
> We might want to think about using the new Omit type, but that can definitely be a later issue.

This is in fact the motivation for this upgrade. I want to test type checkers with missing properties

PR #1142, opened 2019-07-23
Improve BCP type checks

webmaster128:
Required to get https://github.com/iov-one/ponferrada/issues/393 done

I decided to fix this here in IOV-Core in order to make the use of this library more convenient for everybody.


packages/iov-bcp/src/transactions.spec.ts
```diff
@@ -1,6 +1,152 @@
...
+  isUint8Array,
+  PubkeyBundle,
+  PubkeyBytes,
+} from "./transactions";
+
+const { fromHex } = Encoding;
 
 describe("transactions", () => {
+  describe("isNonNullObject", () => {
+    it("returns true for objects", () => {
```
willclarktech: "including arrays" to help with clarity here?

packages/iov-bcp/src/transactions.spec.ts
```diff
@@ -1,6 +1,152 @@
...
+    });
+
+    it("returns false for a bunch of other stuff", () => {
+      expect(isPubkeyBundle("abc")).toEqual(false);
+      expect(isPubkeyBundle({})).toEqual(false);
+      expect(isPubkeyBundle(null)).toEqual(false);
+      expect(isPubkeyBundle(undefined)).toEqual(false);
+      expect(isPubkeyBundle(fromHex("aabb"))).toEqual(false);
+
+      const missingAlgo: Omit<PubkeyBundle, "algo"> = {
```
willclarktech: 👌

packages/iov-bcp/src/transactions.spec.ts
```diff
@@ -1,6 +1,152 @@
...
+        data: fromHex("aabbccdd") as PubkeyBytes,
+      };
+      expect(isPubkeyBundle(missingAlgo)).toEqual(false);
+
+      const missingData: Omit<PubkeyBundle, "data"> = {
+        algo: Algorithm.Ed25519,
+      };
+      expect(isPubkeyBundle(missingData)).toEqual(false);
+    });
```
willclarktech: Expected behaviour with additional properties?

packages/iov-bcp/src/transactions.spec.ts
```diff
@@ -1,6 +1,152 @@
...
+          data: fromHex("aabbccdd") as PubkeyBytes,
+        },
+      };
+      expect(isIdentity(missingChainId)).toEqual(false);
+
+      const missingPubkey: Omit<Identity, "pubkey"> = {
+        chainId: "foobar" as ChainId,
+      };
+      expect(isIdentity(missingPubkey)).toEqual(false);
+    });
```
willclarktech: As above

packages/iov-bcp/src/transactions.ts
```diff
@@ -3,6 +3,33 @@ import { As } from "type-tagger";
 
 import { Hash, Preimage } from "./atomicswaptypes";
 
+/**
+ * Checks is data is a non-null object (i.e. matches the TypeScript object type)
```
willclarktech: "Checks if"

packages/iov-bcp/src/transactions.ts
```diff
@@ -3,6 +3,33 @@ import { As } from "type-tagger";
...
+ *
+ * Only used package-internal.
```
willclarktech: "internally to this package"

packages/iov-bcp/src/transactions.ts
```diff
@@ -3,6 +3,33 @@ import { As } from "type-tagger";
...
+ */
+export function isNonNullObject(data: unknown): data is object {
+  return typeof data === "object" && data !== null;
+}
+
+/** Checks if data is an Uint8Array. Note: Buffer is treatet as not a Uint8Array */
```
willclarktech: "treated"

packages/iov-bcp/src/transactions.ts
```diff
@@ -15,13 +42,12 @@ export interface PubkeyBundle {
   readonly data: PubkeyBytes;
 }
 
-export function isPubkeyBundle(data: any): data is PubkeyBundle {
+export function isPubkeyBundle(data: unknown): data is PubkeyBundle {
```
willclarktech: 👌

packages/iov-bcp/src/transactions.spec.ts
webmaster128: Hah! I bend the tests to the behavior of JavaScript. My original expectation was `expect(isNonNullObject([])).toEqual(false);`.

Looking at the definition of the `object` type, an Array is an object:

> object is a type that represents the non-primitive type, i.e. anything that is not number, string, boolean, symbol, null, or undefined.

https://www.typescriptlang.org/docs/handbook/basic-types.html#object

So 👍 for making that explicit

webmaster128:
All comments implemented

PR #1143, opened 2019-07-23
Update weave testdata

webmaster128:


willclarktech:
Why is the data regeneration necessary?

webmaster128:
> Why is the data regeneration necessary?

The testvectors changed in the current version of the weave repository (somewhere between 0.16.0 and 0.19.0). This PR uses the data that matches our weave version: https://github.com/iov-one/weave/tree/v0.19.0/spec/testvectors

This update is part of an attempt to understand why the transactions signed by the Kissnet faucet are rejected by the blockchain at the moment. It looks like none of the tested encodings changed, so this does not help with the faucet issue. However, it makes sense to have the data in bnsd.json match the data we got from weave.

PR #1146, opened 2019-07-24
Move TransactionEncoder into @iov/encoding and fix bug

webmaster128:
PR notes:

- Move TransactionEncoder into @iov/encoding because I want to have it there for a long time
- TransactionEncoder should be renamed to reflect the fact that it can be used to encode all kind of objects at some point in the future
- Fixes bug when directory entry has value `undefined` (see https://github.com/iov-one/ponferrada/pull/411#discussion_r306306578)


packages/iov-encoding/src/transactionencoder.spec.ts
```diff
@@ -1,20 +1,63 @@
...
-} from "@iov/bcp";
-import { Encoding } from "@iov/encoding";
-
+import { Encoding } from "./encoding";
 import { TransactionEncoder } from "./transactionencoder";
+import { isUint8Array } from "./typechecks";
 
 const { fromJson, toJson } = TransactionEncoder;
 
+/** and example of how this could potentially be used */
```
willclarktech: "an"

PR #1151, opened 2019-07-24
Fix BNS encoding of validators

willclarktech:


willclarktech:
We can definitely have a test, although the new commit updates the `buildMsg` test which kind of covers it. I'm still waiting for feedback from backend about how this all works, but there's an example here: https://github.com/iov-one/weave/blob/7fab2839969ce8842c3d679ecc5ff29171f3fc24/cmd/bnsd/scenarios/validator_test.go#L50-L58

willclarktech:
For the record I believe the structure is modelled on this: https://github.com/tendermint/tendermint/blob/master/abci/types/pubkey.go

See eg: https://github.com/iov-one/weave/blob/31011e9a478231ad033f786555bef22dcf6ee1d9/types.go#L54-L59

Issue #1114, opened 2019-07-04
Add integration tests to bns-governance


willclarktech:
To check proposal process:
creating -> posting -> getting -> voting -> getting

PR #1154, opened 2019-07-25
Add integration tests for BNS governance module

willclarktech:
Part 1 of #1114
Part 2 will cover the guarantee fund/distribution

Two improvements that I hope can be deferred to a later issue/PR:
1. The tests are a little flaky because of timing issues: it seems that the amount of time it takes to handle proposal creation is a little variable, so sometimes the vote transaction comes in too soon. The only way I can see to solve this reliably is to increase timings, i.e. make the test suite substantially slower.
1. The tests don't currently do any teardown, so the blockchain should be restarted before running the suite each time.


willclarktech:
Probably a timing issue like the vote transaction arriving before the proposal transaction has been fully processed.

PR #1157, opened 2019-07-25
Improve HdPaths in keycontrol

willclarktech:
Adds bip44LikeWithCustomPurpose static method, plus purposes and coinTypes constants

### DIFF
--- a/packages/iov-keycontrol/src/hdpaths.ts
+++ b/packages/iov-keycontrol/src/hdpaths.ts
@@ -12,13 +12,6 @@ export class HdPaths {
   }
 
   /**
-   * This function allows custom purposes e.g. for use by the faucet
-   */
-  public static bip43(...indices: readonly number[]): readonly Slip10RawIndex[] {
-    return indices.map(Slip10RawIndex.hardened);
-  }
-
-  /**
    * Only use this for the real BIP-0044 with
    * - 5 component path (m / purpose' / coin_type' / account' / change / address_index)
    * - no ed25519 support (due to the use of unhardened path components)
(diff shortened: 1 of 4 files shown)


packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -54,7 +66,26 @@ export class HdPaths {
...
+      Slip10RawIndex.hardened(HdPaths.purposes.bip44),
+      Slip10RawIndex.hardened(coinType),
+      Slip10RawIndex.hardened(account),
+    ];
+  }
+
+  /**
+   * This function is the same as bip44Like but allows custom purposes, e.g. for use by the faucet
+   */
+  public static bip44LikeWithCustomPurpose(
```
webmaster128: _BIP44 with custom purpose_ does not make much sense, since no aspect of BIP44 is left then. It is a plain BIP43.

From BIP44:

> This BIP is a particular application of BIP43. 

However, I don't see how this function is useful at all, since the faucet needs a 4-component path anyways.

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -1,15 +1,28 @@
...
+  public static readonly coinTypes = {
+    eth: 60,
+    iov: 234,
+  };
+
   /**
    * IOV's SimpleAddress derivation path
    *
-   * @see https://github.com/iov-one/iov-core/blob/v0.6.1/docs/KeyBase.md#simple-addresses
+   * @see https://github.com/iov-one/iov-core/blob/master/docs/address-derivation-v1.md#simple-addresses-deprecated
```
webmaster128: Can we use a specific tag here instead of master to ensure the link works (even if it gets outdated over time)

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -1,15 +1,28 @@
 import { Slip10RawIndex } from "@iov/crypto";
 
 export class HdPaths {
+  public static readonly purposes = {
```
webmaster128: why public?

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -1,15 +1,28 @@
...
+    bip44: 44,
+    iov: 4804438,
+    iovFaucet: 1229936198,
+  };
+  /**
+   * Coin types as registered at
+   * https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+   */
+  public static readonly coinTypes = {
```
webmaster128: why public?

packages/iov-keycontrol/src/hdpaths.ts
willclarktech: Mostly so we don't have to type out the faucet number, but maybe this shouldn't be in the class anyway.

webmaster128: But `private` would do the job as well if there is a faucet path helper

PR #1160, opened 2019-07-29
Upgrade weave to 0.20

webmaster128:
https://github.com/iov-one/weave/blob/master/CHANGELOG.md#0200

### DIFF
--- a/packages/iov-bns/README.md
+++ b/packages/iov-bns/README.md
@@ -16,7 +16,7 @@ Simplest usage, to use auto-detecting tendermint client and standard bns
 transaction parser:
 
 ```ts
-const connection = await BnsConnection.establish("wss://bns.hugnet.iov.one");
+const connection = await BnsConnection.establish("wss://rpc.lovenet.iov.one");
 ```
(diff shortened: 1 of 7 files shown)


packages/iov-bns/README.md
```diff
@@ -16,14 +16,14 @@ Simplest usage, to use auto-detecting tendermint client and standard bns
 transaction parser:
 
 ```ts
-const connection = await BnsConnection.establish("wss://bns.hugnet.iov.one");
+const connection = await BnsConnection.establish("wss://rpc.lovenet.iov.one");
```
willclarktech: When should `bns` be used and when `rpc`?

webmaster128: `bns.` was renamed to `rpc.` to express the fact that we are not running two blockchains per network anymore. The first networks that use the new name are
* Public testnet: 😍net (https://github.com/iov-one/ops/issues/187)
* Internal testnet: 🐗net (https://github.com/iov-one/ops/issues/186)

PR #1164, opened 2019-07-30
Fix transactions data in EthereumConnection.liveTx 

webmaster128:


packages/iov-bcp/src/transactions.ts
```diff
@@ -216,7 +216,7 @@ export function isUnsignedTransaction(data: unknown): data is UnsignedTransactio
 
 /** A signable transaction knows how to serialize itself and how to store signatures */
 export interface SignedTransaction<T extends LightTransaction = UnsignedTransaction> {
-  /** transaction is the user request */
+  /** The transaction content. Keep this compatible to the ConfirmedTransaction interface */
```
willclarktech: What does "keep this compatible" mean exactly?

packages/iov-bcp/src/transactions.ts
```diff
@@ -225,6 +225,56 @@ export interface SignedTransaction<T extends LightTransaction = UnsignedTransact
...
+  /** a unique identifier (hash of the transaction) */
+  readonly transactionId: TransactionId;
+  /** application specific data from executing tx (result, code, tags...) */
+  readonly result?: Uint8Array;
+  /**
+   * Application specific logging output in an arbitrary text format that
+   * may change at any time.
+   */
+  readonly log?: string;
+  // readonly tags: ReadonlyArray<Tag>;
```
willclarktech: For later inclusion?

packages/iov-bcp/src/transactions.ts
webmaster128: It means `SignedTransaction` and `ConfirmedTransaction` should both have a property with the same name and the same type, such that the intersection `SignedTransaction & ConfirmedTransaction` requires setting the transaction only once to conform to both interfaces at the same time.

Ideas to improve the comment?

packages/iov-bcp/src/transactions.ts
webmaster128: This code is carried around since a time when the sketched BCP types were not even compiled. Time to remove it

packages/iov-bcp/src/transactions.ts
willclarktech: Something like "The generic specified here should remain the same as the corresponding generic defined for the `ConfirmedTransaction` interface."?

willclarktech: But isn't the compatibility going to be ensured anyway when we define the `ConfirmedAndSignedTransaction` type?

webmaster128: When you intersect

```ts
interface SignedTransaction<T> {
  readonly transaction: T,
}
```

and

```ts
interface ConfirmedTransaction<T> {
  readonly original: T,
}
```

the result looks like

```
interface ConfirmedAndSignedTransaction<T> {
  readonly transaction: T,
  readonly original: T,
}
```

willclarktech: Where's this `original` key coming from?

webmaster128: The example I gave uses the property name `transaction` in one interface and the property name `original` in the other interface to demonstrate that sharing the same type `T` is not sufficient for the compatibility I am talking about

willclarktech: Oh I see, we're also worried about people changing the property names. Hm, can't we enforce this with another type? Something like this?
```ts
interface TransactionWrapper<T extends LightTransaction = UnsignedTransaction> {
  transaction: T,
}

interface ConfirmedTransaction extends Transaction {}
```

webmaster128: Good idea! Like this? https://github.com/iov-one/iov-core/pull/1164/commits/11d06bd3829f1030f4df14450bff8ee658c8e9c6

willclarktech: Beautiful!

PR #1165, opened 2019-07-31
Bech32 cleanup

webmaster128:


willclarktech:
Why do we need the custom types?

webmaster128:
> Why do we need the custom types?

1. Because the types from @types are wrong in many aspects
2. It allows us to pass binary data as Uint8Array directly instead of creating a copy to Buffer

willclarktech:
Is it worth submitting a PR to the DefinitelyTyped project?

webmaster128:
> Is it worth submitting a PR to the DefinitelyTyped project?

I'm sumitting it to the project directly: https://github.com/bitcoinjs/bech32/pull/29#issuecomment-516733500

PR #1166, opened 2019-07-31
Prettify all the type declarations

### LINKED ISSUE #1106: Run prettier on .d.ts files
Some time ago we decided to avoid formatting auto-generated .d.ts, as they come with consistency by default. However, as the number of exported types increase, we get very long lines like https://github.com/iov-one/iov-core/blob/v0.15.0/packages/iov-bns/types/index.d.ts#L5, which are hard to review and which cause merge conflicts, that create manual work.

To improve readability, diffs and mergability, we should run prettier on them.

webmaster128:
Closes #1106


webmaster128:
> Should @iov/cli's package.json match the same pattern for build/format-types?

I don't think it makes sense to prettify the build/ folder, which is not checked-in. We run prettier consistenty after mocing typed into the ./types folder, which has the advantage that we don't run prettier in spec files. I'd leave it as it is unless we have a good reason to change.

PR #1168, opened 2019-07-31
 Add optional error logging callback to SigningServerCore 

### LINKED ISSUE #1153: Allow setting error logger in SigningServerCore
In 

```ts
    try {
      authorized = await this.authorizeSignAndPost(reason, transaction, meta);
    } catch (error) {
      // don't expose callback error details over the server
      throw new Error("Internal server error");
    }
```

the server error is not propagated to the web app in oder to avoid leaking information to a potentially untrusted party. As a developer, I want to be able to see the error in the logs. It would be good if the SigningServerCore allowed setting a log function in the constructor that is used to log the error. The interface should alllow passing `console.error` in there directly.

webmaster128:
Closes #1153

### DIFF
--- a/packages/iov-multichain/src/signingservercore.ts
+++ b/packages/iov-multichain/src/signingservercore.ts
@@ -57,17 +57,20 @@ export class SigningServerCore {
   private readonly authorizeGetIdentities: GetIdentitiesAuthorization;
   private readonly authorizeSignAndPost: SignAndPostAuthorization;
   private readonly signedAndPostedProducer = new DefaultValueProducer<readonly SignedAndPosted[]>([]);
+  private readonly logError: (error: any) => void;
 
   public constructor(
     profile: UserProfile,
     signer: MultiChainSigner,
     authorizeGetIdentities: GetIdentitiesAuthorization,
     authorizeSignAndPost: SignAndPostAuthorization,
+    logError?: (error: any) => void,
   ) {
     this.signer = signer;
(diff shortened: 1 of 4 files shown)


packages/iov-multichain/src/signingservercore.ts
```diff
@@ -57,17 +57,20 @@ export class SigningServerCore {
   private readonly authorizeGetIdentities: GetIdentitiesAuthorization;
   private readonly authorizeSignAndPost: SignAndPostAuthorization;
   private readonly signedAndPostedProducer = new DefaultValueProducer<readonly SignedAndPosted[]>([]);
+  private readonly logError: (error: any) => void;
```
willclarktech: Couldn't we make the `error` an `Error`?

webmaster128: Good idea, but …

packages/iov-multichain/src/signingservercore.ts
```diff
@@ -92,6 +95,7 @@ export class SigningServerCore {
     try {
       authorizedIdentities = await this.authorizeGetIdentities(reason, matchingIdentities, meta);
     } catch (error) {
+      this.logError(error);
```
webmaster128: … we don't know what we are catching here. This could be anything and is not in IOV-Core's control. A caller you throw a string or a number or any kind of object that we catch here.

Issue #1097, opened 2019-06-27
Make iov-cli work with NodeJs 12+


webmaster128:
We get the error

```
>> (node:71798) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_REPL_INPUT] [ERR_INVALID_REPL_INPUT]: Listeners for `uncaughtException` cannot be used in the REPL
    at process.<anonymous> (repl.js:224:15)
    at process.emit (events.js:205:15)
    at process.EventEmitter.emit (domain.js:471:20)
    at process.emit (/place/node_modules/source-map-support/source-map-support.js:461:21)
    at _addListener (events.js:226:14)
    at process.addListener (events.js:274:10)
    at Object.C.useBackupModule (/place/node_modules/libsodium/dist/modules/libsodium.js:1:1244)
    at /place/node_modules/libsodium/dist/modules/libsodium.js:1:502840
    at processTicksAndRejections (internal/process/task_queues.js:89:5)
(node:71798) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
```

after https://github.com/nodejs/node/commit/c709c524fd674f3762b3da9e005f645255bc1e2f because of code compiled into `node_modules/libsodium/dist/modules/libsodium.js`.

webmaster128:
Filed a bugreport in libsodium.js: https://github.com/jedisct1/libsodium.js/issues/205

willclarktech:
@webmaster128 How do I reproduce this error?

webmaster128:
> How do I reproduce this error?

* Build project
* Install nodeJs 12 or higher and make it the default for `#!/usr/bin/env node`
* Go to `packages/iov-cli` and run `./bin/iov-cli`

willclarktech:
PR submitted to libsodium: https://github.com/jedisct1/libsodium/pull/853

webmaster128:
I subscribed to https://github.com/jedisct1/libsodium.js releases. Let's see and wait. For now we can see this as a known limitation.

webmaster128:
There is a new version of libsodium.js available. `"libsodium-wrappers": "^0.7.6",` should include the fix (untested)

PR #1173, opened 2019-08-01
Proof of concept: Fix libsodium uncaught error handler

willclarktech:
Addresses #1097

I wanted to make a draft PR to make sure everything was fine before submitting a PR to libsodium.

### DIFF
--- a/packages/iov-crypto/package.json
+++ b/packages/iov-crypto/package.json
@@ -36,7 +36,7 @@
     "bn.js": "^4.11.8",
     "elliptic": "^6.4.0",
     "js-sha3": "^0.8.0",
-    "libsodium-wrappers": "^0.7.5",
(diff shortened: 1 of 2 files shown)


webmaster128:
Removing from board since there is nothing to do here

webmaster128:
https://github.com/jedisct1/libsodium.js/commits/master was updated (but not yet released). I think this is a good chance to check again if the patch was applied as we need it

willclarktech:
@webmaster128 Doesn't look like it: https://github.com/jedisct1/libsodium/commits/stable

webmaster128:
The patch was cherry-picked on stable now. Should be in current libsodium.js master (https://github.com/jedisct1/libsodium.js/commit/96a07a701dba96b1c0574a235691a7d4a8c8d170)

willclarktech:
Just checked, it is.

webmaster128:
Can we close this PR?

willclarktech:
I think so, the fix should be included in an official release at some point and we're following those releases.

PR #1176, opened 2019-08-06
Create bytesToSign test data

webmaster128:


packages/iov-bns/src/bnscodec.spec.ts
```diff
@@ -14,8 +29,84 @@ import {
   swapClaimTxJson,
   swapOfferTxJson,
 } from "./testdata.spec";
+import { encodeBnsAddress } from "./util";
 
 describe("bnscodec", () => {
+  fit("generate test data", () => {
+    const nonces = [0 as Nonce, 1 as Nonce, Number.MAX_SAFE_INTEGER as Nonce] as const;
```
willclarktech: I never tried property-based testing in JS/TS but might be worth looking into something like this: https://github.com/dubzzz/fast-check

packages/iov-bns/src/bnscodec.spec.ts
```diff
@@ -14,8 +29,339 @@ import {
...
+        {
+          address: bnsCodec.identityToAddress(newbie),
+          weight: 5,
+        },
+      ];
+      for (const multisig of multisigs) {
+        const send: UpdateMultisignatureTx & MultisignatureTx = {
+          kind: "bns/update_multisignature_contract",
+          chainId: participant.chainId,
+          contractId: new Uint8Array([1]),
```
davepuchyr: @webmaster128 , I will change this in conjunction with https://github.com/iov-one/iov-core/pull/1387.

packages/iov-bns/src/bnscodec.spec.ts
```diff
@@ -14,8 +29,339 @@ import {
...
+        {
+          address: bnsCodec.identityToAddress(newbie),
+          weight: 5,
+        },
+      ];
+      for (const multisig of multisigs) {
+        const update: UpdateMultisignatureTx & MultisignatureTx = {
+          kind: "bns/update_multisignature_contract",
+          chainId: participant.chainId,
+          contractId: new Uint8Array([0, 0, 0, 1]),
```
webmaster128: this must be 8 bytes, not 4

davepuchyr: Oops.

Why aren't `proposalId`, `escrowId`, etc `Uint8Array`s?  I dug into weave code and saw that `proposalId` is [8](https://github.com/iov-one/weave/blob/625c72a0414b5f0ee87bcf498a509800a3e3e353/x/gov/msg.go#L65) bytes.

webmaster128: I answered this in the other thread

PR #1177, opened 2019-08-07
Fix wording in governance integration tests

willclarktech:


webmaster128:
@willclarktech could you look into this test fail? Seems to be something serious

willclarktech:
Whoops yes, missed an assertion. Should be fine now.

PR #1178, opened 2019-08-07
Add script to BNSD deployment to initialise proposals

willclarktech:
Part of https://github.com/iov-one/ponferrada/issues/495

### DIFF
--- a/scripts/bnsd/start.sh
+++ b/scripts/bnsd/start.sh
@@ -25,3 +25,5 @@ sleep 3
 # for debug output
 cat "$LOGS_FILE_TM"
 cat "$LOGS_FILE_APP"
+
+"${SCRIPT_DIR}"/init.sh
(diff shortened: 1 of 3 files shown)


scripts/bnsd/start.sh
```diff
@@ -25,3 +25,5 @@ sleep 3
 # for debug output
 cat "$LOGS_FILE_TM"
 cat "$LOGS_FILE_APP"
+
+"${SCRIPT_DIR}"/init.sh
```
webmaster128: Can we wrap this call in `INIT_PROPOSALS` env variable instead? Such that when you call `init.sh`, you get an init. You might change your mind when the blockchain started already

willclarktech: My idea was that `init.sh` might eventually include a few different things to initialise, so we just run the script whatever the case here and inside the file it finds out which things you want it to init. Happy either way though.

webmaster128: Works for me. I don't have any strong points for either solution. Lets start with the current one and adapt if needed.

PR #1180, opened 2019-08-07
Test EnglishMnemonic.wordlist against spec

### LINKED ISSUE #1179: Check bip39WordlistEnglish against BIP39
Right now we use a wordlist from an npm package and we hope it is correct. I would like to check the value of

```ts
import bip39WordlistEnglish from "bip39/wordlists/english.json";
```

against the file `english.txt` from https://github.com/bitcoin/bips/tree/master/bip-0039 (sha256:2f5eed53a4727b4bf8880d8f3f199efc90e58503646d9ff8eff3a2ed3b24dbda) and verify:

1. We can load `english.txt` from the speck into a test (binary) and the checksum matches
2. When split line-by-line, we get the exact same wordlist as `bip39WordlistEnglish`

This way we can verify the wordlist in the package is the same as in the spec. Looking at some of the latest commits to the wordlist, it seems worth checking them bit-by-bit:

- https://github.com/bitcoin/bips/commit/50c4f1255eed6b1c08ca9c20b4d8f380879ed2f5
- https://github.com/bitcoin/bips/commit/144ef37e5314f8f968b8abb5a6fe8038d32e46ce
- https://github.com/bitcoin/bips/commit/25f10b9f7bf844ebd98dc9ca6556228c3c34c44b

webmaster128:
Closes #1179


packages/iov-crypto/src/englishmnemonic.spec.ts
```diff
@@ -1,6 +1,36 @@
...
+import { Sha256 } from "./sha";
+import wordlists from "./testdata/bip39_wordlists.json";
+
+const { fromAscii, fromHex } = Encoding;
 
 describe("EnglishMnemonic", () => {
+  describe("wordlist", () => {
+    it("matches the words from the bitcoin/bips/bip-0039 spec", () => {
+      const lineFeed = 0x0a;
+      const bip39EnglishTxt = fromHex(wordlists["english.txt"]);
```
willclarktech: Could we change the key to just `english`? Also why is this hex encoded?

packages/iov-crypto/src/englishmnemonic.spec.ts
```diff
@@ -1,6 +1,36 @@
...
+      const checksum = new Sha256(bip39EnglishTxt).digest();
+      expect(checksum).toEqual(fromHex("2f5eed53a4727b4bf8880d8f3f199efc90e58503646d9ff8eff3a2ed3b24dbda"));
```
willclarktech: Could we have a link for this in a comment?

packages/iov-crypto/src/englishmnemonic.spec.ts
webmaster128: We could. I used english.txt to make explicit that this is english.txt (from https://github.com/bitcoin/bips/tree/master/bip-0039)

If we want test data that would equally well in browsers and nodejs, I think we need to use json, which we can import as a module. nodejs will read it from disk and webpack will inline it. But correct me of I'm wrong.

The choice between hex or base64 encoding was random.

What do you think?

willclarktech: I think if conciseness is something to prioritise then base64 would be better. But what's wrong with utf8? Since this issue is about making sure we have the correct wordlist, anything that's not human-readable is going to make it more difficult to notice if we've got out of sync with the spec somehow.

webmaster128: The point is I want to store and process binary data and avoid any kind of string encoding between english.txt and `bip39EnglishTxt`. As soon as we use a JSON string, we receive a string in JavaScript that involves a bunch of implicit conversion logic (UTF8 source code to unicode string). I cannot test the checksum of a string directly.

I'm updating to `.english` and base64 now, The additional comment is coming too.

webmaster128:
All implemented

PR #1193, opened 2019-08-13
Export pubkeyToAddress from @iov/bns 

webmaster128:
This allows address generation in the IOV Address Generator as well as the Token Finder without using a fake testnet chain ID.

### DIFF
--- a/packages/iov-bns/src/util.ts
+++ b/packages/iov-bns/src/util.ts
@@ -58,7 +58,8 @@ function keyToIdentifier(key: PubkeyBundle): Uint8Array {
 }
 
 /**
- * Creates an IOV address from a given Ed25519 pubkey and a prefix that represents the network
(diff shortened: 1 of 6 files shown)


packages/iov-bns/src/util.ts
```diff
@@ -57,10 +57,21 @@ function keyToIdentifier(key: PubkeyBundle): Uint8Array {
   return Uint8Array.from([...algoToPrefix(key.algo), ...key.data]);
 }
 
+/**
+ * Creates an IOV address from a given Ed25519 pubkey and a prefix that represents the network
```
willclarktech: Suggest "represents the network kind (ie mainnet or testnet)"

webmaster128: Added, thanks

PR #1197, opened 2019-08-14
Make CashConfiguration.minimalFee nullable

### LINKED ISSUE #1192: Make CashConfiguration.minimalFee nullable
In

```ts
export function decodeCashConfiguration(config: codecImpl.cash.Configuration): CashConfiguration {
  const minimalFee = decodeAmount(ensure(config.minimalFee, "minimalFee"));
  return {
    minimalFee: minimalFee,
  };
}
```

we need to handle the case that config.minimalFee null (or better falsy in general). The type of `CashConfiguration.minimalFee` becomes `Amount | null`.

See https://github.com/iov-one/weave/issues/921

willclarktech:
Closes #1192

### DIFF
--- a/packages/iov-bns/src/decode.spec.ts
+++ b/packages/iov-bns/src/decode.spec.ts
@@ -12,6 +12,7 @@ import { Bech32, Encoding } from "@iov/encoding";
 
 import {
   decodeAmount,
+  decodeCashConfiguration,
   decodeElectionRule,
   decodeElectorate,
   decodePrivkey,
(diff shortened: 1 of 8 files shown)


packages/iov-bns/src/decode.spec.ts
```diff
@@ -194,6 +195,46 @@ describe("Decode", () => {
     });
   });
 
+  describe("decodeCashConfiguration", () => {
+    it("can decode configuration with non-null minimal fee", () => {
+      const config: codecImpl.cash.Configuration = {
```
webmaster128: Can you change the argument of `decodeCashConfiguration` to `cash.IConfiguration`? Then we can we can use data-only `cash.IConfiguration` here as well and don't need the dummy `toJSON`. Also `owner` and `collectorAddress` become optional and can be skipped for now

PR #1201, opened 2019-08-15
Add reconnected handler to tendermint websocket client

### LINKED ISSUE #1200: Use reconnectedHandler from ReconnectingSocket
Looks like we closed https://github.com/iov-one/iov-core/issues/229 a bit too early. We now have `ReconnectingSocket`, which accepts a `reconnectedHandler`. However, this handler is never used.

What we need for https://github.com/iov-one/ponferrada/issues/14 is clients that keep track of subsciptions they have. As soon as the socket reconnects, those subscriptions need to be recreated. This affects `WebsocketClient` from the Tendermint package and `WsEthereumRpcClient`.

This is needed for the browser extension to track transaction state after `postTx`. For Ethereum we use polling in this case, so this is lower priority. But in case of BNS, we use both transaction and block subscriptions in `postTx`, so this is needed for https://github.com/iov-one/ponferrada/issues/14.

- [ ] Use reconnectedHandler in Tendermint
- [ ] Use reconnectedHandler in Ethereum

willclarktech:
Closes #1200

I’m a little unclear on the acceptance criteria for this issue, so opening this draft PR now to get feedback. Is this basically what we need or is there something I’ve missed?


packages/iov-tendermint-rpc/src/rpcclients/websocketclient.ts
```diff
@@ -212,4 +213,13 @@ export class WebsocketClient implements RpcStreamingClient {
   protected async responseForRequestId(id: JsonRpcId): Promise<JsonRpcResponse> {
     return firstEvent(this.jsonRpcResponseStream.filter(r => r.id === id));
   }
+
+  private reconnectedHandler(): void {
+    [...this.subscriptionStreams.keys()].forEach(query => {
+      const request = createJsonRpcRequest("subscribe", { query: query });
+      const producer = new RpcEventProducer(request, this.socket);
+      const stream = Stream.create(producer);
+      this.subscriptionStreams.set(query, stream);
```
webmaster128: This sets `subscriptionStreams[subscriptionStreams]` to a new reference. The old stream the caller got via `const eventStream = this.client.listen(req);` will not change, so the new events do not arrive.

Also this re-establishes subscribtions that nobody listens to anymore (where the original producer was stopped).

My feeling is that in `public listen(request: JsonRpcRequest): Stream<SubscriptionEvent> {` there needs to be a producer (maybe a wrapper around `RpcEventProducer`) which survives the reconnect and routes events from different sources.

packages/iov-tendermint-rpc/src/rpcclients/websocketclient.ts
```diff
@@ -131,6 +131,40 @@ class RpcEventProducer implements Producer<SubscriptionEvent> {
   }
 }
 
+class ReconnectingRpcEventProducer implements Producer<SubscriptionEvent> {
+  public stopped: boolean = false;
```
webmaster128: this can become private, right?

willclarktech: Yes, thanks, I originally had the check outside the class.

willclarktech:
@webmaster128 I’ll have a think about tests and then convert this to a non-draft PR.

PR #1206, opened 2019-08-21
Reimplement Random.getBytes

### LINKED ISSUE #1203: Re-implement Random.getBytes
~For the 0.16 branch~ either 0.16 or master is good since we will release 0.17.0 soon

------

The implemention

```ts
export class Random {
  // Get `count` bytes of cryptographically secure random bytes
  public static async getBytes(count: number): Promise<Uint8Array> {
    await sodium.ready;
    return sodium.randombytes_buf(count);
  }
}
```

should be replaced with native entropy sources instead of calling from JS through webassebly compiled C into an injected JS entropy source to make it easy to review.

For inspiration, this is what is injected into libsodium: https://github.com/jedisct1/libsodium/blob/1.0.18/src/libsodium/randombytes/randombytes.c#L88-L112

We should error when no cryptographically secure entropy source is found instead of falling back to something insecure.

webmaster128:
Closes #1203

### DIFF
--- a/packages/iov-crypto/src/random.ts
+++ b/packages/iov-crypto/src/random.ts
@@ -1,0 +1,27 @@
+declare const window: any;
+declare const self: any;
+
+export class Random {
+  /**
+   * Returns `count` cryptographically secure random bytes
+   */
+  public static getBytes(count: number): Uint8Array {
+    try {
+      const globalObject = "object" === typeof window ? window : self;
+      const cryptoApi =
+        typeof globalObject.crypto !== "undefined" ? globalObject.crypto : globalObject.msCrypto;
+
+      const out = new Uint8Array(count);
+      cryptoApi.getRandomValues(out);
+      return out;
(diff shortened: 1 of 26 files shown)


packages/iov-crypto/src/random.ts
```diff
@@ -0,0 +1,27 @@
+declare const window: any;
+declare const self: any;
+
+export class Random {
+  /**
+   * Returns `count` cryptographically secure random bytes
+   */
+  public static getBytes(count: number): Uint8Array {
+    try {
+      const globalObject = "object" === typeof window ? window : self;
```
willclarktech: I was momentarily confused by this because I think I've only ever seen "typeof x === y" never the other way around.

webmaster128: Oh, yeah this is an artifact from external code. Will update since I don't like it as well.

PR #1207, opened 2019-08-22
Update to weave v0.21.0

### LINKED ISSUE #1182: Adapt to new weave VersionedIDRef format
The fields that currently use the `VersionedIDRef` will change to 16 bytes as described in https://github.com/iov-one/weave/issues/918.

For this we need to adapt 

```ts
function decodeVersionedId(versionedId: codecImpl.orm.IVersionedIDRef): VersionedId {
  return {
    id: decodeNumericId(ensure(versionedId.id, "id")),
    version: ensure(versionedId.version, "version"),
  };
}
```

and pull out the two components using

```ts
function decodeVersionedId(versionedId: Uint8Array): VersionedId {
  return {
    id: Uint64.fromBytesBigEndian(versionedId.slice(0, 8)).toNumber(),
    version: Uint64.fromBytesBigEndian(versionedId.slice(8, 16)).toNumber(),
  };
}
```

This requires the next breaking release of weave.

willclarktech:
Closes #1182

Opening a draft PR while waiting for weave to be available from DockerHub.

### DIFF
--- a/packages/iov-bns/src/decode.ts
+++ b/packages/iov-bns/src/decode.ts
@@ -21,7 +21,7 @@ import {
   UnsignedTransaction,
   WithCreator,
 } from "@iov/bcp";
-import { Encoding } from "@iov/encoding";
+import { Encoding, Uint32, Uint64 } from "@iov/encoding";
 import BN from "bn.js";
 import * as Long from "long";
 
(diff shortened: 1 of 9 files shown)


packages/iov-bns/src/decode.ts
```diff
@@ -107,6 +107,13 @@ function decodeVersionedId(versionedId: codecImpl.orm.IVersionedIDRef): Versione
   };
 }
 
+function decodeVersionedIdArray(versionedId: Uint8Array): VersionedId {
+  return {
```
webmaster128: Can you throw an error if the input does not have the expected length?

packages/iov-bns/src/decode.ts
```diff
@@ -107,6 +107,13 @@ function decodeVersionedId(versionedId: codecImpl.orm.IVersionedIDRef): Versione
...
+    id: Uint64.fromBytesBigEndian(versionedId.slice(0, 8)).toNumber(),
+    version: Uint64.fromBytesBigEndian(versionedId.slice(8, 16)).toNumber(),
```
webmaster128: The final imlementation ended up encoding the versin as Uint32 (4 bytes): https://github.com/iov-one/weave/pull/950/files, so we can use `Uint32` instead

willclarktech: To clarify: 12 bytes total (8 for id, 4 for version)?

webmaster128: Exactly

webmaster128:
In addition to the current diff, we need to configure the name/namespace patterns in the genesis file (see https://github.com/iov-one/weave/pull/980/files#diff-19980658f2035866d4b2bef354b2827eR35-R40)

Issue #638, opened 2018-12-21
Evaluate and trim down dependencies


ethanfrey:
The recent [event-stream exploit](https://schneid.io/blog/event-stream-vulnerability-explained/) leads to renewed concern about security of dependencies.

The easiest steps to take are to limit the number of dependencies, and maximize the quality of the dependencies. (Before starting to audit them all). By quality, we can usually look at activity and reputation of maintainers.

To start with, it would be good to determine what dependencies we actually install when we import the library, rather than those used to compile and test (exclude devDependencies). From that list, we should see if anything stands out as dangerous to be eliminated, and if there are any easy ones to trim (only used by one place in the code, easy to pull out, etc).

I consider this issue a first pass maintenance chore, to get current state mainly, not a deep audit.

ethanfrey:
Current result of `yarn install --prod` yields 157 packages in node_modules. (Maybe more, I think we need `--flat` as well to really show them all).

A quick glance shows some likely redudancy if nothing else:
```
safe-buffer
safer-buffer
```

As well as what look like devDependencies:
```
source-map
source-map-support
```

This is not high priority, but more just some repo dusting for a rainy day when the mind is not ready for hard coding...

nB: I have discovered many npm repos to be in effect one index.js file with less than 10 lines of code. Since this is usually simple logic not going to change, we can copy/reimplement (depending on license) and place in a util package.
Example: https://github.com/brummelte/sleep-promise to allow `await sleep`. Overkill, huh? (v8.0.1 on github???)

webmaster128:
Reducing dependencies is a continuously ongoing operation with the following steps done most recently:

- Convert Ganache CLI installation from npm to docker (removed 1/3 of all deps https://github.com/iov-one/iov-core/pull/625)
- Convert @iov/bns from runtime to dev dependency in @iov/core (https://github.com/iov-one/iov-core/pull/632)
- Implement Ethereum support with only 1 external dependency (https://github.com/iov-one/iov-core/blob/v0.10.2/packages/iov-ethereum/package.json#L33)
- Remove Ripemd160 support asap (https://github.com/iov-one/iov-core/issues/589)

The problem with the packages mentioned above is: we don't import them directly. Those are transient dependencies introduced by other maintainers and as long there is nothing seriously wrong with those packages it will be hard to convince them to remove those packages.

webmaster128:
Great article by the way. Time for reproducible build culture in JS world.

webmaster128:
[...]

PR #1211, opened 2019-08-27
Upgrade dependencies

webmaster128:
The result of a `yarn upgrade`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,9 +3,9 @@
 
 
 "@babel/code-frame@^7.0.0":
-  version "7.0.0"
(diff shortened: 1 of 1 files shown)


webmaster128:
> Do we not also want to bump the package.json?

Is there a command for that? Could be helpful since applications don't get the yarn.lock shipped.

willclarktech:
> Is there a command for that?

Doesn't look like it, but there are community tools like this: https://github.com/vasilevich/sync-yarnlock-into-packagejson#readme

webmaster128:
https://github.com/iov-one/iov-core/issues/638 tracks those kind of questions. I think we don't have great solutions yet. Feel free to explore the options.

For now we follow the approach: update package.json if we have explicit, manual requirements for certain versions. Otherwise use yarn upgrade.

PR #1212, opened 2019-08-27
Add TreasurySend Proposal type

### LINKED ISSUE #1210: Create TreasurySend proposal type
It is expected that not all contributors will have their mainnet address ready before mainnet launch. One option to distibute funds is having a treasury account and use governance to trigger token send from this account.

In order to make this happen, we need a `TreasurySend` proposal type that allows sending tokens to contributors. It should be implemented by a batch SendTransaction to allow for multiple sends in one proposal (backend limit is 15 if I understood correctly). An optional treasury address is required in the `GovernorOptions`.

The new proposal type is very similar to `DistributeFunds`, but with absolute values instead of weights.

willclarktech:
Closes #1210

### DIFF
--- a/packages/iov-bns-governance/src/integrationtests/proposals.spec.ts
+++ b/packages/iov-bns-governance/src/integrationtests/proposals.spec.ts
@@ -3,6 +3,7 @@ import {
   Address,
   isBlockInfoPending,
   isBlockInfoSucceeded,
+  SendTransaction,
   TokenTicker,
   UnsignedTransaction,
   WithCreator,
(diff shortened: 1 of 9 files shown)


packages/iov-bns-governance/src/integrationtests/proposals.spec.ts
```diff
@@ -25,6 +26,7 @@ import { ReadonlyDate } from "readonly-date";
 import { CommitteeId } from "../committees";
 import { Governor, GovernorOptions } from "../governor";
 import { ProposalOptions, ProposalType } from "../proposals";
+import { triggerAsyncId } from "async_hooks";
```
webmaster128: ?

willclarktech: VSCode doesn't understand my habit of writing code that references variables I need before writing the code that declares them.

PR #1214, opened 2019-08-28
Support multisig txs in BNS

### LINKED ISSUE #1119: Support using multisig in eg. SendTx
PR #985 implemented Create and Update Multisig transactions to manage the contracts. However, they didn't expose the field to use them to authorize another transaction. This is still the missing part of #678.

Here is a desired workflow:

* Create two walllets
* Create a 2 of 2 multisig with those 2 wallets
* Send tokens to that multisig
* Create sendtx from that multisig address to another account
* Sign it with one participant -> fails
* Sign with both *and* "activate" multisig -> succeeds

[This is the field](https://github.com/iov-one/iov-core/blob/ffbd2b257de8110a194e5e2107b710d070194f9d/packages/iov-bns/src/generated/codecimpl.d.ts#L103-L105)  that should contain an array of Multisig id's. (The 8 byte responses to create multisig, which are also used in update multisig).

We also need to build a condition (like for aswap) to calculate the address the contract controls to send money to and from the contract.

willclarktech:
Closes #1119

I'm not clear on the ramifications of my changes to `encode.ts`. I also didn't know who should pay the fee in the event that more than one multisignature contract is referenced (probably quite an unusual situation anyway so we may not even want to support it client-side).

### DIFF
--- a/packages/iov-bns/src/encode.ts
+++ b/packages/iov-bns/src/encode.ts
@@ -26,6 +26,7 @@ import {
   isBnsTx,
   isCreateTextResolution,
   isExecuteProposalBatch,
+  isMultisignatureTx,
   isReleaseEscrow,
   isSend,
   isSetValidators,
@@ -44,7 +45,12 @@ import {
   VoteOption,
   VoteTx,
 } from "./types";
-import { decodeBnsAddress, identityToAddress } from "./util";
+import {
+  conditionToWeaveAddress,
+  decodeBnsAddress,
+  identityToAddress,
+  multisignatureCondition,
+} from "./util";
 
 function encodeInt(intNumber: number): number | null {
   if (!Number.isInteger(intNumber)) {
@@ -128,15 +134,10 @@ function encodeNumericId(id: number): Uint8Array {
 // Token sends
 
 function buildSendTransaction(tx: SendTransaction & WithCreator): codecImpl.bnsd.ITx {
-  const { prefix: prefix1, data: data1 } = decodeBnsAddress(identityToAddress(tx.creator));
-  const { prefix: prefix2, data: data2 } = decodeBnsAddress(tx.sender);
-  if (prefix1 !== prefix2 || data1.length !== data2.length || data1.some((b, i) => b !== data2[i])) {
-    throw new Error("Sender and creator do not match (currently unsupported)");
-  }
   return {
     cashSendMsg: codecImpl.cash.SendMsg.create({
       metadata: { schema: 1 },
-      source: decodeBnsAddress(identityToAddress(tx.creator)).data,
+      source: decodeBnsAddress(tx.sender).data,
       destination: decodeBnsAddress(tx.recipient).data,
       amount: encodeAmount(tx.amount),
       memo: encodeString(tx.memo),
@@ -479,15 +480,19 @@ export function buildMsg(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
 
 export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
   const msg = buildMsg(tx);
+  const feePayer = isMultisignatureTx(tx)
+    ? conditionToWeaveAddress(multisignatureCondition(tx.multisig[0]))
+    : decodeBnsAddress(identityToAddress(tx.creator)).data;
   return codecImpl.bnsd.Tx.create({
     ...msg,
     fees:
       tx.fee && tx.fee.tokens
         ? {
             fees: encodeAmount(tx.fee.tokens),
-            payer: decodeBnsAddress(identityToAddress(tx.creator)).data,
+            payer: feePayer,
           }
         : null,
+    multisig: isMultisignatureTx(tx) ? [...tx.multisig] : null,
   });
 }
 
--- a/packages/iov-bns/src/integrationtests/multisignature.spec.ts
+++ b/packages/iov-bns/src/integrationtests/multisignature.spec.ts
@@ -1,0 +1,281 @@
+import {
+  Address,
+  Amount,
+  ChainId,
+  Identity,
+  isBlockInfoPending,
+  isBlockInfoSucceeded,
+  SendTransaction,
+  SignedTransaction,
+  TokenTicker,
+  UnsignedTransaction,
+  WithCreator,
+} from "@iov/bcp";
+import { Slip10RawIndex } from "@iov/crypto";
+import { Ed25519HdWallet, HdPaths, UserProfile } from "@iov/keycontrol";
+import BN from "bn.js";
+
+import { bnsCodec } from "../bnscodec";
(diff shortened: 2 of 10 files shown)


packages/iov-bns/src/integrationtests/multisignature.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  public async getBalance(): Promise<BN> {
+    const account = await this.connection.getAccount({ pubkey: this.identity.pubkey });
+    const balance = account ? account.balance : [];
+    const amount = balance.find(row => row.tokenTicker === CASH);
+    return new BN(amount ? amount.quantity : 0);
+  }
+
+  public async getMultisignatureBalance(address: Address): Promise<BN> {
```
webmaster128: This works for any address, right? Why not merge into

```ts
/** get balance from address. Defaults to currect actor's address */
public async getBalance(address?: Address): Promise<BN> {
```

packages/iov-bns/src/integrationtests/multisignature.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+      tokenTicker: CASH,
+    };
+    const sendTx = await alice.createSendTransaction(
+      multisignatureId,
+      multisignatureAddress,
+      alice.address,
+      amount,
+    );
+    const signedByAlice = await alice.signTransaction(sendTx);
+    const signedByBob = await bob.appendSignature(signedByAlice);
```
webmaster128: wohooo! After all those month I finally see `appendSignature` in action

packages/iov-bns/src/types.ts
```diff
@@ -487,3 +488,14 @@ export function isBnsTx(transaction: LightTransaction): transaction is BnsTx {
...
   return transaction.kind.startsWith("bns/");
 }
+
+export interface MultisigTransaction extends LightTransaction {
+  readonly multisig: readonly Uint8Array[];
+}
+
+export function isMultisigTransaction(transaction: LightTransaction): transaction is MultisigTransaction {
+  return (
+    Array.isArray((transaction as any).multisig) &&
```
webmaster128: Did you test this for a non-MultisigTransaction input? I'd expect `transaction.multisig` to raise a reference error even before it is processed by `Array.isArray`. This can be avoided by a `typeof transaction.multisig === "object"`.

packages/iov-bns/src/types.ts
```diff
@@ -487,3 +488,14 @@ export function isBnsTx(transaction: LightTransaction): transaction is BnsTx {
...
+    (transaction as any).multisig.every((id: any) => id instanceof Uint8Array)
```
webmaster128: `.every(isUint8Array)`

packages/iov-bns/src/encode.ts
```diff
@@ -128,15 +130,10 @@ function encodeNumericId(id: number): Uint8Array {
...
-  const { prefix: prefix1, data: data1 } = decodeBnsAddress(identityToAddress(tx.creator));
-  const { prefix: prefix2, data: data2 } = decodeBnsAddress(tx.sender);
-  if (prefix1 !== prefix2 || data1.length !== data2.length || data1.some((b, i) => b !== data2[i])) {
-    throw new Error("Sender and creator do not match (currently unsupported)");
-  }
   return {
     cashSendMsg: codecImpl.cash.SendMsg.create({
       metadata: { schema: 1 },
-      source: decodeBnsAddress(identityToAddress(tx.creator)).data,
+      source: decodeBnsAddress(tx.sender).data,
```
webmaster128: This change looks good to me. All those consistency checks were implemented when we added `sender` and did not have cases where sender would be different from creator.

packages/iov-bns/src/encode.ts
```diff
@@ -479,15 +476,20 @@ export function buildMsg(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
 
 export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
   const msg = buildMsg(tx);
+  const feePayer = isMultisigTransaction(tx)
+    ? // The chainId is irrelevant because we only need the data
+      conditionToAddress("xxx" as ChainId, multisignatureCondition(tx.multisig[0]))
```
webmaster128: We could pull out a `conditionToWeaveAddress` out of `conditionToAddress` that returns the 20 bytes directly. Since this is @iov/bns internal only, it does not affect caller code where addresses are always bech32 encoded.

packages/iov-bns/src/types.ts
```diff
@@ -487,3 +488,14 @@ export function isBnsTx(transaction: LightTransaction): transaction is BnsTx {
 
   return transaction.kind.startsWith("bns/");
 }
+
+export interface MultisigTransaction extends LightTransaction {
```
webmaster128: Can you export both new types in `index.ts`?

packages/iov-bns/src/integrationtests/multisignature.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+      multisig: [multisignatureId],
+      creator: this.identity,
+      sender: sender,
+      recipient: recipient,
+      amount: amount,
+    });
+  }
+}
+
+fdescribe("Multisignature wallets", () => {
```
webmaster128: `describe`

webmaster128:
Great job by the way!

packages/iov-bns/src/types.ts
webmaster128: hmm, looks like I am wrong when it comes to accessing object properties. I don't remember in which cases the ReferenceError occurs exactly. Feel free to ignore and resolve.

willclarktech:
@webmaster128 

> The doc comment is only for a single contract ID

The comment is, but look at the type. Talking with @husio about the weave code also made clear that more than one contract can be specified, with requirements from each being combined using `AND`.

husio:
In `weave`, multisig validation is implemented as a middleware. The handler iterates over all multisig attached to the transaction:

https://github.com/iov-one/weave/blob/master/x/multisig/decorator.go#L59-L60

And then for every multisig contract, it gathers total power of all signatures present in this transaction and fails if it is not enough signatures power

https://github.com/iov-one/weave/blob/master/x/multisig/decorator.go#L82-L86

Multisig middleware fails on the first contract that did not pass authorization step.

packages/iov-bns/src/integrationtests/multisignature.spec.ts
willclarktech: Better yet, we don't actually use the current `getBalance`, so I'll remove it and rename this one.

willclarktech:
@webmaster128 OK to merge?

webmaster128:
Just a small step for a developer …

Amazing to have this in, tested and almost understood!

willclarktech:
one does not simply

Issue #1216, opened 2019-08-29
Change type of multisig ID in MultisignatureTx to hex


webmaster128:
When we want to use multisig in CLI or Ledger, it is convenient to have a standardized string representation exposed from IOV-Core level. This allows for easy copy/past as well as displaying in the Leder.

I suggest using uppercase hex, which is the tendermint style somehow (like in `TransactionId`). We could potentially use a tagged type like `Address`/`TransactionId` but I'm not really sure if the small additional type safety is worth the effort of writing `as Address` all the time.

webmaster128:
As discussed here (https://github.com/iov-one/iov-core/pull/1218#discussion_r319016960), it is nicer to represent the ID as an uint64.

I just verified that only suceeded transactions from chain need to be parsable, so a value > MAX_SAFE_INTEGER that failed to execute but is in chain still does not get parsed. So we can use JS `number`.

PR #1218, opened 2019-08-29
Support multisig txs in BNS decode

### LINKED ISSUE #1215: Support decoding multisig transactions
As of https://github.com/iov-one/iov-core/pull/1214 we have the ability to post send transactions from a multisig contract. But we currently can't correctly decode such transactions.

willclarktech:
Closes #1215

### DIFF
--- a/packages/iov-bns/src/decode.spec.ts
+++ b/packages/iov-bns/src/decode.spec.ts
@@ -3,6 +3,7 @@ import {
   Algorithm,
   Amount,
   ChainId,
+  isSendTransaction,
   Nonce,
   PubkeyBytes,
   TokenTicker,
@@ -42,6 +43,7 @@ import {
   isCreateEscrowTx,
   isCreateMultisignatureTx,
   isCreateProposalTx,
+  isMultisignatureTx,
   isRegisterUsernameTx,
   isReleaseEscrowTx,
   isReturnEscrowTx,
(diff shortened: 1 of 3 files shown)


packages/iov-bns/src/decode.spec.ts
```diff
@@ -394,6 +396,54 @@ describe("Decode", () => {
...
+              ed25519: fromHex("c9df7bcba2238bedcc681e8b17bb21c1625d21d285b70c20cf53fdd473db9dfb"),
+            },
+            signature: {
+              ed25519: fromHex(
+                "7e899078b35b2d301893b82b0d840f4e1ceb715a86f4d28b2e88312ed605fea83d6b37f7ced4c8023a0d41f79ac9d5556df72c5ec58e26d7db4a6f8c6a537d0a",
+              ),
+            },
+          },
+        ],
+        multisig: [fromHex("8877665544332211")],
```
webmaster128: Question: is 8 bytes the real-workd length to expect?

packages/iov-bns/src/decode.ts
```diff
@@ -471,7 +472,7 @@ function parseSendTransaction(
   return {
     ...base,
     kind: "bcp/send",
-    sender: identityToAddress(base.creator),
+    sender: encodeBnsAddress(prefix, ensure(msg.source, "source")),
```
webmaster128: Nice bugfix

packages/iov-bns/src/decode.spec.ts
willclarktech: Yes, the IDs are sequential and covered by https://github.com/iov-one/weave/blob/22678889d3e1750c21574c340aebf8c7f09a7679/orm/sequence.go#L62-L78

webmaster128: Oh, that's interesting. Then we might even want to encode them numerically in https://github.com/iov-one/iov-core/issues/1216 like we do with proposals.

If they are sequential, we can use `number`s in the safe integer range and throw an error for anything else.

PR #1233, opened 2019-09-13
Test "any account can return an escrow"

webmaster128:


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1104,6 +1111,63 @@ describe("BnsConnection", () => {
       connection.disconnect();
     });
 
+    it("any account can return an escrow", async () => {
```
willclarktech: This description makes it sound like anyone can cancel anyone else's escrow.

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -1104,6 +1111,63 @@ describe("BnsConnection", () => {
...
+
+        const timeout = Math.floor(Date.now() / 1000) + 3;
+
+        const createEscrowTx = await connection.withDefaultFee<CreateEscrowTx & WithCreator>({
+          kind: "bns/create_escrow",
+          creator: sender,
+          sender: senderAddress,
+          arbiter: encodeBnsAddress("tiov", fromHex("0000000000000000000000000000000000000000")),
+          recipient: encodeBnsAddress("tiov", fromHex("0000000000000000000000000000000000000000")),
+          amounts: [defaultAmount], // no tokens
```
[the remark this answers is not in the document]
webmaster128: Thank! I tried and empty array first but the chain rejected it

packages/iov-bns/src/bnsconnection.spec.ts
webmaster128: This is exactly what happens, but only after timeout. Any idea for better wording? I wanted to pick up the verb from the transaction name

willclarktech: Ah yeah, I remember now. Perhaps just "any account can return an escrow (after the timeout)".

webmaster128: This behaviour is of critical importance when we want to use escrows for vesting, which after the lockup period send to a Ledger address. This Ledger address can only sign send transactions, so "a friend" is needed to help out triggering the returns.

PR #1240, opened 2019-09-18
Allow empty arrays in Governor.getElectionRules

### LINKED ISSUE #1239: Don't throw if number of rules for electorate is 0
in

```ts
  public async getElectionRules(electorateId: number): Promise<readonly ElectionRule[]> {
    const electionRules = await this.connection.getElectionRules();
    const filteredRules = electionRules.filter(rule => rule.electorateId === electorateId);

    if (filteredRules.length === 0) {
      throw new Error("No election rule found for electorate");
    }

    const groupedRules = groupByCallback(filteredRules, rule => rule.id);
    return groupedRules.map(group => maxWithComparatorCallback(group.values, compareByVersion));
  }
```

willclarktech:
Closes #1239

### DIFF
--- a/packages/iov-bns-governance/src/governor.spec.ts
+++ b/packages/iov-bns-governance/src/governor.spec.ts
@@ -138,18 +138,14 @@ describe("Governor", () => {
   });
 
   describe("getElectionRules", () => {
-    it("throws for non-existent electorateId", async () => {
+    it("returns an empty array for non-existent electorateId", async () => {
       pendingWithoutBnsd();
       const options = await getGovernorOptions();
       const governor = new Governor(options);
 
       const electorateId = 100;
-      await governor
-        .getElectionRules(electorateId)
-        .then(
-          () => fail("Expected promise to be rejected"),
-          err => expect(err.message).toMatch(/no election rule found/i),
-        );
+      const electionRules = await governor.getElectionRules(electorateId);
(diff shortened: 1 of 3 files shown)


packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -138,18 +138,14 @@ describe("Governor", () => {
   });
 
   describe("getElectionRules", () => {
-    it("throws for non-existent electorateId", async () => {
+    it("returns an empty array for non-existent electorateId", async () => {
```
webmaster128: hah! This test discriptions reveals an interesting differentiation. We have two cases to consider:
1. Electorate exists but has no rules
2. Electorate does not exist

For 2. it is fair to throw I think. However, I don't know if we get the data easily without an additional query to the blockchain. Throughts?

willclarktech: Yeah, so I think the non-existence of an election rule was being used as a proxy for the non-existence of an electorate to avoid an additional request. I think it's unlikely people will be passing the wrong electorate ID here, and if they do the empty array should be a clue that maybe the electorate doesn't exist.

webmaster128: +1
If we consider `electorateId` a filter, it makes perfect sense. And most of all, it fulfils the requirements by the governance app

PR #1247, opened 2019-09-18
Add isFullSignature typechecker to BCP

### LINKED ISSUE #1244: Add isFullSignature type checker
Needed for the Multisignature too.

I guess a good implementation is

```ts
export function isFullSignature(data: unknown): data is FullSignature {
  if (!isNonNullObject(data)) return false;

  const { nonce, pubkey, signature } = data as FullSignature;
  if (typeof nonce !== "number") return false;
  if (!isPubkeyBundle(pubkey)) return false;
  if (!isUint8Array(signature)) return false;

  return true;
}
```

however, I don't know if I am missing something

willclarktech:
Closes #1244

### DIFF
--- a/packages/iov-bcp/src/transactions.spec.ts
+++ b/packages/iov-bcp/src/transactions.spec.ts
@@ -3,13 +3,17 @@ import { Encoding } from "@iov/encoding";
 import {
   Algorithm,
   ChainId,
+  FullSignature,
   Identity,
   isBlockHeightTimeout,
+  isFullSignature,
   isIdentity,
   isPubkeyBundle,
   isTimestampTimeout,
+  Nonce,
   PubkeyBundle,
   PubkeyBytes,
+  SignatureBytes,
 } from "./transactions";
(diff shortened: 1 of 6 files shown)


packages/iov-bcp/src/transactions.spec.ts
```diff
@@ -117,6 +121,66 @@ describe("transactions", () => {
...
+    });
+
+    it("returns false for a bunch of other stuff", () => {
+      expect(isFullSignature("abc")).toEqual(false);
+      expect(isFullSignature({})).toEqual(false);
+      expect(isFullSignature(null)).toEqual(false);
+      expect(isFullSignature(undefined)).toEqual(false);
+      expect(isFullSignature(fromHex("aabb"))).toEqual(false);
+
+      const missingChainId: Omit<Identity, "chainId"> = {
```
webmaster128: Those tests should be ` Omit<FullSignature, "some property">`

PR #1248, opened 2019-09-18
Resolve circular dependencies

### LINKED ISSUE #1245: Circular dependency in @iov/bcp
`atomicswaptypes.ts` imports from `transactions.ts` and vice versa.

willclarktech:
Closes #1245

Note that this might not even be best practice in the case of types. See https://github.com/benmosher/eslint-plugin-import/issues/1453

### DIFF
--- a/packages/iov-bcp/src/atomicswaptypes.ts
+++ b/packages/iov-bcp/src/atomicswaptypes.ts
@@ -2,7 +2,77 @@ import { As } from "type-tagger";
 import { Stream } from "xstream";
 
 import { BlockchainConnection } from "./connection";
-import { Address, Amount, SwapId, SwapTimeout } from "./transactions";
+import { Address, Amount, LightTransaction, SwapId, SwapTimeout } from "./transactions";
+
+export type Preimage = Uint8Array & As<"preimage">;
+export type Hash = Uint8Array & As<"hash">;
+
+// Transactions
+
+/** A swap offer or a counter offer */
+export interface SwapOfferTransaction extends LightTransaction {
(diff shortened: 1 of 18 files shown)


packages/iov-bcp/src/atomicswaptypes.ts
```diff
@@ -22,9 +21,6 @@ export function isSwapProcessStateAborted(state: SwapProcessState): state is Swa
   return state === SwapProcessState.Aborted;
 }
 
-export type Preimage = Uint8Array & As<"preimage">;
-export type Hash = Uint8Array & As<"hash">;
```
webmaster128: Those are needed for `SwapOfferTransaction`/`SwapClaimTransaction`. Doesn't it make more sense to move the atomic swap transaction types into `atomicswaptypes.ts` instead? This should be very easy now that transaction type registraion is decentralized. This was not always the case: `transactions.ts` has some history in which UnsignedTransaction was a union type.

packages/iov-bcp/src/index.ts
```diff
@@ -50,13 +50,5 @@ export {
...
   createTimestampTimeout,
-  SwapOfferTransaction,
-  SwapClaimTransaction,
-  SwapAbortTransaction,
-  SwapTransaction,
   isSendTransaction,
-  isSwapOfferTransaction,
-  isSwapClaimTransaction,
-  isSwapAbortTransaction,
-  isSwapTransaction,
```
webmaster128: Can we have those explors explicitely from `./atomicswaptypes`?

PR #1251, opened 2019-09-19
Add Decimal class

webmaster128:
We are implementing too many of this functionality again and again in caller projects like ponferrada, the Token finder, the multisignature tool, the faucet, including code like

```ts
// This produces a human readable format of the amount, value and token ticker
export function amountToString(amount: Amount): string {
  const { tokenTicker } = amount;
  const value = amountToNumber(amount);

  return `${value} ${tokenTicker}`;
}
```

where we rely on the bahaviour of an IEEE 754 float, which cannot express all amounts precisely.

The basic functionality included here allows converting our IOV-Core amounts to human readable numbers like:

```ts
export function amountToString(amount: Amount): string {
  const { quantity, fractionalDigits, tokenTicker } = amount;
  return `${Decimal.fromAtomics(quantity, fractionalDigits).toString()} ${tokenTicker}`;
}
```

and reading user input into amounts.

### DIFF
--- a/packages/iov-encoding/src/decimal.spec.ts
+++ b/packages/iov-encoding/src/decimal.spec.ts
@@ -85,6 +85,12 @@ describe("Decimal", () => {
       expect(Decimal.fromUserInput("", 2).atomics).toEqual("0");
       expect(Decimal.fromUserInput("", 3).atomics).toEqual("0");
     });
+
+    it("accepts american notation with skipped leading zero", () => {
+      expect(Decimal.fromUserInput(".1", 3).atomics).toEqual("100");
+      expect(Decimal.fromUserInput(".12", 3).atomics).toEqual("120");
+      expect(Decimal.fromUserInput(".123", 3).atomics).toEqual("123");
+    });
   });
 
   describe("toString", () => {
--- a/packages/iov-encoding/src/decimal.spec.ts
+++ b/packages/iov-encoding/src/decimal.spec.ts
@@ -13,6 +13,11 @@ describe("Decimal", () => {
     it("throws for more than one separators", () => {
       expect(() => Decimal.fromUserInput("1.3.5", 5)).toThrowError(/more than one separator found/i);
       expect(() => Decimal.fromUserInput("1..3", 5)).toThrowError(/more than one separator found/i);
+      expect(() => Decimal.fromUserInput("..", 5)).toThrowError(/more than one separator found/i);
+    });
+
+    it("throws for separator only", () => {
+      expect(() => Decimal.fromUserInput(".", 5)).toThrowError(/fractional part missing/i);
     });
 
     it("throws for more fractional digits than supported", () => {
--- a/packages/iov-encoding/src/decimal.spec.ts
+++ b/packages/iov-encoding/src/decimal.spec.ts
@@ -1,0 +1,89 @@
+import { Decimal } from "./decimal";
+
+describe("Decimal", () => {
+  describe("fromUserInput", () => {
+    it("throws helpful error message for invalid characters", () => {
+      expect(() => Decimal.fromUserInput(" 13", 5)).toThrowError(/invalid character at position 1/i);
+      expect(() => Decimal.fromUserInput("1,3", 5)).toThrowError(/invalid character at position 2/i);
+      expect(() => Decimal.fromUserInput("13-", 5)).toThrowError(/invalid character at position 3/i);
+      expect(() => Decimal.fromUserInput("13/", 5)).toThrowError(/invalid character at position 3/i);
+      expect(() => Decimal.fromUserInput("13\\", 5)).toThrowError(/invalid character at position 3/i);
+    });
+
+    it("throws for more than one separators", () => {
+      expect(() => Decimal.fromUserInput("1.3.5", 5)).toThrowError(/more than one separator found/i);
+      expect(() => Decimal.fromUserInput("1..3", 5)).toThrowError(/more than one separator found/i);
+    });
+
+    it("throws for more fractional digits than supported", () => {
+      expect(() => Decimal.fromUserInput("44.123456", 5)).toThrowError(
+        /got more fractional digits than supported/i,
+      );
+      expect(() => Decimal.fromUserInput("44.1", 0)).toThrowError(
+        /got more fractional digits than supported/i,
+      );
+    });
+
+    it("throws for fractional digits that are not non-negative integers", () => {
+      // no integer
+      expect(() => Decimal.fromUserInput("1", Number.NaN)).toThrowError(
+        /fractional digits is not an integer/i,
+      );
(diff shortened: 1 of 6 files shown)


packages/iov-encoding/src/decimal.spec.ts
```diff
@@ -0,0 +1,126 @@
+import { Decimal } from "./decimal";
+
+describe("Decimal", () => {
+  describe("fromUserInput", () => {
+    it("throws helpful error message for invalid characters", () => {
+      expect(() => Decimal.fromUserInput(" 13", 5)).toThrowError(/invalid character at position 1/i);
```
willclarktech: Definitely positions starting from 1 not 0?

packages/iov-encoding/src/decimal.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      expect(() => Decimal.fromUserInput("1,3", 5)).toThrowError(/invalid character at position 2/i);
+      expect(() => Decimal.fromUserInput("13-", 5)).toThrowError(/invalid character at position 3/i);
+      expect(() => Decimal.fromUserInput("13/", 5)).toThrowError(/invalid character at position 3/i);
+      expect(() => Decimal.fromUserInput("13\\", 5)).toThrowError(/invalid character at position 3/i);
+    });
+
+    it("throws for more than one separators", () => {
```
willclarktech: "one separator"

packages/iov-encoding/src/decimal.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    it("cuts tailing zeros", () => {
+      expect(Decimal.fromUserInput("4.12", 5).atomics).toEqual("412000");
+      expect(Decimal.fromUserInput("4.120", 5).atomics).toEqual("412000");
+      expect(Decimal.fromUserInput("4.1200", 5).atomics).toEqual("412000");
+      expect(Decimal.fromUserInput("4.12000", 5).atomics).toEqual("412000");
+      expect(Decimal.fromUserInput("4.120000", 5).atomics).toEqual("412000");
+      expect(Decimal.fromUserInput("4.1200000", 5).atomics).toEqual("412000");
+    });
+
+    it("interpretes the empty string as zero", () => {
```
willclarktech: "interprets"

packages/iov-encoding/src/decimal.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    });
+
+    it("only shows significant digits", () => {
+      expect(Decimal.fromUserInput("44.1", 2).toString()).toEqual("44.1");
+      expect(Decimal.fromUserInput("44.10", 2).toString()).toEqual("44.1");
+      expect(Decimal.fromUserInput("44.100", 2).toString()).toEqual("44.1");
+    });
+
+    it("fills up leading zeros", () => {
+      expect(Decimal.fromAtomics("3", 0).toString()).toEqual("3");
```
willclarktech: Tests for `fromAtomics`?

packages/iov-encoding/src/decimal.ts
```diff
@@ -0,0 +1,86 @@
...
+ * A type for arbitrary precision, non-negative decimals.
+ *
+ * Instances of this class are immutable.
+ */
+export class Decimal {
+  public static fromUserInput(input: string, fractionalDigits: number): Decimal {
+    if (!Number.isInteger(fractionalDigits)) throw new Error("Fractional digits is not an integer");
+    if (fractionalDigits < 0) throw new Error("Fractional digits must not be negative");
+
+    for (let index = 0; index < input.length; index++) {
```
willclarktech: It would be cleaner and faster to use a negated set:
```ts
const badCharacter = input.match(/[^\d\.]/);
if (badCharacter) {
    throw new Error(`Invalid character at position ${badCharacter.index + 1}`);
}
```

packages/iov-encoding/src/decimal.ts
```diff
@@ -0,0 +1,86 @@
...
+        default:
+          throw new Error("More than one separator found");
+      }
+    }
+
+    if (fractional.length > fractionalDigits) {
+      throw new Error("Got more fractional digits than supported");
+    }
+
+    const quantity = `${whole}${fractional.padEnd(fractionalDigits, "0")}`;
```
willclarktech: Didn't know about these string methods! Thanks.

packages/iov-encoding/src/decimal.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (input.search(/\./) === -1) {
+      // integer format, no separator
+      whole = input;
+      fractional = "";
+    } else {
+      const parts = input.split(".");
+      switch (parts.length) {
+        case 0:
+        case 1:
+          throw new Error("Less than two elements in split result. This must not happen here.");
```
willclarktech: "Fewer" 😛

packages/iov-encoding/src/decimal.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    let whole: string;
+    let fractional: string;
+
+    if (input.search(/\./) === -1) {
+      // integer format, no separator
+      whole = input;
+      fractional = "";
+    } else {
+      const parts = input.split(".");
```
willclarktech: You could avoid the `if`/`else` with something like this:
```ts
const parts = input.split(".");
if (parts.length === 0 || parts.length > 2) throw new Error("...");
const [whole, rawFractional = ""] = parts;
const fractional = rawFractional.replace(/0+$/, "");
```

packages/iov-encoding/src/decimal.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    const quantity = `${whole}${fractional.padEnd(fractionalDigits, "0")}`;
+
+    return new Decimal(quantity, fractionalDigits);
+  }
+
+  public static fromAtomics(atomics: string, fractionalDigits: number): Decimal {
+    if (!Number.isInteger(fractionalDigits)) throw new Error("Fractional digits is not an integer");
+    if (fractionalDigits < 0) throw new Error("Fractional digits must not be negative");
```
willclarktech: Consider a separate validation function for these lines.

packages/iov-encoding/src/decimal.spec.ts
webmaster128: The idea was to allow propagating this error message into the user interface, where 1-based positions might be more intuitive. However, I don't have string opinions on this one.

willclarktech: Me neither, that sounds sensible.

packages/iov-encoding/src/decimal.ts
webmaster128: I used this code befor. However, this made "." a valid input, which I wanted to avoid. Then I decided to write it a little more explicit.

webmaster128:
Okay, I think I addressed all feedback (starting at commit "Fix language")

Issue #1220, opened 2019-09-02
New chain connector: Cosmos


kmw101:
Add a new chain connector to support Cosmos.

MVP: support Cosmos Hub and ATOM (we can generalise to other chains/tokens at a later point). We should be able to reuse our Tendermint code.

- [x] Research which transaction types we want to support
- [x] Test setup of Cosmos Hub node (https://hub.cosmos.network/docs/deploy-testnet.html)
- [x] Start Cosmos node in Travis
- [x] @iov/cosmos package setup
- [x] Encoding/decoding
- [x] Codec
- [x] Connection [MVP]
- [x] Connector

willclarktech:
Based on modules listed here: https://github.com/cosmos/gaia/blob/master/docs/what-is-gaia.md

Transactions to include:
- `bank/send`: https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/internal/types/msgs.go#L10-L15

Transactions to exclude:
- `bank/multisend`: https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/internal/types/msgs.go#L57-L61
- `staking/*`: https://github.com/cosmos/cosmos-sdk/blob/master/x/staking/handler.go#L16-L41
- `distribution/*`: https://github.com/cosmos/cosmos-sdk/blob/master/x/distribution/handler.go#L12-L31
- `slashing/*`: https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/handler.go#L10-L24
- `gov/*`: https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/handler.go#L10-L30

willclarktech:
@webmaster128 Under which version should CHANGELOG entries appear?

webmaster128:
> @webmaster128 Under which version should CHANGELOG entries appear?

I suggest collecting them at an extra place (here, wiki, private notes, Confluence, gist) and add them to the first release that publishes a @iov/cosmos package.

willclarktech:
@webmaster128 OK to close this issue and open follow-ups for improvements eg RPC support?

webmaster128:
👍 We just need to create an overview of what is missing before we can un-private @iov/cosmos. Using the chain IDs from https://github.com/ChainAgnostic/CAIPs/issues/5 is one open thing. Balance subscriptions would be needed for wallet integration.

PR #1252, opened 2019-09-19
Calculate BlockId from Tendermint header

### LINKED ISSUE #618: Calculate BlockId from Tendermint Header
A BCP BlockId for Tendermint-based chains is an uppercase hex representation of a Tendermint [block hash](https://github.com/tendermint/tendermint/wiki/Block-Structure#block-hash).

Since we currently have no way to calculate the block hash from a block header, we need to do [an extra request](https://github.com/iov-one/iov-core/blob/v0.10.1/packages/iov-bns/src/bnsconnection.ts#L571-L572) for every event when watching block headers. This is something we want to avoid.

This is how to calculate:

* Check the [BlockHeader](https://github.com/tendermint/tendermint/blob/v0.27.0/types/block.go#L349-L380) structure
* [Hash the header](https://github.com/tendermint/tendermint/blob/v0.27.0/types/block.go#L404-L432)
* Using [cdcEncode](https://github.com/tendermint/tendermint/blob/v0.27.0/types/encoding_helper.go) and [merkle.SimpleHashFromByteSlices](https://github.com/tendermint/tendermint/blob/v0.27.0/crypto/merkle/simple_tree.go#L21-L34)
* "cbc" means codec and is implemented by [go-amino](https://github.com/tendermint/go-amino)

willclarktech:
Closes #618

Draft PR to check a few design decisions:
- [x] `ReadonlyDateWithNanoseconds` approach (awkward but fairly simple and avoids breaking change)
- [x] Location of encoding functions
- [x] Assuming small numbers for eg hash lengths so we don’t have to do a full implementation of the `VarBuffer`s etc

Still to do:
- [x] Test cases for encoding functions
- [x] Test case for block with transactions
- [x] Explanation of where certain magic numbers come from/other documentation links
- [x] Attribution to nomic-io for inspiration behind many functions?
- [x] Actually use the function in bnsconnection.ts
- [x] ~~Copy and paste to v0.29~~ [Just use dummy function since it's going to be removed anyway]
- [x] CHANGELOG entry

### DIFF
--- a/packages/iov-tendermint-rpc/src/encodings.ts
+++ b/packages/iov-tendermint-rpc/src/encodings.ts
@@ -1,6 +1,9 @@
 import { Encoding, Int53 } from "@iov/encoding";
-import { ReadonlyDate } from "readonly-date";
 import { As } from "type-tagger";
+
+import { BlockId, ReadonlyDateWithNanoseconds, Version } from "./responses";
+
+const { toUtf8 } = Encoding;
 
 export type Base64String = string & As<"base64">;
 export type HexString = string & As<"hex">;
@@ -156,8 +159,13 @@ export class Base64 {
 }
 
 export class DateTime {
-  public static decode(dateTimeString: DateTimeString): ReadonlyDate {
-    return Encoding.fromRfc3339(dateTimeString);
+  public static decode(dateTimeString: DateTimeString): ReadonlyDateWithNanoseconds {
+    const readonlyDate = Encoding.fromRfc3339(dateTimeString);
+    const nanosecondsMatch = dateTimeString.match(/\.(\d+)Z$/);
+    const nanoseconds = nanosecondsMatch ? nanosecondsMatch[1].slice(3) : "";
+    // tslint:disable-next-line:no-object-mutation
+    (readonlyDate as any).nanoseconds = parseInt(nanoseconds.padEnd(6, "0"), 10);
+    return readonlyDate as ReadonlyDateWithNanoseconds;
   }
 }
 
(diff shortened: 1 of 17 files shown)


packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -156,8 +157,13 @@ export class Base64 {
 }
 
 export class DateTime {
-  public static decode(dateTimeString: DateTimeString): ReadonlyDate {
-    return Encoding.fromRfc3339(dateTimeString);
+  public static decode(dateTimeString: DateTimeString): ReadonlyDateWithNanoseconds {
```
webmaster128: This new type is brilliant in my opinion

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -156,8 +157,13 @@ export class Base64 {
...
+    const readonlyDate = Encoding.fromRfc3339(dateTimeString);
+    const nanosecondsMatch = dateTimeString.match(/\.(\d+)Z$/);
+    const nanoseconds = nanosecondsMatch ? nanosecondsMatch[1] : "";
+    // tslint:disable-next-line:no-object-mutation
+    (readonlyDate as any).nanoseconds = parseInt(nanoseconds.padEnd(9, "0"), 10);
+    return readonlyDate as ReadonlyDateWithNanoseconds;
```
webmaster128: If JavaScript's _everything is an object mentality_ is true, then this code should do the same:

```ts
    const readonlyDate = Encoding.fromRfc3339(dateTimeString);
    const nanosecondsMatch = dateTimeString.match(/\.(\d+)Z$/);
    const nanoseconds = nanosecondsMatch ? nanosecondsMatch[1] : "";
    return {
      ...readonlyDate,
      nanoseconds: parseInt(nanoseconds.padEnd(9, "0"), 10),
    };
```

At least TypeScript is happy. Not runtime tested.

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -156,8 +157,13 @@ export class Base64 {
...
 
 export class DateTime {
-  public static decode(dateTimeString: DateTimeString): ReadonlyDate {
-    return Encoding.fromRfc3339(dateTimeString);
+  public static decode(dateTimeString: DateTimeString): ReadonlyDateWithNanoseconds {
+    const readonlyDate = Encoding.fromRfc3339(dateTimeString);
+    const nanosecondsMatch = dateTimeString.match(/\.(\d+)Z$/);
+    const nanoseconds = nanosecondsMatch ? nanosecondsMatch[1] : "";
+    // tslint:disable-next-line:no-object-mutation
+    (readonlyDate as any).nanoseconds = parseInt(nanoseconds.padEnd(9, "0"), 10);
```
webmaster128: This duplicates data (fractional digits 1-3). In

```
    // fractional seconds match either undefined or a string like ".1", ".123456789"
    const milliSeconds = matches[7] ? Math.floor(+matches[7] * 1000) : 0;
```

position 4-9 are truncated. So I wonder if we want to only store position 4-9 in the extra field, to be interpreted as _nanoseconds after the value stored in readonlyDate_

packages/iov-tendermint-rpc/src/v0-31/hasher.spec.ts
```diff
@@ -1,14 +1,59 @@
...
     expect(hashTx(txData)).toEqual(txId);
   });
+
+  it("creates block hash equal to local test for empty block", () => {
+    // This was taken from a result from /block of some random empty block
+    // curl "http://localhost:11131/block"
+    const blockId = fromHex("5B5D3F7E77A4BD6CB6067947E478BC3BD493DD24A981535F0ADEBDAAA0498480");
+    const time = new ReadonlyDate("2019-09-19T10:41:24.898178746Z");
+    // tslint:disable-next-line:no-object-mutation
+    (time as any).nanoseconds = 898178746;
```
webmaster128: Untested but makes TS happy: 

```ts
const time: ReadonlyDateWithNanoseconds = {
  ...new ReadonlyDate("2019-09-19T10:41:24.898178746Z"),
  nanoseconds: 898178746,
};
```

willclarktech:
@webmaster128 Any thoughts on where the encoding functions should go and/or skipping the full `VarBuffer` implementation?

webmaster128:
> Any thoughts on where the encoding functions should go 

I'd assume the encoding is 1. Tendermint specific and 2. the same for all supported versions of Tendermint. So what about `packages/iov-tendermint-rpc/src/encoding`?

> and/or skipping the full VarBuffer implementation?

Let's do the minimal implementation necessary for block headers. [At one point in the not too far furture](https://github.com/iov-one/iov-core/issues/1220), we will need a full amino implementation, but that can be done later on.

packages/iov-tendermint-rpc/src/v0-31/hasher.spec.ts
willclarktech: This doesn't work because you lose the `Date` methods like `getTime`.

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -170,3 +178,55 @@ export class Hex {
...
+  const secondsArray = seconds ? [0x08, ...encodeInt(seconds)] : new Uint8Array();
+  const nanoseconds = (time.nanoseconds || 0) + (milliseconds % 1000) * 1e6;
+  const nanosecondsArray = nanoseconds ? [0x10, ...encodeInt(nanoseconds)] : new Uint8Array();
+  return Uint8Array.from([...secondsArray, ...nanosecondsArray]);
+}
+
+// See https://github.com/tendermint/go-amino/blob/v0.15.0/encoder.go#L180-L187
+export function encodeHash(hash: Uint8Array): Uint8Array {
+  // Since we're only dealing with short hashes we don't need a full VarBuffer implementation yet
+  return hash.length ? Uint8Array.from([hash.length, ...hash]) : hash;
```
webmaster128: Since `Uint8Array`s are mutable, I'd avoid returning the input object directly. I did not find a way to mess around with this in this case since it only works for the empty array, but in general it would be unexpected if the input data changes when manipulating the encoded data.

packages/iov-tendermint-rpc/src/encodings.ts
```diff
@@ -170,3 +178,49 @@ export class Hex {
...
+}
+
+export function encodeTime(time: ReadonlyDateWithNanoseconds): Uint8Array {
+  const milliseconds = time.getTime();
+  const seconds = Math.floor(milliseconds / 1000);
+  const nanoseconds = (time.nanoseconds || 0) + (milliseconds % 1000) * 1e6;
+  return Uint8Array.from([0x08, ...encodeInt(seconds), 0x10, ...encodeInt(nanoseconds)]);
+}
+
+export function encodeHash(hash: Uint8Array): Uint8Array {
```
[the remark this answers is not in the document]
willclarktech: It should be generic. I'll update.

PR #1257, opened 2019-09-24
Change return type of toBytesBigEndian/toBytesLittleEndian to Uint8Array

### LINKED ISSUE #1250: Change return type of toBytesBigEndian/toBytesLittleEndian to Uint8Array
Uint8Array is our primary type for binary data. Right now I need to write code like

```ts
Uint8Array.from(multisigId.toBytesBigEndian())
```

when we use the byte result. The idea of using `number[]` came from bn.js but is not very consistent for our wrapper (and uses at least 8 bytes of memory per byte).

This affects

```
Uint32.toBytesBigEndian
Uint64.toBytesBigEndian
Uint64.toBytesLittleEndian
```

webmaster128:
Closes #1250


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 1.0.0
+
+- @iov/encoding: Add `Uint32.toBytesLittleEndian`.
+
+Breaking changes
+
+- @iov/encoding: Change return type of `Uint32.toBytesBigEndian`,
+  `.toBytesLittleEndian`, `Uint64.toBytesBigEndian` and `.toBytesLittleEndian`
```
willclarktech: `Uint32.toBytesLittleEndian` doesn't need to be included here if it's new.

webmaster128:
Updated

PR #1267, opened 2019-09-25
Make escrow IDs numeric

### LINKED ISSUE #1235: Change type of escrowId to number
Like in multisig contracts, those are sequences that can be represented as numbers (uint64).

~Non-breaking since escrow types are not exported.~

- [x] ReleaseEscrowTx, ReturnEscrowTx, UpdateEscrowPartiesTx
- [x] ReleaseEscrowAction
- [ ] escrowIdToAddress

willclarktech:
Closes #1235


packages/iov-bns-governance/src/governor.spec.ts
```diff
@@ -23,7 +23,7 @@ function pendingWithoutBnsd(): void {
 const adminMnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
 const adminPath = HdPaths.iov(0);
 const bnsdUrl = "ws://localhost:23456";
-const guaranteeFundEscrowId = Encoding.fromHex("88008800");
+const guaranteeFundEscrowId = 2281736192;
```
webmaster128: Is this even used anywhere? Looks like a very very big number given this is a counter

packages/iov-bns-governance/src/integrationtests/proposals.spec.ts
```diff
@@ -36,8 +36,11 @@ import { CommitteeId, ProposalOptions, ProposalType } from "../proposals";
 const adminMnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
 const adminPath = HdPaths.iov(0);
 const bnsdUrl = "ws://localhost:23456";
-const guaranteeFundEscrowId = Encoding.fromHex("0000000000000001");
-const guaranteeFundAddress = escrowIdToAddress("local-iov-devnet" as ChainId, guaranteeFundEscrowId);
+const guaranteeFundEscrowId = 1;
+const guaranteeFundAddress = escrowIdToAddress(
+  "local-iov-devnet" as ChainId,
+  Encoding.fromHex("0000000000000001"),
+);
```
webmaster128: Can we change the input type of escrowIdToAddress too please? I added this a couple of days ago to the ticket

packages/iov-bns/src/decode.ts
```diff
@@ -97,7 +97,7 @@ export function ensure<T>(maybe: T | null | undefined, msg?: string): T {
   return maybe;
 }
 
-function decodeNumericId(id: Uint8Array): number {
+export function decodeNumericId(id: Uint8Array): number {
```
webmaster128: For escrow IDs I suggest using an implementation that ensures 8 byte inputs, like `Uint64.fromBytesBigEndian(id).toNumber()`. Maybe we want a dedicated function for that, depending on how `decodeNumericId` is used

packages/iov-bns-governance/src/governor.spec.ts
willclarktech: It's used in the test for `ReleaseGuaranteeFunds`. I can change it to a more plausible number.

packages/iov-bns/src/decode.ts
webmaster128: Did you delete your last comment?

Versions are 4 byte values as far as I know

willclarktech: Ah sorry, not versions, but versioned IDs.

PR #1274, opened 2019-09-26
Add setup for Cosmos test network

willclarktech:
Part of #1220


scripts/cosmos/cli.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+command -v shellcheck > /dev/null && shellcheck "$0"
+
+# Choose from https://hub.docker.com/r/tendermint/gaia/tags
+VERSION="v2.0.0"
+CURRENT_DIR="$(realpath "$(dirname "$0")")"
+GAIAD_CONTAINER_NAME="gaiad"
+
+docker run \
```
webmaster128: We typically set `--user="$UID"` to `docker run` commands in order to make sure the files written can be deleted by the current user. I don't now if that is relevant for mac, where the docker deamon does not run as root. But for Linux it avoids issues.

PR #1279, opened 2019-10-01
Improve create proposal tests

webmaster128:


packages/iov-bns-governance/src/integrationtests/proposals.spec.ts
```diff
@@ -482,14 +482,18 @@ describe("Proposals", () => {
     const governor = new Governor(governorOptions);
     const numProposalsBefore = (await governor.getProposals()).length;
     const guaranteeFundBeforeRelease = await connection.getAccount({ address: guaranteeFundAddress });
-    const guaranteeFundQuantityBeforeRelease = getCashQuantity(guaranteeFundBeforeRelease!);
+    const guaranteeFundQuantityBeforeRelease = getCashQuantity(guaranteeFundBeforeRelease);
+
+    const rewardFundInitial = await connection.getAccount({ address: rewardFundAddress });
+    const rewardFundInitialQuantity = getCashQuantity(rewardFundInitial);
+    expect(rewardFundInitialQuantity.toString()).toEqual("0"); // if this is not 0, the remainder test below will not work
```
willclarktech: Maybe just throw because this is an assumption of the test rather than something we're testing.

webmaster128: Yeah I agree. I think the next iteration should allow this test to pass multiple times in a row. But this is no priority. Merging for now

PR #1281, opened 2019-10-02
Add @iov/cosmos package setup

willclarktech:
Part of #1220

### DIFF
--- a/packages/iov-cosmos/package.json
+++ b/packages/iov-cosmos/package.json
@@ -24,7 +24,7 @@
     "docs": "shx rm -rf docs && typedoc --options typedoc.js",
     "format": "prettier --write --loglevel warn \"./src/**/*.ts\"",
     "lint": "eslint --max-warnings 0 \"**/*.{js,ts}\" && tslint -t verbose --project .",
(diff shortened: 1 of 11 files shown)


packages/iov-cosmos/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "*.md",
+    "!*.spec.*",
+    "!**/testdata/"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/iov-one/iov-core/tree/master/packages/iov-cosmos"
+  },
+  "publishConfig": {
+    "access": "public"
```
webmaster128: Can you set this to private (which should ensure it is not published)?

webmaster128: I do not yet understand the difference between this setting and the main `"private": true`. See https://github.com/lerna/lerna/issues/1942#issuecomment-468001574

Maybe you can have a look into this

willclarktech: The only valid values for `publishConfig.access` are `public` and `restricted`, neither of which prevents publishing. What we want is `private` set to `true`.

PR #1284, opened 2019-10-29
Implement Cosmos codec etc

willclarktech:
Part 1 of #1220 

- encoding/decoding
- address conversion
- codec
- connector
- ripemd160 hash
- secp256k1 pubkey compression/signature recovery byte trimming
- REST server

Not covered by this PR:
- connection implementation (see #1285)


packages/iov-cosmos/src/cosmoscodec.ts
```diff
@@ -0,0 +1,96 @@
...
+function sortJson(json: any): any {
+  if (typeof json !== "object" || json === null) {
+    return json;
+  }
+  if (Array.isArray(json)) {
+    return (json as readonly any[]).map(sortJson);
+  }
+  const sortedKeys = Object.keys(json).sort();
+  const result = sortedKeys.reduce(
+    (acc, k) => ({
```
webmaster128: Can we afford a few more characters for `accumulator?` and `key?`? I think it saves a good amount of brain power when reading the code

packages/iov-cosmos/src/cosmoscodec.ts
```diff
@@ -0,0 +1,96 @@
...
+    const prefix = "cosmos";
+    return pubkeyToAddress(identity.pubkey, prefix);
+  }
+
+  public isValidAddress(address: string): boolean {
+    return isValidAddress(address);
+  }
+}
+
+export const cosmosCodec = new CosmosCodec();
```
webmaster128: 👍 The codec instance is definitely better than the original way using a constant object export. We should have the same setup for BNS as well where we could use a class instance to configure testnet vs. mainnet.

packages/iov-cosmos/src/cosmoscodec.ts
```diff
@@ -0,0 +1,96 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
webmaster128: why is this required?

packages/iov-cosmos/src/decode.ts
```diff
@@ -0,0 +1,121 @@
...
+  return {
+    nonce: nonce as Nonce,
+    pubkey: decodePubkey(signature.pub_key),
+    signature: decodeSignature(signature.signature),
+  };
+}
+
+export function decodeAmount(amount: amino.Coin): Amount {
+  return {
+    fractionalDigits: 9,
```
webmaster128: Is this correct? If I understand correctly, Atom has 6 decimal digits (https://github.com/UCRegistry/registry/issues/14)

packages/iov-cosmos/src/decode.ts
```diff
@@ -0,0 +1,121 @@
...
+    quantity: amount.amount,
+    tokenTicker: amount.denom as TokenTicker,
+  };
+}
+
+export function parseMsg(msgs: readonly amino.Msg[]): SendTransaction {
```
webmaster128: Why does this function accept an array? We should have a length check at least of we want to process 1-element arrays.

packages/iov-cosmos/src/decode.ts
```diff
@@ -0,0 +1,121 @@
...
+    sender: msgValue.from_address as Address,
+    recipient: msgValue.to_address as Address,
+    amount: decodeAmount(msgValue.amount[0]),
+  };
+}
+
+export function parseFee(fee: amino.StdFee): Fee {
+  return {
+    tokens: {
+      fractionalDigits: 9,
```
webmaster128: as above

packages/iov-cosmos/src/encode.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+  const defaultCreator: Identity = {
+    chainId: "not-used" as ChainId,
+    pubkey: defaultPubkey,
+  };
+  const defaultSender = "cosmos1h806c7khnvmjlywdrkdgk2vrayy2mmvf9rxk2r" as Address;
+  const defaultRecipient = "cosmos1z7g5w84ynmjyg0kqpahdjqpj7yq34v3suckp0e" as Address;
+  const defaultAmount: Amount = {
+    quantity: "11657995",
+    fractionalDigits: 9,
+    tokenTicker: "vatom" as TokenTicker,
```
webmaster128: Why do we use uatom and vatom?

packages/iov-cosmos/tsconfig.json
```diff
@@ -4,9 +4,8 @@
     "baseUrl": ".",
     "outDir": "build",
     "declarationDir": "build/types",
-    "rootDir": "src"
+    "rootDir": "src",
+    "lib": ["es6", "dom"]
```
webmaster128: "lib" "dom" means that we're telling the TS compiler the dom API is available in the target JS engine. This is only the case for browsers. Do we really need this option?

packages/iov-crypto/src/index.ts
```diff
@@ -1,9 +1,11 @@
 export * from "./bip39";
 export { EnglishMnemonic } from "./englishmnemonic";
+export * from "./hash";
```
webmaster128: Can we make all new export types explicit?

packages/iov-crypto/src/secp256k1.spec.ts
```diff
@@ -556,4 +556,42 @@ describe("Secp256k1", () => {
...
+      expect(Secp256k1.compressPubkey(pubkey)).toEqual(pubkey);
+    });
+
+    it("compresses an uncompressed pubkey", () => {
+      // Test data generated at https://iancoleman.io/bitcoin-key-compression/
+      const pubkey = fromHex(
+        "044f04181eeba35391b858633a765c4a0c189697b40d216354d50890d350c7029013b587a681e836cc187a8164b98a5848a2b89b3173315fdd0740d5032e259cd5",
+      );
+      const compressed = "034f04181eeba35391b858633a765c4a0c189697b40d216354d50890d350c70290";
+      expect(toHex(Secp256k1.compressPubkey(pubkey))).toEqual(compressed);
```
webmaster128: Can we do this comparison on the binary data directly instad of the hex representation?

packages/iov-crypto/src/ripemd.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+
+describe("Ripemd160", () => {
+  it("exists", () => {
+    expect(Ripemd160).toBeTruthy();
+  });
+
+  it("works for empty input", () => {
+    {
+      const hash = new Ripemd160(new Uint8Array([])).digest();
+      expect(toHex(hash)).toEqual("9c1185a5c5e9fc54612808977ee8f548b2258d31");
```
webmaster128: Can we do this comparison on the binary data directly instad of the hex representation?

packages/iov-cosmos/src/cosmoscodec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { buildSignedTx, buildUnsignedTx } from "./encode";
+
+const { toHex, toUtf8 } = Encoding;
+
+function sortJson(json: any): any {
+  if (typeof json !== "object" || json === null) {
+    return json;
+  }
+  if (Array.isArray(json)) {
+    return (json as readonly any[]).map(sortJson);
```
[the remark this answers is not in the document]
willclarktech: I think this is left over from when I tried to make the types for the function more explicit. I'll take it out.

packages/iov-cosmos/src/cosmoscodec.ts
willclarktech: See `bytesToSign`: amino needs camelcase.

packages/iov-cosmos/src/decode.ts
willclarktech: Is there a source for that? I put this in as a placeholder because I didn't find a definitive reference.

packages/iov-cosmos/src/encode.spec.ts
willclarktech: I set up our local test chain with "vatom" to make it clear it wasn't mainnet, and "uatom" comes from real life transactions from the CosmosHub chain that are used in some test cases. I didn't think too rigorously about when to use which.

packages/iov-cosmos/tsconfig.json
willclarktech: I added this because amino-js seems to require it. Maybe we can define some global instead?
```
../../node_modules/@tendermint/amino-js/src/util.ts:21:49 - error TS2304: Cannot find name 'window'.

21     ({ atob, btoa, TextDecoder, TextEncoder } = window);
```

packages/iov-crypto/src/index.ts
willclarktech: I made them all explicit since I was there anyway.

packages/iov-cosmos/tsconfig.json
webmaster128: That's a bug in the library. Or the library is browser-only.

Anyways, why are we even compiling an external .ts file? We should only consume .js and .d.ts from dependencies.

packages/iov-cosmos/src/decode.ts
willclarktech: I found an official blog post: https://blog.cosmos.network/phase-ii-initiated-cosmos-atom-transfers-enabled-by-governance-831a7e555ab6#edeb

webmaster128: I found this:

> The smallest unit of ATOM is the uATOM, or the micro ATOM, where 1 atom = 1,000,000 uATOM.

https://blog.cosmos.network/phase-ii-initiated-cosmos-atom-transfers-enabled-by-governance-831a7e555ab6

This indicates that the u is an ASCII approximation of [µ](https://de.wiktionary.org/wiki/%C2%B5).

packages/iov-cosmos/tsconfig.json
willclarktech: Issue: https://github.com/cosmos/amino-js/issues/4

willclarktech:
Blocked by #1300

PR #1285, opened 2019-10-29
Implement Cosmos connection

willclarktech:
Part 2 of #1220 
Merge after #1284 

- connection implementation
- very basic REST client (RPC requires Tendermint v0.32)
- Cosmos HD paths

Connection methods not implemented:
- anything requiring streams
- `getToken`/`getAllTokens`: no obvious (documented) way to query for these via the REST API

### DIFF
--- a/packages/iov-cosmos/src/cosmosconnection.ts
+++ b/packages/iov-cosmos/src/cosmosconnection.ts
@@ -1,19 +1,23 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/camelcase,@typescript-eslint/no-unused-vars */
 import {
   Account,
   AccountQuery,
   AddressQuery,
+  Algorithm,
   BlockchainConnection,
   BlockHeader,
+  BlockId,
   ChainId,
   ConfirmedAndSignedTransaction,
   ConfirmedTransaction,
   FailedTransaction,
   Fee,
+  isPubkeyQuery,
   LightTransaction,
   Nonce,
   PostableBytes,
   PostTxResponse,
+  PubkeyBytes,
   PubkeyQuery,
   Token,
   TokenTicker,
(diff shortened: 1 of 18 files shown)


willclarktech:
CI is currently failing in browsers because CORS is not enabled on the Cosmos Gaia REST server. See https://github.com/cosmos/cosmos-sdk/issues/4622

webmaster128:
@willclarktech can you please update this PR?

packages/iov-cosmos/src/cosmosconnection.ts
```diff
@@ -75,34 +167,79 @@ export class CosmosConnection implements BlockchainConnection {
...
+      throw new Error("Received transaction of unsupported kind.");
+    }
+    return {
+      tokens: {
+        fractionalDigits: 6,
+        quantity: "5000",
+        tokenTicker: atom,
+      },
+      gasLimit: "200000",
+    };
```
webmaster128: Do we have any documentation why those values are hardcoded?

packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -94,6 +94,10 @@ export class HdPaths {
     return HdPaths.bip44(HdPaths.coinTypes.eth, 0, 0, account);
   }
 
+  public static cosmos(account: number): readonly Slip10RawIndex[] {
+    return HdPaths.bip44(HdPaths.coinTypes.atom, 0, 0, account);
```
webmaster128: Do we have a documentation link for this path? My usual source (https://github.com/trezor/trezor-firmware/blob/master/core/docs/misc/coins-bip44-paths.md) does not list Cosmos

willclarktech: - Coin type is listed here: https://github.com/satoshilabs/slips/blob/master/slip-0044.md
- Account we're choosing to set to 0 for simplicity.
- Change address doesn't apply (so 0).
- `account` is really address index.

See https://github.com/cosmos/cosmos-sdk/blob/2c96bbb/crypto/keys/types.go#L34-L36

webmaster128: That cosmos SDK link is very helpful. There however, a two dimentional space is create (`account` and `index` but let's ignore the naming of the axes for now). In https://github.com/cosmos/cosmos-sdk/blob/v0.37.4/client/keys/add.go#L39-L259 it looks like both dimensions are used.

It would be interesting to find out what wallets typically do.

willclarktech: - Cosmostation uses the same: https://github.com/cosmostation/cosmosjs/blob/master/src/index.js#L20
- As does Lunie: https://github.com/luniehq/cosmos-keys/blob/7fe5bb4b85d40b6272cfccc12f1ff3a5f0ee380e/src/cosmos-keys.ts#L8

packages/iov-cosmos/src/cosmosconnection.ts
willclarktech: I can't actually remember where I got these from. I see it in an example here: https://github.com/cosmostation/cosmosjs#usage

I don't know what counts as a sensible default, and taking a brief look at a block explorer it seems like there are wildly varying values used on mainnet.

webmaster128: 👍 we can adapt this on demand. The worst that can happen is overpay or tx rejections.

packages/iov-keycontrol/src/hdpaths.ts
webmaster128: Finally a confirmation here! https://github.com/cosmos/cosmos-sdk/issues/4278#issuecomment-561238038

PR #1288, opened 2019-11-06
Rename ASH to MASH in local bnsd config

### LINKED ISSUE #1287: Rename ASH token in local BNSd config
ASH and TRASH are our ERC20 test tokens. At the same time the BNSd genesis defines an ASH token on the local BNS chain. This becomes a problem then the chain setup scripts are used in the [wallet repo](https://github.com/iov-one/ponferrada/tree/master/scripts), where conflicting token tickers are not supported.

Since renaming the (mostly unused) ASH in bnsd is far easier than renaming the ASH contract, this is the preferred option.

webmaster128:
to avoid naming conflict with local ERC20 test token

Closes  #1287


webmaster128:
Weave does only support up to 4 letters as the ticker symbol :( I'll come up with something different.

Issue #1289, opened 2019-11-11
Implement mutable product fees


webmaster128:
Requirements:
https://iov.atlassian.net/wiki/spaces/PRODUCT/pages/1081349/Mutable+product+fees+-+product+specification

In https://github.com/iov-one/weave/pull/992 (released as part of Weave 0.21.3) the message fees become mutable.

In IOV Core we should

- [x] Update codec to 0.21.3
- [x] Update local bnsd chain to 0.21.3
- [x] Update local test genesis to set an `app_state.conf.msgfee.fee_admin` and `.owner` (see https://github.com/iov-one/weave/pull/992/files#diff-af7902b005c880ea7df3443289f25ea0R26-R29). This should be the economic committee
- [x] Implement encode/decode of transactions with [SetMsgFeeMsg](https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.21.3/docs/proto/index.html#msgfee.SetMsgFeeMsg)
- [x] Create proposal type that allows supports `SetMsgFeeMsg` as an action
- [x] Write test that changes to configuration of the local chain via governance (we can only have one admin, so we need to choose between direct address or governance rule)

Note: right now deploying 0.21.3 to IOV mainnet leads to a situation where we cannot access the new functionality since no `app_state.conf.msgfee.fee_admin` is set in mainnet genesis. This will be fixed one way or another. But IOV Core can implement and test the new functionality as long as it connects to newly spawned blockchains.

willclarktech:
@webmaster128 Is it definitely `UpdateConfigurationMsg` we need to support, not `SetMsgFeeMsg` (or both)?

webmaster128:
> Is it definitely `UpdateConfigurationMsg` we need to support, not `SetMsgFeeMsg` (or both)?

It is not, my fault. My thinking was that it is `UpdateConfigurationMsg` because in the `cash` module, the fee (`minimal_fee`) is part of the [Configuration](https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/v0.21.3/docs/proto/index.html#cash.Configuration) object.

`SetMsgFeeMsg` is what we need. Configuration updates are not required at the moment.

PR #1295, opened 2019-11-13
Add SetMsgFee ProposalAction and support encode/decode

willclarktech:
Part 3 of #1289

### DIFF
--- a/packages/iov-bns/src/types.ts
+++ b/packages/iov-bns/src/types.ts
@@ -125,6 +125,7 @@ export enum ActionKind {
   ExecuteProposalBatch = "execute_proposal_batch",
   ReleaseEscrow = "escrow_release",
   Send = "cash_send",
+  SetMsgFee = "msgfee_set_msg_fee",
   SetValidators = "validators_apply_diff",
   UpdateElectionRule = "gov_update_election_rule",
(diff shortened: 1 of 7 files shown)


packages/iov-bns/src/types.ts
```diff
@@ -177,6 +179,16 @@ export function isSendAction(action: ProposalAction): action is SendAction {
   return action.kind === ActionKind.Send;
 }
 
+export interface SetMsgFeeAction {
+  readonly kind: ActionKind.SetMsgFee;
+  readonly msgPath: string;
+  readonly fee: Fee;
```
webmaster128: We could use `readonly fee: Amount;` to simplify types a bit, since we will never need the generic transaction `Fee` type for this price change functionality.

willclarktech: I thought about this, but then I thought it might be more confusing than the simplicity was worth.

willclarktech: It does reduce the need for validation checks though.

webmaster128: I don't have a proper argument for any of the two options. I just think using `Amount` is simpeler. But both ways work.

PR #1298, opened 2019-11-20
Add BNS integration test for setting msg fee

### LINKED ISSUE #1289: Implement mutable product fees

willclarktech:
Closes #1289

We should have a follow-up issue to update the bnsd dockerhub image version once published.

### DIFF
--- a/scripts/bnsd/start.sh
+++ b/scripts/bnsd/start.sh
@@ -5,7 +5,7 @@ command -v shellcheck > /dev/null && shellcheck "$0"
 # Choose from https://hub.docker.com/r/iov1/tendermint/tags/
 export BNSD_TM_VERSION=v0.31.5
 # Choose from https://hub.docker.com/r/iov1/bnsd/tags/
-export BNSD_VERSION=v0.21.3
+export BNSD_VERSION=v0.22.0
 
 # get this files directory regardless of pwd when we run it
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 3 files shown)


scripts/bnsd/start.sh
```diff
@@ -5,7 +5,7 @@ command -v shellcheck > /dev/null && shellcheck "$0"
 # Choose from https://hub.docker.com/r/iov1/tendermint/tags/
 export BNSD_TM_VERSION=v0.31.5
 # Choose from https://hub.docker.com/r/iov1/bnsd/tags/
-export BNSD_VERSION=v0.21.3
+export BNSD_VERSION=latest
```
webmaster128: We have tag v0.22.0 now that we can use: https://hub.docker.com/r/iov1/bnsd/tags

Issue #1299, opened 2019-11-20
Allow fee payer to be specified in BNS


willclarktech:
Registrars should be able to pay all costs for a user's starname registration and/or target update.

webmaster128:
@willclarktech in which context is this needed? As far as I know, executing batch messages are only used for proposal executions. Do we need to decode those?

willclarktech:
@webmaster128 See https://iov.atlassian.net/wiki/spaces/PRODUCT/pages/14221370/Batch+transactions

PR #1301, opened 2019-11-27
Add registrar username BNS connection test

willclarktech:
Relates to #1299

This PR adds a test which proves the possibility (using existing functionality) of a registrar registering a username on behalf of a user and updating the targets of that username, with no cost to the user. The steps are as follows:

1. Registrar registers the desired username (only registrar signature required).
2. Registrar transfers the desired username to the user (only registrar signature required).
3. Username targets are updated with registrar as main signer and user as additional signer.

Batch transactions offer no benefit over existing functionality when it comes to updating the targets of a username. They do, however, provide potential benefits in the case of registering a username in the first place:
1. They are faster since registering/transferring can be performed in a single block.
2. They are cheaper because only one anti-spam fee needs to be paid.

All the same, IOV Core already provides sufficient functionality to meet the current acceptance criteria/success metrics for this issue.

### DIFF
--- a/packages/iov-bns/src/bnsconnection.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.spec.ts
@@ -52,6 +52,7 @@ import { decodeNumericId } from "./decode";
 import { bnsSwapQueryTag } from "./tags";
 import {
   ActionKind,
+  ChainAddressPair,
   CreateEscrowTx,
   CreateMultisignatureTx,
   CreateProposalTx,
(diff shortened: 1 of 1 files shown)


packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -2882,6 +2883,88 @@ describe("BnsConnection", () => {
     connection.disconnect();
   });
 
+  it("can register/transfer and update a username for an empty account", async () => {
+    pendingWithoutBnsd();
+    const connection = await BnsConnection.establish(bnsdTendermintUrl);
+    const chainId = connection.chainId();
+
+    const { profile, faucet, walletId } = await userProfileWithFaucet(chainId);
+    const emptyPath = HdPaths.iov(666);
```
webmaster128: `brokeAccountPath`?

packages/iov-bns/src/bnsconnection.spec.ts
```diff
@@ -2882,6 +2883,88 @@ describe("BnsConnection", () => {
...
+    const user = await profile.createIdentity(walletId, chainId, emptyPath);
+    const userAddress = identityToAddress(user);
+    const username = `user${Math.random()}*iov`;
+
+    const userAccount = await connection.getAccount({ address: userAddress });
+    if (userAccount && userAccount.balance.length) {
```
webmaster128: Is `userAccount.balance` still empty when the account exists on chain? Or to rephrase the question: can this test pass multiple times on the same chain?

willclarktech: Yes, the NFTs don't appear in the balance.

PR #1304, opened 2019-11-27
Remove support for Node.js <v10.13

### LINKED ISSUE #1300: Prepare releasing IOV Core 2
Edit Simon:
The new requirement for IOV Core 2 environments should be native BigInteger support, introduced in TypeScript 3.2. See also https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#caveats-1 and see what this means for us.

For me it means

- [x] Cutting a 1.2.0 release (~after #1298 is merged~), creating a 1.2 release branch for patches of necessary
- [x] Writing a nice little CHANGELOG entry for 2.0.0 defining the new runtime environment requirements
- [x] Change CI configuration to use the lowest version of nodejs we support from now

The last point is what unblocks merging Cosmos codec

willclarktech:
Closes #1300

I didn’t go too deep into the README as it could probably use a general update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 2.0.0
+
+Breaking changes
+
+- ALL: With v2.0.0 we are moving to the new native `BigInt` type. As a result
+  the minimum Node.js version is now 10.13. Only browsers which support `BigInt`
+  will be compatible with IOV Core. This currently includes Chrome, Firefox and
+  Opera, but not Edge, Internet Explorer or Safari.
+
 ## 1.2.0
 
 - @iov/bns: Upgrade weave to v0.22.0.
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ The following projects use IOV-Core (add yourself via a PR):
 ## Compatibility
 
 The compiled code from this package, which is published on npm, should work on
-any modern (2018) browser, and node 8+. The development environment has been
-tested on node 8.7.0 LTS and node 10.x.
+any modern (2018) browser, and node 10.13+. The development environment has been
+tested on node 10.13.
 
 **Yarn not Npm** Please `npm install -g yarn` and use `yarn install`,
 `yarn build`, etc. Developers who installed with `npm i` have reported problems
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.0.0
+
+Breaking changes
+
+- ALL: Minimum Node.js version is now 10.13.
```
webmaster128: Can we add a note on browser support?

README.md
```diff
@@ -39,26 +39,24 @@ The following projects use IOV-Core (add yourself via a PR):
 | Project                                                      | Tech stack / JavaScript environments |
 | ------------------------------------------------------------ | ------------------------------------ |
 | [TokenFinder](https://github.com/iov-one/token-finder)       | TypeScript / React / Browsers        |
-| [iov-faucet](https://github.com/iov-one/iov-faucet)          | TypeScript / NodeJS 10               |
+| [iov-faucet](https://github.com/iov-one/iov-faucet)          | TypeScript / NodeJS 10.13            |
```
webmaster128: [10.15 right now](https://github.com/iov-one/iov-faucet/blob/master/Dockerfile) but [could be changed to 12 or something like that](https://github.com/iov-one/iov-faucet/issues/50). It would be good if we just track the major version here.

CHANGELOG.md
willclarktech: Hmm, no `BigInt` support in Edge, IE or Safari right now according to MDN. Is it definitely the right time for this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Browser_compatibility

webmaster128: The only browser environments we that are important for us right now are the ones that suport chrome-like browser extensions. So I think modern Chromium-based browsers and Firefox is sufficient.

As far as I know, Microsoft is already planning to build a browser based on Chromium. The only use case for Safari I see is the token finder and a potential block explorer, which will not be out before summer.

So I think it is a far trade-off to require BigInt support.

webmaster128: Thanks for checking this by the way!

PR #1305, opened 2019-11-28
Allow fee payer to be specified in BNS

### LINKED ISSUE #1299: Allow fee payer to be specified in BNS

willclarktech:
Closes #1299

Exposes functionality required for #1299 without implementing batch transactions.

~~Not covered by this PR: decoding transactions to include the payer, which would perhaps require our `Fee` type to be updated.~~

### DIFF
--- a/packages/iov-bns/src/encode.ts
+++ b/packages/iov-bns/src/encode.ts
@@ -499,8 +499,8 @@ export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
   const msg = buildMsg(tx);
 
   let feePayerAddressBytes: Uint8Array;
-  if (tx.fee && tx.fee.feePayer) {
-    feePayerAddressBytes = decodeBnsAddress(tx.fee.feePayer).data;
+  if (tx.fee && tx.fee.payer) {
+    feePayerAddressBytes = decodeBnsAddress(tx.fee.payer).data;
   } else if (isMultisignatureTx(tx)) {
     const firstContract = tx.multisig.find(() => true);
     if (firstContract === undefined) throw new Error("Empty multisig arrays are currently unsupported");
--- a/packages/iov-bns/src/encode.ts
+++ b/packages/iov-bns/src/encode.ts
@@ -498,13 +498,15 @@ export function buildMsg(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
 export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
   const msg = buildMsg(tx);
 
-  let feePayer: Uint8Array;
-  if (isMultisignatureTx(tx)) {
+  let feePayerAddressBytes: Uint8Array;
+  if (tx.feePayer) {
+    feePayerAddressBytes = decodeBnsAddress(tx.feePayer).data;
+  } else if (isMultisignatureTx(tx)) {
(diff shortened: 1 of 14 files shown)


packages/iov-bns/src/encode.ts
```diff
@@ -513,15 +515,15 @@ export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
...
-            payer: feePayer,
+            payer: feePayerAddressBytes,
           }
         : null,
     multisig: isMultisignatureTx(tx) ? tx.multisig.map(encodeNumericId) : null,
   });
 }
 
-export function buildSignedTx(tx: SignedTransaction): codecImpl.bnsd.ITx {
+export function buildSignedTx(tx: SignedTransaction, feePayer?: Address): codecImpl.bnsd.ITx {
```
webmaster128: This signature seems to be incorrect. How can `tx` be "signed" when it does not contain the `feePayer` information?

I guess the `feePayer?` needs to become an attribute of `tx.transaction.fee`

willclarktech: When the signature was created, the function `buildUnsignedTx` would have been called with the fee payer, meaning the fee payer is in the `bytesToSign` result. But by the time this gets back to `buildSignedTx` we've turned it back into a BCP tx object, so we've dropped that information and need to supply it again.

We can either make it part of the fee, or add a (potentially optional) `feePayer` attribute to a transaction and assume that only one person pays all the fees for the whole transaction.

webmaster128: > by the time this gets back to buildSignedTx we've turned it back into a BCP tx object, so we've dropped that information and need to supply it again.

So it should be in the BCP tx object

> We can either make it part of the fee, or add a (potentially optional) feePayer attribute to a transaction

I think an optional `payer` in `Fee` is clean.

> and assume that only one person pays all the fees for the whole transaction

I am not aware of any case where multiple accounts could split the fee

willclarktech: > assume that only one person pays all the fees for the whole transaction
I meant "assume that all the fees for the whole transaction are paid by only one person". The question is whether we individuate the fee payer on a per-transaction or per-fee basis. Per-fee is better aligned with the back end.

webmaster128: > The question is whether we individuate the fee payer on a per-transaction or per-fee basis. Per-fee is better aligned with the back end.

I guess I don't understand. A sindle transaction has a single fee field (called plural "fees" for some reason). This single fee field can be unset or a single `FeeInfo` object, containing a single amount and a payer.

willclarktech: Ah, sorry, my mistake, I was confusing the fees associated with different message types with the fee associated with a transaction.

packages/iov-bns/src/encode.ts
```diff
@@ -513,7 +515,7 @@ export function buildUnsignedTx(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
       tx.fee && tx.fee.tokens
         ? {
             fees: encodeAmount(tx.fee.tokens),
-            payer: feePayer,
+            payer: feePayerAddressBytes,
           }
```
webmaster128: What about moving the payer into the `Fee` object? It ensures no payer is set if no fee is set. Here you see that this would be closer to the weave implementation.

PR #1312, opened 2019-12-05
Specify signing identity in UserProfile.signTransaction

### LINKED ISSUE #1308: Make signing identity explicit in UserProfile.signTransaction
We want to get to the point where an independent 3rd party can create an `UnsignedTransaction`, which is then signed by the required parties. To prepare for that, the signing identity should be passed explicitely into `UserProfile.signTransaction`, the same way it is done in `UserProfile.appendSignature`.

willclarktech:
Closes #1308

### DIFF
--- a/packages/iov-keycontrol/src/userprofile.ts
+++ b/packages/iov-keycontrol/src/userprofile.ts
@@ -315,21 +315,26 @@ export class UserProfile {
    * creator field specifies the keypair to be used for signing.
    */
   public async signTransaction(
+    identity: Identity,
     transaction: UnsignedTransaction,
     codec: TxCodec,
     nonce: Nonce,
   ): Promise<SignedTransaction> {
-    const wallet = this.findWalletInPrimaryKeyringByIdentity(transaction.creator);
+    const wallet = this.findWalletInPrimaryKeyringByIdentity(identity);
+    const transactionWithExplicitSigner = {
+      ...transaction,
+      creator: identity,
+    };
 
-    const { bytes, prehashType } = codec.bytesToSign(transaction, nonce);
+    const { bytes, prehashType } = codec.bytesToSign(transactionWithExplicitSigner, nonce);
     const signature: FullSignature = {
-      pubkey: transaction.creator.pubkey,
+      pubkey: identity.pubkey,
       nonce: nonce,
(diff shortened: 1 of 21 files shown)


packages/iov-keycontrol/src/userprofile.ts
```diff
@@ -315,21 +315,26 @@ export class UserProfile {
...
   ): Promise<SignedTransaction> {
-    const wallet = this.findWalletInPrimaryKeyringByIdentity(transaction.creator);
+    const wallet = this.findWalletInPrimaryKeyringByIdentity(identity);
+    const transactionWithExplicitSigner = {
+      ...transaction,
+      creator: identity,
+    };
 
-    const { bytes, prehashType } = codec.bytesToSign(transaction, nonce);
+    const { bytes, prehashType } = codec.bytesToSign(transactionWithExplicitSigner, nonce);
```
webmaster128: Can we add `signer: Identity` as a new, second argument to `TxCodec.bytesToSign` and not use the creator anymore in the codec? This would avoid the need for the `transactionWithExplicitSigner` construction

willclarktech: I was thinking of doing this in a separate PR, but can definitely include it here.

webmaster128: Great, let's merge this then.

willclarktech: Actually `bytesToSign` just needs the chain ID, not the full signer anyway.

webmaster128: Even better

webmaster128: Hmm, looks like we need a sender pubkey for Lisk send transactions:

```ts
      return new Uint8Array([
        0, // transaction type
        ...timestampBytes,
        ...unsigned.creator.pubkey.data,
        ...recipient.toBytesBE(),
        ...amount.toBytesLittleEndian(),
        ...memoBytes,
      ]);
```

willclarktech: Ah yeah, I was just thinking about BNS. Hm, wonder why it's the public key not the address?

PR #1319, opened 2019-12-05
Specify signing identity in json rpc signAndPost call

### LINKED ISSUE #1315: Add signing identity to signAndPost RPC call
Follow-up to #1308

willclarktech:
Closes #1315


packages/iov-multichain/src/jsonrpcsigningserver.spec.ts
```diff
@@ -247,6 +247,7 @@ describe("JsonRpcSigningServer", () => {
       method: "signAndPost",
       params: {
         reason: "string:Please sign",
+        identity: TransactionEncoder.toJson(signer),
```
webmaster128: Can we call this `signer` and update https://github.com/iov-one/iov-core/blob/master/docs/out-of-process-rpc.md?

willclarktech: Updated. It's a little confusing that we also use `signer` in lots of examples to refer to an instance of `MultiChainSigner` though.

webmaster128: Hmm, yes, right, good point. The tool and the tool's user have the same name. I don't see an easy way to resolve this right now.

willclarktech: Red signer and blue signer?

webmaster128: "Open signer" would be good for a tool that is open to be used by anybody, such as closed signers.

PR #1325, opened 2019-12-10
Update docs

webmaster128:


packages/iov-tendermint-rpc/README.md
```diff
@@ -3,8 +3,8 @@
 [![npm version](https://img.shields.io/npm/v/@iov/tendermint-rpc.svg)](https://www.npmjs.com/package/@iov/tendermint-rpc)
 
 This package provides a type-safe wrapper around
-[Tendermint RPC](https://tendermint.com/rpc/). Notably, all binary data is
-passed in and out as Uint8Array, and this module is reponsible for the
+[Tendermint RPC](https://docs.tendermint.com/master/rpc/). Notably, all binary
```
willclarktech: Can we give a version rather than brittle `master`?

webmaster128: I don't see a way to do that. Both branch names and version tags did not work.

PR #1327, opened 2019-12-11
Remove creator from bytesToSign

### LINKED ISSUE #1314: Refactor `codec.bytesToSign` to accept an unsigned transaction without a creator
Follow-up to #1308 
This may directly involve replacing the `WithCreator` type with a `WithChainId` type.

willclarktech:
Closes #1314

Also replaces `creator` with `chainId` in `UnsignedTransaction` and removes `LightTransaction`. An optional `senderPublicKey` field is added to `SendTransaction` for codecs such as Lisk’s which require it.


packages/iov-bns/src/encode.ts
```diff
@@ -498,15 +497,15 @@ export function buildMsg(tx: UnsignedTransaction): codecImpl.bnsd.ITx {
...
-  let feePayerAddressBytes: Uint8Array;
-  if (tx.fee && tx.fee.payer) {
-    feePayerAddressBytes = decodeBnsAddress(tx.fee.payer).data;
-  } else if (isMultisignatureTx(tx)) {
-    const firstContract = tx.multisig.find(() => true);
-    if (firstContract === undefined) throw new Error("Empty multisig arrays are currently unsupported");
-    feePayerAddressBytes = conditionToWeaveAddress(buildMultisignatureCondition(firstContract));
-  } else {
-    feePayerAddressBytes = decodeBnsAddress(identityToAddress(tx.creator)).data;
+  let feePayerAddressBytes: Uint8Array = new Uint8Array();
```
webmaster128: We add a little bit of type strangeness here by requiring to set `feePayerAddressBytes` for the case `tx.fee && tx.fee.tokens` is false. However, we only need the variable for `tx.fee && tx.fee.tokens` is true (see below). Can we rewrite this bit in a way that only sets `const feePayerAddressBytes` for `tx.fee && tx.fee.tokens`?

packages/iov-cli/src/cli.ts
```diff
@@ -44,7 +44,7 @@ export function main(originalArgs: readonly string[]): void {
         "SendTransaction",
         "TokenTicker",
         "TransactionId",
-        "WithCreator",
+        "WithChainId",
```
webmaster128: Does this still exist?

PR #1335, opened 2019-12-12
Implement transaction tracking in Cosmos postTx

### LINKED ISSUE #1317: Implement transaction's processing progress in CosmosConnection.postTx
~Right now I get fails like~

```
1) CosmosConnection integration tests can post and get a transaction
  - Error: Transaction does not exist
      at CosmosConnection.<anonymous> (blub/packages/iov-cosmos/src/cosmosconnection.ts:181:15)
      at <Jasmine>
      at rejected (blub/packages/iov-cosmos/build/cosmosconnection.js:6:65)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

~See also https://travis-ci.com/iov-one/iov-core/jobs/263542715#L1549~

~This makes sense, since the test code does not wait for the transaction to be processed. Adding `await response.blockInfo.waitFor(info => !isBlockInfoPending(info));` however leads to a timeout, since the `response.blockInfo` is never updated.~

--------

The above error might be unrelated. However, we should implement transaction tracking.

willclarktech:
Closes #1317

I suggest merging this as a quick solution for now, and doing a proper version if/when we choose to support RPC connections. Is this everything required to close #1317?

### DIFF
--- a/packages/iov-cosmos/src/cosmosconnection.ts
+++ b/packages/iov-cosmos/src/cosmosconnection.ts
@@ -13,6 +13,7 @@ import {
   ConfirmedTransaction,
   FailedTransaction,
   Fee,
+  isConfirmedTransaction,
   isPubkeyQuery,
   isSendTransaction,
   Nonce,
@@ -29,6 +30,7 @@ import {
 } from "@iov/bcp";
 import { Encoding, Uint53 } from "@iov/encoding";
 import { DefaultValueProducer, ValueAndUpdates } from "@iov/stream";
+import equal from "fast-deep-equal";
 import { ReadonlyDate } from "readonly-date";
 import { Stream } from "xstream";
 
@@ -42,6 +44,9 @@ const { fromBase64 } = Encoding;
 interface ChainData {
   readonly chainId: ChainId;
 }
+
+// poll every 0.5 seconds (block time 1s)
+const defaultPollInterval = 500;
 
 function buildQueryString({
   height,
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -3216,6 +3216,11 @@ fast-deep-equal@^2.0.1:
   resolved "https://registry.yarnpkg.com/fast-deep-equal/-/fast-deep-equal-2.0.1.tgz#7b05218ddf9667bf7f370bf7fdb2cb15fdd0aa49"
   integrity sha1-ewUhjd+WZ79/Nwv3/bLLFf3Qqkk=
 
+fast-deep-equal@^3.1.1:
```
webmaster128: Would be cool if we could at some point use fast-deep-equal 3 for all our packages where we currently have a fast-deep-equal 2 dependency

packages/iov-cosmos/src/cosmosconnection.ts
```diff
@@ -191,7 +196,35 @@ export class CosmosConnection implements BlockchainConnection {
...
     const transactionId = txhash as TransactionId;
     const firstEvent: BlockInfo = { state: TransactionState.Pending };
-    const producer = new DefaultValueProducer<BlockInfo>(firstEvent);
+    let blockInfoInterval: NodeJS.Timeout;
+    let lastEventSent: BlockInfo;
+    const producer = new DefaultValueProducer<BlockInfo>(firstEvent, {
+      onStarted: () => {
+        blockInfoInterval = setInterval(async () => {
+          const search = await this.searchTx({ id: transactionId });
+          if (search.length) {
```
webmaster128: I think this "first" implementation gives you better type safety:

```ts
const foundTransaction = (await this.searchTx({ id: transactionId })).find(() => true);
```

However, this is a matter of taste I guess.

webmaster128:
We could use that in the "can post and get a transaction" test now, since the line `const { transactionId } = await connection.postTx(postableBytes);` does not guarantee the transaction is in a block. Maybe this is ensured by the implementation of the rest server, but the interface does not guarantee it.

packages/iov-cosmos/src/cosmosconnection.ts
willclarktech: I was just copying the pattern from the Lisk connection but I agree this is nicer.

yarn.lock
willclarktech: https://github.com/iov-one/iov-core/issues/1337

PR #1336, opened 2019-12-12
Combine primary and other signatures in SignedTransaction

### LINKED ISSUE #1334: Combine primarySignature and otherSignatures
On some chains the order of signatures is important, on others (eg BNS) it isn't, and the term `primary` is misleading. We can use a type which represents a non-empty array as follows:

```ts
type NonEmptyArray<T> = {
    0: T
} & Array<T>

const a: NonEmptyArray<string> = [] // got expected error
const b: NonEmptyArray<string> = ["a"] // no error
const c: NonEmptyArray<string> = ["a", "b"] // no error

function isNonEmptyArray<T>(data: T[]): data is NonEmptyArray<T> {
    return data.length >= 1;
}

const d = isNonEmptyArray<string>([]); // false
const e = isNonEmptyArray<string>(["a"]); // true
const f = isNonEmptyArray<string>(["a", "b"]); // true
```

willclarktech:
Closes #1334

Would need a CHANGELOG entry before merging.

### DIFF
--- a/packages/iov-bns/src/bnsconnection.txs.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.txs.spec.ts
@@ -123,7 +123,10 @@ describe("BnsConnection (txs)", () => {
       if (isFailedTransaction(result)) {
         throw new Error("Expected ConfirmedTransaction, received FailedTransaction");
       }
-      const { transaction, primarySignature: signature } = result;
+      const {
+        transaction,
+        signatures: [signature],
+      } = result;
       if (!isSendTransaction(transaction)) {
         throw new Error("Expected send transaction");
       }
--- a/packages/iov-bns/src/decode.ts
+++ b/packages/iov-bns/src/decode.ts
@@ -5,6 +5,7 @@ import {
   ChainId,
   FullSignature,
   Hash,
+  newNonEmptyArray,
   Nonce,
   Preimage,
   PubkeyBundle,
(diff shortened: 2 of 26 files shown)


packages/iov-bns/src/bnsconnection.txs.spec.ts
```diff
@@ -123,7 +123,10 @@ describe("BnsConnection (txs)", () => {
       if (isFailedTransaction(result)) {
         throw new Error("Expected ConfirmedTransaction, received FailedTransaction");
       }
-      const { transaction, primarySignature: signature } = result;
+      const {
+        transaction,
+        signatures: [signature],
+      } = result;
```
webmaster128: nice!

packages/iov-bns/src/decode.ts
```diff
@@ -756,11 +757,14 @@ function parseBaseTx(tx: codecImpl.bnsd.ITx, sig: FullSignature, chainId: ChainI
 }
 
 export function parseTx(tx: codecImpl.bnsd.ITx, chainId: ChainId): SignedTransaction {
-  const sigs = ensure(tx.signatures, "signatures").map(decodeFullSig);
-  const sig = ensure(sigs[0], "first signature");
+  let signatures;
+  try {
+    signatures = newNonEmptyArray(ensure(tx.signatures, "signatures").map(decodeFullSig));
+  } catch (error) {
+    throw new Error("Transaction has no signatures");
```
webmaster128: or it has but decodeFullSig failed for one of them

webmaster128:
👍 just the CHANGELOG missing

PR #1368, opened 2020-01-15
WIP: Begin process of implementing Cosmos RPC

### LINKED ISSUE #1339: Connect to Cosmos via Tendermint RPC
The ultimate goal is to utilize the subscription functionality for transactions and accounts. Especially a `watchAccount` and a `liveTx` implementation are required for wallet integration.

willclarktech:
Indication of the process required to close #1339

### DIFF
--- a/packages/iov-cosmos/src/cosmosconnection.ts
+++ b/packages/iov-cosmos/src/cosmosconnection.ts
@@ -26,18 +26,21 @@ import {
   TransactionId,
   TransactionQuery,
   TransactionState,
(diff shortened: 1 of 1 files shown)


webmaster128:
Connecting to Tendermint directly is almost impossible to get right, since you need a local Amino encoding/decoding library. Such a library does not exist in any well-maintained way. Things will change with the upcoming Cosmos SDK 0.39 that uses protobuf.

For now I recommend closing this PR.

PR #1370, opened 2020-01-16
Update Caip5 implementation to latest version of spec

webmaster128:
See https://github.com/ChainAgnostic/CAIPs/pull/9


antoineherzog:
this is not the final version We need to wait a little more

webmaster128:
The CAIP-5 standard was merged and this implementation is up-to-date.

PR #1371, opened 2020-01-23
License cleanup and new year's update

webmaster128:


NOTICE
```diff
@@ -1,10 +1,11 @@
-Copyright 2018-2019 IOV SAS
+Copyright 2018-2020 IOV SAS
+Copyright 2020 Simn Warta
```
davepuchyr: I though your name was "Simon". ;)

I'm happy to merge this after you've made a contribution since your departure.

webmaster128: Ah, thanks. Typing without reading is not a good idea.

Issue #1181, opened 2019-08-08
Collect important tslint rules and find eslint alternatives


webmaster128:
The purpose of this ticket is to track the status of eslint alternatives for tslint rules we like. This allows us to see how far we are from ditching tslint.

tslint core:

- [ ] [deprecation](https://palantir.github.io/tslint/rules/deprecation/)
- [ ] [no-floating-promises](https://palantir.github.io/tslint/rules/no-floating-promises/)
- [ ] [prefer-readonly](https://palantir.github.io/tslint/rules/prefer-readonly/)

tslint-immutable:

- [ ] [no-method-signature](https://github.com/jonaskello/tslint-immutable#no-method-signature)
- [ ] [readonly-keyword](https://github.com/jonaskello/tslint-immutable#readonly-keyword)
- [ ] [readonly-array](https://github.com/jonaskello/tslint-immutable#readonly-array)
- [ ] [no-delete](https://github.com/jonaskello/tslint-immutable#no-delete)

PR #1381, opened 2020-02-07
Add await-promise linter rule and fix code

webmaster128:
https://palantir.github.io/tslint/rules/await-promise/


webmaster128:
I added this to the list of missing features in eslint: https://github.com/iov-one/iov-core/issues/1181

PR #1382, opened 2020-02-10
Deposit query

albertandrejev:
"Implement `/deposits` query" part of the https://github.com/iov-one/ponferrada/issues/997 issue.

### DIFF
--- a/packages/iov-bns/src/bnsconnection.ts
+++ b/packages/iov-bns/src/bnsconnection.ts
@@ -68,13 +68,16 @@ import {
   decodeUsernameNft,
   decodeVote,
 } from "./decodeobjects";
+import { decodeTermDepositNft } from "./decodeobjects";
 import * as codecImpl from "./generated/codecimpl";
 import { bnsSwapQueryTag } from "./tags";
 import {
+  BnsTermDepositNft,
   BnsTx,
   BnsUsernameNft,
   BnsUsernamesQuery,
   Decoder,
+  DepositContractIdBytes,
   ElectionRule,
   Electorate,
   isBnsTx,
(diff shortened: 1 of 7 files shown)


packages/iov-bns/src/bnsconnection.ts
```diff
@@ -745,6 +754,43 @@ export class BnsConnection implements AtomicSwapConnection {
...
+    } else if (isBnsDepositsByContractIdQuery(query)) {
+      keyPrefix = "";
+      results = (await this.query("/deposits/contract", query.depositContractId)).results;
+    } else if (isBnsDepositByDepositIdQuery(query)) {
+      results = (await this.query("/deposits", query.depositId)).results;
+    } else {
+      throw new Error("Unsupported query");
+    }
+
+    const parser = createParser(codecImpl.termdeposit.Deposit, keyPrefix);
```
albertandrejev: For unknown reason keyPrefix for `depositor` and `contract` indexes should be empty, because BNS returns result and this results key contains record id only and did not contain any prefix. Is this a bug or feature? @davepuchyr

packages/iov-bns/src/bnsconnection.ts
```diff
@@ -745,6 +754,43 @@ export class BnsConnection implements AtomicSwapConnection {
     return votes;
   }
 
+  public async getDeposits(query: BnsDepositQuery): Promise<readonly BnsTermDepositNft[]> {
+    let keyPrefix = "deposit:";
```
webmaster128: When you write `let keyPrefix: string`, TypeScript will ensure that you set the value in every code branch. No way to accidentally forget it.

packages/iov-bns/src/bnsconnection.ts
davepuchyr: `/deposits/depositor` and `/deposits/contract` are both paths that point directly to indexes and, therefore, don't need a prefix.  `/deposits` points to a bucket whose keys are prefixed with `deposit:`, hence the need for a non-empty `keyPrefix` in the parser for the bucket query.  Does that make sense?

albertandrejev: But in [this case](https://github.com/iov-one/iov-core/blob/b271c195f199f59d96daa7c9ef4c935284c60337/packages/iov-bns/src/bnsconnection.ts#L777) index returns keys with prefix in it.

davepuchyr: Ah.  Piotr refactored the indexing algorithm.  Since `usernames` are deprecated I guess he (correctly) didn't seek backward compatibility.

PR #1387, opened 2020-02-13
Test format of contract ID

webmaster128:
1. Migrate the linter rule `no-unnecessary-type-assertion` from tslint to eslint, since the tslint implementation did not work well and tslint is deprecated.
2. Test format of `contractId` for `bns/create_multisignature_contract`

@davepuchyr as you can see, the contract ID is a fixed length 8 byte value containing a small number (a counter), e.g. `0000000000000004` (hex).


davepuchyr:
@webmaster128 , why is `contractId` a [Uint8Array](https://github.com/iov-one/iov-core/blob/master/packages/iov-bns/src/types.ts#L629) when all other ids (`electorateId`, `escrowId`, etc) are [number](https://github.com/iov-one/iov-core/blob/master/packages/iov-bns/src/types.ts#L183)s?

webmaster128:
Historic reasons: Multisig contracts existed first. And from the backend's point of view it makes sense to "simply" use uint64. But then in JavaScript world cou can only store integers up to `Number.MAX_SAFE_INTEGER` without rounding (which is about 53 bit integers). Now you have two ways to handle that:
1. Represent the IDs as raw binary data, allowing the full uint64 range to work
2. (Ab-)use the knowledge that the values are auto incrementing counters starting at 1 that will not exceed `MAX_SAFE_INTEGER`

1 was the original strategy and it has the advantage that the client does not make assumptions on how the IDs are structured. Backend could choose to use random values instead of a counter. But the development and user experience was very bad, which is why we moved to 2. when working on all this governance.

Concerting contract ID to `number` would make sense to me, espelcially since `MultisignatureTx.multisig` uses an array of `number` already. So yeah, it should be done.

PR #1388, opened 2020-02-13
Specify VoteOption values

davepuchyr:
https://github.com/iov-one/ledger-iov/pull/30#discussion_r378744878

### DIFF
--- a/packages/iov-tendermint-rpc/package.json
+++ b/packages/iov-tendermint-rpc/package.json
@@ -46,5 +46,8 @@
     "readonly-date": "^1.0.0",
     "type-tagger": "^1.0.0",
     "xstream": "^11.10.0"
+  },
+  "devDependencies": {
+    "@iov/utils": "^2.0.0-alpha.7"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/iov-tendermint-rpc/package.json
```diff
@@ -46,5 +46,8 @@
     "readonly-date": "^1.0.0",
     "type-tagger": "^1.0.0",
     "xstream": "^11.10.0"
+  },
+  "devDependencies": {
+    "@iov/utils": "^2.0.0-alpha.7"
```
webmaster128: Good fix. This is used in `packages/iov-tendermint-rpc/src/client.spec.ts`: `import { sleep } from "@iov/utils";` but the dependency was missing.

davepuchyr: I didn't like sneaking it in with the VoteOptions stuff, but must press forward.

webmaster128: You could have afforded a second commit at least ;)

PR #1391, opened 2020-02-13
IOV Core 2.0.0 release

### LINKED ISSUE #1359: Release final version of IOV Core 2.0.0
Everthing is done from my side and this is ready to be released. It just needs a tag and be published to npm.

albertandrejev:
closes #1359.


webmaster128:
2.0.1 is already tagged and on npm. Let's merge this.

There are already breaking changes on master, so I suggest creating a 2.0.2 on master and rename the 2.0.0 CHANGELOG entry to 2.0.2.

davepuchyr:
> 2.0.1 is already tagged and on npm. Let's merge this.
> 
> There are already [breaking](https://github.com/iov-one/iov-core/pull/1394) [changes](https://github.com/iov-one/iov-core/pull/1392) on master, so I suggest creating a 2.0.2 on master and rename the 2.0.0 CHANGELOG entry to 2.0.2.

Abel is adding #1393 to the mix and will follow your suggestions.

lerna.json
```diff
@@ -1,6 +1,6 @@
 {
   "lerna": "2.10.2",
-  "version": "2.0.0-alpha.7",
+  "version": "2.0.1",
```
davepuchyr: @webmaster128 , should `version` be `2.0.2` given the mess that we've created?  I'm not familiar with lerna.

webmaster128: Yes. Here you find all the info how to make the release (directly from master, no PR):

https://github.com/iov-one/iov-core/wiki/Publishing

webmaster128: `./node_modules/.bin/lerna publish` tags, pushes and publishes on npm. So if you did that, you're done and having a PR is pointless.

webmaster128: The add nonces step 3. ensures that all packages have a change (creation of an empty file), which ensures that all packages get the same version. If you don't do that, only packages with changes get a new version.

davepuchyr: Thank-you, Simon.

PR #1392, opened 2020-02-13
Convert UpdateMultisignatureTx.contractId to number

davepuchyr:
https://github.com/iov-one/iov-core/pull/1387#issuecomment-585864737

### DIFF
--- a/packages/iov-bns/src/bnsconnection.posttxs.spec.ts
+++ b/packages/iov-bns/src/bnsconnection.posttxs.spec.ts
@@ -12,7 +12,7 @@ import {
   UnsignedTransaction,
 } from "@iov/bcp";
 import { Random } from "@iov/crypto";
-import { Encoding, Uint64 } from "@iov/encoding";
+import { Encoding } from "@iov/encoding";
 import { Ed25519HdWallet, HdPaths, UserProfile } from "@iov/keycontrol";
 import { assert, sleep } from "@iov/utils";
 import BN from "bn.js";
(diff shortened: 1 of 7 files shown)


packages/iov-bns/src/bnsconnection.posttxs.spec.ts
```diff
@@ -930,7 +930,7 @@ describe("BnsConnection (post txs)", () => {
       assert(contractId, "Contract ID must be set");
       // Contract ID is an incrementing fixed length uint64
       expect(contractId.length).toEqual(8);
-      expect(Uint64.fromBytesBigEndian(contractId).toNumber()).toBeLessThan(800);
+      expect(decodeNumericId(contractId)).toBeLessThan(Number.MAX_SAFE_INTEGER);
```
webmaster128: Here you want to convert from the blockchain type to the JavaScript type ASAP. E.g.

```ts
const { result: createContractResult, transaction: firstSearchResultTransaction } = searchResult1[0];
assert(createContractResult, "Contract ID must be set in the result of the create contract transaction");
const contractId = decodeNumericId(createContractResult);
```

Then there is no point in testing anything about `contractId` anymore, since `decodeNumericId` ensures it is in the safe integer range.

packages/iov-bns/src/bnsconnection.posttxs.spec.ts
```diff
@@ -953,7 +953,7 @@ describe("BnsConnection (post txs)", () => {
         {
           kind: "bns/update_multisignature_contract",
           chainId: registryChainId,
-          contractId: contractId,
+          contractId: decodeNumericId(contractId),
```
webmaster128: this simply becomes `          contractId: contractId,`

packages/iov-bns/src/bnsconnection.posttxs.spec.ts
davepuchyr: https://github.com/iov-one/iov-core/pull/1176/commits/545df0dc6ab749131fd0fb2a08bbef3bc5191069 (wrong branch; I need coffee)

PR #1415, opened 2020-04-03
Upgrade bns codec to v1.0.4

davepuchyr:


packages/iov-bns/src/decodeobjects.ts
```diff
@@ -170,6 +170,7 @@ export function decodeDomain(prefix: IovBech32Prefix, domain: codecImpl.account.
     hasSuperuser: ensure(domain.hasSuperuser, "hasSuperuser"),
     msgFees: ensure(domain.msgFees, "msgFees").map(decodeAccountMsgFee),
     accountRenew: asIntegerNumber(ensure(domain.accountRenew, "accountRenew")),
+    broker: domain.broker ? encodeBnsAddress(prefix, domain.broker) : ("" as Address),
```
davepuchyr, line 173: @albertandrejev, is `("" as Address)` the right value for the case when there's no broker?  If not then what's the right value?

abefernan: I believe so, the Domain interface does not let broker be optional. The only concern is what bnsd does with and empty string broker but I think it's correct.

packages/iov-bns/src/decodemsg.ts
```diff
@@ -231,7 +231,7 @@ function decodeRegisterAccountTx(
     name: ensure(msg.name, "name"),
     owner: encodeBnsAddress(prefix, ensure(msg.owner, "owner")),
     targets: ensure(msg.targets, "targets").map(decodeBlockchainAddress),
-    broker: encodeBnsAddress(prefix, ensure(msg.thirdPartyToken, "owner")) || undefined,
+    broker: encodeBnsAddress(prefix, ensure(msg.broker, "owner")) || undefined,
```
abefernan, line 234: Is it intentional that broker is "owner" in `decodeRegisterAccountTx` but "broker" in `decodeRegisterDomainTx`?.
I understand that `thirdPartyToken` also did this but just wanted to be sure.

davepuchyr: Great catch!!!  That's a bug.  I'll fix it.

davepuchyr: Fixed by https://github.com/iov-one/iov-core/pull/1419

PR #1435, opened 2020-05-29
Add HdPaths.iovSecondGen path maker

webmaster128:
To refactor code like

```ts
const identity = await wallet.createIdentity(chainId, HdPaths.bip44(234, 0, 0, 0));
```

to

```ts
const identity = await wallet.createIdentity(chainId, HdPaths.iovSecondGen(0));
```

making it much harder to create bugs. Happy for better naming suggestions, if you have one.

The second commit is just test structure cleanup and doen't change things.

### DIFF
--- a/packages/iov-keycontrol/src/hdpaths.ts
+++ b/packages/iov-keycontrol/src/hdpaths.ts
@@ -59,6 +59,16 @@ export class HdPaths {
   }
 
   /**
+   * A second generation IOV HD path in the form m/44'/234'/0'/0/a.
+   * This is used on the Cosmos SDK based chain starting from summer 2020.
+   *
+   * @param account The account index `a` starting at 0
+   */
+  public static iovSecondGen(account: number): readonly Slip10RawIndex[] {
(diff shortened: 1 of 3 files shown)


packages/iov-keycontrol/src/hdpaths.ts
```diff
@@ -58,6 +58,16 @@ export class HdPaths {
     return HdPaths.bip44Like(HdPaths.coinTypes.iov, account);
   }
 
+  /**
+   * A second generation IOV HD path in the form m/44'/234'/0'/0/a.
+   * This is used on the Cosmos SDK based chain starting from summer 2020.
+   *
+   * @param account The account index `a` starting at 0
+   */
+  public static iovSecondGen(account: number): readonly Slip10RawIndex[] {
```
davepuchyr, line 67: Let's be more explicit with the name. Any protest against `iovCosmosSdk`?

webmaster128: No protest, will update

webmaster128: updated

PR #1441, opened 2020-06-16
Use CosmJS implementations in crypto/encoding/utils

webmaster128:
Trim down codebase in a backwards compatible way by using @cosmjs/* packages (required by @iov/cosmos-sdk already). This prepares [IOV Core 3.0](https://github.com/iov-one/iov-core/issues/1442).


webmaster128:
This can be released as 2.5.0 since it only swaps out implementations and does not break interfaces. Then you can do the 3.0 as part of the mainnet transition.

